PR #423, opened 2014-05-28
(MODULES-160) Fix creating and checking role in Postgresql::server::role

mradcliffe:
A role will not be created when updating puppet if it was not created the first
time without this patch. This is a problem when puppet fails to switch locales
on certain systems. This patch changes the behavior of Postgresql::server::role
similar to logic in Postgresql::server::database.

### DIFF
--- a/manifests/server/role.pp
+++ b/manifests/server/role.pp
@@ -31,6 +31,7 @@ define postgresql::server::role(
     $environment  = []
   }
 
+  # Set the defaults for the postgresql_psql resource
   Postgresql_psql {
     db         => $db,
     port       => $port,
@@ -43,10 +44,12 @@ define postgresql::server::role(
     ],
   }
 
-  postgresql_psql { "CREATE ROLE ${username} ENCRYPTED PASSWORD ****":
+  postgresql_psql { "Check for existence of role '${username}'":
     command     => "CREATE ROLE \"${username}\" ${password_sql} ${login_sql} ${createrole_sql} ${createdb_sql} ${superuser_sql} ${replication_sql} CONNECTION LIMIT ${connection_limit}",
     unless      => "SELECT rolname FROM pg_roles WHERE rolname='${username}'",
     require     => Class['Postgresql::Server'],
+    db          => $db,
+    port        => $port,
     environment => $environment,
   }
 


tphoney:
Could you please squash your commits, @hunner or myself will merge the changes.
Thanks for the effort you put in, and sorry about the delay on getting back to you.

mradcliffe:
Sure thing. I'll hopefully be able to take a look at it again within the week or so.

manifests/server/role.pp
```diff
@@ -41,13 +42,15 @@
...
       Class['postgresql::server'],
     ],
+    psql_path  => $psql_path
   }
 
-  postgresql_psql { "CREATE ROLE ${username} ENCRYPTED PASSWORD ****":
-    command     => "CREATE ROLE \"${username}\" ${password_sql} ${login_sql} ${createrole_sql} ${createdb_sql} ${superuser_sql} ${replication_sql} CONNECTION LIMIT ${connection_limit}",
-    unless      => "SELECT rolname FROM pg_roles WHERE rolname='${username}'",
-    require     => Class['Postgresql::Server'],
-    environment => $environment,
```
mradcliffe: This probably should not be deleted.

manifests/server/role.pp
```diff
@@ -41,13 +42,15 @@
       Postgresql_psql["CREATE ROLE ${username} ENCRYPTED PASSWORD ****"],
       Class['postgresql::server'],
     ],
+    psql_path  => $psql_path
```
mradcliffe: Duplicate key/value after rebase against upstream master

mradcliffe:
I'll need to test this manually again with a full vagrant / puppet rebuild.

mradcliffe:
I built a virtual machine running puphpet/debian75-x64 and was no longer able to reproduce.

The last debian box I used was debian72-x64-vbox43.

I made the changes that I noted in the most recent diff, but I don't think those changes are relevant to the issue anymore. The addition of $db and $port may still be useful, but not in the scope of the original issue.

I should check on debian72 and centos65 boxes next.

mhaskel:
@mradcliffe I think changing the title of the `postgresql_psql` resource caused a unit test failure. Can you either update the test or revert the title change? Thanks!

tphoney:
Changes needed in:
spec/unit/defines/server/role_spec.rb

+  it 'should Check for existence of role "test"' do		
+    is_expected.to contain_postgresql_psql("Check for existence of role 'test'").with({

tphoney:
Thanks for the work you have put in to this pull request. However due to inactivity and failing unit tests / n needs rebased .

This pull request is being closed. If you want to re-submit this pr with the necessary changes, we will look at the new pull request.

PR #477, opened 2014-08-27
Fix lsbmajdistreleasee fact for Ubuntu

kbarber:
Since facter 2.2.0 'fixed' the lsbmajdistrelease fact for Ubuntu, we now have to regexp
for the value. The new value would be '10.04' whereas the old is '10'.

Signed-off-by: Ken Barber ken@bob.sh


hunner:
Rebased commit: 5ac19f97f143e9f73431033ec3cc5833d50f3ca1

PR #500, opened 2014-09-10
    Changes to support OpenBSD.

buzzdeee:
I messed up the other fork a bit, therefore created just new one, and new pull request 
for the same changes to support OpenBSD:    

Added $service_reload parameter to params.pp, in order to allow
    a reload of the postgresql server on OpenBSD, which apparently doesn't
    have a service binary.

    Add $service_reload to the global variables.

    document service_reload parameter in README.md

    Add the service_reload parameter to the freshly added Suse OS.

    undef is a keyword, not a string.
    remove the firewall_supported parameter from OpenBSD too, as
    it got removed from original master.

    pick function doesn't work with undef, so don't use it,
    and fix the default package name for the perl package.
    Both just for OpenBSD.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -173,9 +173,9 @@ class postgresql::params inherits postgresql::globals {
       $user                = pick($user, '_postgresql')
       $group               = pick($group, '_postgresql')
 
-      $client_package_name  = pick($client_package_name, "postgresql-client")
-      $server_package_name  = pick($server_package_name, "postgresql-server")
-      $contrib_package_name = pick($contrib_package_name, "postgresql-contrib")
+      $client_package_name  = pick($client_package_name, 'postgresql-client')
+      $server_package_name  = pick($server_package_name, 'postgresql-server')
+      $contrib_package_name = pick($contrib_package_name, 'postgresql-contrib')
       $devel_package_name   = pick($devel_package_name, 'postgresql-client')
       $java_package_name    = pick($java_package_name, 'postgresql-jdbc')
       $perl_package_name    = pick($perl_package_name, 'databases/p5-DBD-Pg')
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -3,7 +3,7 @@ class postgresql::params inherits postgresql::globals {
   $version                    = $globals_version
   $postgis_version            = $globals_postgis_version
   $listen_addresses           = 'localhost'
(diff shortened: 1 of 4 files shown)


buzzdeee:
Anything I can do further to get this merged?

manifests/params.pp
```diff
@@ -3,7 +3,7 @@
   $version                    = $globals_version
   $postgis_version            = $globals_postgis_version
   $listen_addresses           = 'localhost'
-  $port                       = 5432
+  $port                       = '5432'
```
igalic: this will make parser=future unhappy…

cmurphy: Agreed, plus this seems unnecessary and unrelated to the rest of the changes.

buzzdeee: I do run the parser=future all the time. But since October, some time passed by, maybe that was when I had Puppet 3.7.1 or 3.7.2 running, and where I got an error without the ''. I just tested again, and with and without them, no problem anymore.
I just removed them.

thanks.

manifests/params.pp
```diff
@@ -165,6 +168,31 @@
...
+      $needs_initdb         = pick($needs_initdb, true)
+    }
+
+    'OpenBSD': {
+      $user                = pick($user, '_postgresql')
+      $group               = pick($group, '_postgresql')
+
+      $client_package_name  = pick($client_package_name, "postgresql-client")
+      $server_package_name  = pick($server_package_name, "postgresql-server")
+      $contrib_package_name = pick($contrib_package_name, "postgresql-contrib")
```
igalic: this here will also give puppet-lint warnings, can you please `s/"/'/`?

igalic:
otherwise we should be fine for merging this.

buzzdeee:
Hope its fine now.

igalic:
@buzzdeee yesno: this will need a proper [rebase](http://git-scm.com/book/en/Git-Branching-Rebasing). not sure if that's caused by the Merge branch 'master'...  commit, it's easier to do rebasing when you use the [github flow](https://guides.github.com/introduction/flow/index.html) work-flow for Pull Requests.

we'd also prefer this to be [squashed down](http://www.git-scm.com/book/en/Git-Tools-Rewriting-History#Squashing-Commits) to one commit.

mhaskel:
Thanks @buzzdeee, this was merged in #573

PR #523, opened 2014-11-04
set postgres version on Ubuntu 14.10

avelino:

PR #524, opened 2014-11-04
set postgres version on Ubuntu 14.10

avelino:


mhaskel:
@avelino thanks for the contribution, but in general you should only submit pull requests against master, so I'm going to close this and respond to #523. We don't have any planned releases for postgresql 3.x.

PR #528, opened 2014-11-12
Fix $service_status in postgresql::params for 'RedHat', 'Linux' and 'Archlinux'

franzs:
Hi!

I experienced some problems restarting postgresql due to the empty variable $postfix::params::service_status on CentOS. I fixed setting the variable where it was missing.

Please have a look at it.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -57,7 +57,7 @@ class postgresql::params inherits postgresql::globals {
       }
       $psql_path            = pick($psql_path, "${bindir}/psql")
 
-      $service_status      = $service_status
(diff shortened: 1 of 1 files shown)


tphoney:
Based on @igalic comment, could you re-base this and we will try and merge it.

mhaskel:
@franzs thanks for the contribution, and sorry for the delay in getting back to this. From our testing though, if $service_status is `undef` puppet will use `service <whatever> status` as the default. If this is still an issue you're seeing and something you'd like to have added to the module, could you provide more details into the problems you're having and resubmit? thanks again!

PR #530, opened 2014-11-14
Using enviromental varables to hide actual password in psql commands

cardil:
### Motivation

Executing commands by Puppet is crucial, because the most concerns production environments. I wanted to avoid displaing password, even in the form of "encrypted" in the logs and reports. These reports are usually stored for a long time, making it possible vector of attack.

### Design

I decided to design it in such a way as to take advantage of the fact that the environment variables are not displayed by the `Puppet::Util::Execution` - even in debug mode. It displays actual command, in this case: 
```shell
/usr/bin/psql -d base -t -c "SELECT 'something' as \"Custom column\""
```

### Development

1. Edited `postgresql_psql` ruby provider, `postgres::server::role` and `postgresql::server::passwd` class
 1. to use environment variables to store actual passwords
 1. to properly escape `"` sign when passing command to `psql` executable
1. Writen a spec test for `postgresql::server::passwd` and `postgres::server::role`
1. Corrected spec test to `postgresql_psql` ruby provider

### Testing

I've tested those changes by running spec tests on:

* rvm: 1.8.7, env: PUPPET_GEM_VERSION="~> 2.7.0" FACTER_GEM_VERSION="~> 1.6.0"
* rvm: 1.9.3, env: PUPPET_GEM_VERSION="~> 3.5" STRICT_VARIABLES="yes"

I've also tested those changes against a vagrant machine `wavesoftware/trusty64-nocm` using both of settings from `.travis.yaml` with:

```shell
tail -f /var/log/syslog | grep postgres &
puppet apply test.pp --debug
```
a test.pp contents are:

```puppet
include postgresql::server
postgresql::server::db { 'mydatabasename':
  user     => 'mydatabaseuser',
  password => postgresql_password('mydatabaseuser', 'mypassword'),
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 pkg/
 Gemfile.lock
 vendor/
+.vendor/
 spec/fixtures/
 .vagrant/
 .bundle/
--- a/lib/puppet/provider/postgresql_psql/ruby.rb
+++ b/lib/puppet/provider/postgresql_psql/ruby.rb
@@ -14,7 +14,7 @@ Puppet::Type.type(:postgresql_psql).provide(:ruby) do
     command = [resource[:psql_path]]
     command.push("-d", resource[:db]) if resource[:db]
     command.push("-p", resource[:port]) if resource[:port]
-    command.push("-t", "-c", sql)
+    command.push("-t", "-c", '"' + sql.gsub('"', '\"') + '"')
 
     if resource[:cwd]
       Dir.chdir resource[:cwd] do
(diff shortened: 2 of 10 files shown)


hunner:
Manually merged

cardil:
Nice. Thanks!

antaflos:
This seems to break setting the Postgres password!

I observe that when defining a `postgresql::server::password` of `foobar` it sets the password literally to `$$foobar$$`. The `unless` clause now fails because the test doesn't work any more:

```
postgres@db01:~$ PGPASSWORD='foobar' psql -h localhost -p 5432 -c 'select 1'
psql: FATAL:  password authentication failed for user "postgres"
FATAL:  password authentication failed for user "postgres"
```

But when doing this it works:

```
postgres@db01:~$ PGPASSWORD='$$foobar$$' psql -h localhost -p 5432 -c 'select 1'
 ?column? 
----------
        1
(1 row)
```

I believe the definition of the `NEWPASSWD_ESCAPED` environment variable must not include the single quotes, instead the variable should be single-quoted inside the `command` exec string.

Currently testing that.

antaflos:
I have created an issue: https://tickets.puppetlabs.com/browse/MODULES-1869

PR #531, opened 2014-11-18
properly quoted the default port parameter

revhazroot:


tphoney:
This does not really add anything.
 https://docs.puppetlabs.com/guides/style_guide.html#namespacing-variables
Looking at other modules Apache does not quote fixnums. 
re-open if you feel strongly about this.

PR #535, opened 2014-11-21
Fixing autodetected version for Debian Jessie, which should use postgresql 9.4

Zouuup:


mmoll:
BTW, this would solve MODULES-1603

mfn:
Are there plans for a new release? Would be great to have 9.4 support in the module available.

igalic:
@mfn: you could propose a release yourself, like so: https://github.com/puppetlabs/puppetlabs-postgresql/pull/478

mfn:
Wouldn't [pg_hba_rule.php](https://github.com/puppetlabs/puppetlabs-postgresql/blob/b0f2a8f765f4bf05950e15700dfaff93f569936c/manifests/server/pg_hba_rule.pp#L28) also need an entry for 9.4?

igalic:
@mfn looking at this code, it should work with the default, aaand, we should fix it so it works better.
There's way too much repetition, and, i have no idea what, exactly, that default is supposed to be.

PR #536, opened 2014-11-24
add option to make postgresql.conf unmanaged

CharlieR-o-o-t:
add option to make postgresql.conf file unmanaged by puppet. 
at this time it's impossible to use own config file.


igalic:
@frconil to be precise, `postgresql_conf` is its own type, which parses/inserts into the file.
it's not based on augeas.

mhaskel:
@CharlieR-o-o-t thanks for the contribution, but closing since it sounds like this isn't necessary from the comments.

PR #540, opened 2014-12-05
Fix/encoding

mcanevet:


mcanevet:
@hunner @mhaskel @cmurphy could you please review/comment/merge this PR?

cmurphy:
What's the rationale for putting this in postgresql::server::initdb instead of, say, postgresql::server::config? Any why make it an elsif with $needs_initdb? Wouldn't there be a case where $needs_initdb was true and also the encoding needed to be fixed?

cmurphy:
Discussed at puppet contributors summit, lgtm. Will just wait on TravisCI.

igalic:
@cmurphy travis fails for 2.7, which imo can be ignored >_>

mcanevet:
@cmurphy could you please merge this if it's OK for you also?

cmurphy:
@mcanevet the travis errors can be fixed by rebasing against master, where we pinned rspec-core to fix the issue on puppet 2.7/ruby 1.8.7.

@igalic I disagree with ignoring 2.7 errors. We plan to remove support for puppet 2.7 on the next naturally occurring major release, but while we still officially support it we need to keep CI green on it.

mcanevet:
@cmurphy Travis still fails, but because of this : https://github.com/puppetlabs/puppetlabs-concat/pull/270 this time...

cmurphy:
Ignoring the strict variables failures for now. Thanks!

PR #543, opened 2014-12-08
(MODULES-1383) Add restart parameter

juniorsysadmin:
This commit adds a restart parameter to both postgresql::server and postgresql::server::config_entry.

When set to true, a restart of the postgresql service will be preferred when changing config settings, rather than a reload. The parameter in server::config_entry takes the value set in postgresql::server by default, but can be set independently.

I'm open to suggestions in making the code less kludgy.
Should more rspec tests be added?

### DIFF
--- a/README.md
+++ b/README.md
@@ -393,6 +393,9 @@ Value to pass through to the `package` resource when creating the server instanc
 ####`plperl_package_name`
 This sets the default package name for the PL/Perl extension. Defaults to utilising the operating system default.
 
+####`restart`
+Whether the service should be restarted when things change. Defaults to `false`.
+
 ####`service_manage`
 This setting selects whether Puppet should manage the service. Defaults to `true`.
 
@@ -567,6 +570,12 @@ Name of the setting to change.
 
 ####`ensure`
 Set to `absent` to remove an entry.
+
+####`restart`
+Whether the service should be restarted when things change. Defaults to
+`false`.
+Note that the service will still be restarted if the setting requires one in
+order for the change to take effect.
 
 ####`value`
 Value for the setting.
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -10,6 +10,7 @@ class postgresql::params inherits postgresql::globals {
   $ipv6acls                   = []
   $encoding                   = $postgresql::globals::encoding
   $locale                     = $postgresql::globals::locale
+  $restart                    = false
   $service_ensure             = 'running'
   $service_enable             = true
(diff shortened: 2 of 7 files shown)


hunner:
What is the motivation for this? When would you always want a restart, instead of reloading?

bradq:
@hunner @juniorsysadmin: The motivation was my own, though Ben actually put in the work! This stems from my original report, https://tickets.puppetlabs.com/browse/MODULES-1383. I originally intended quite the opposite (should avoid reloading automatically unless requested...in general I prefer that unless otherwise indicated, service refreshes be manual and orchestrated outside Puppet), but I can't disagree with this change either. 

In regards to reload/restart, I occasionally have config changes that require a full restart (A new SSL cert, a new listen address, etc) on systems for development where I'd rather not bother with manual restarts, knowing that it'll be in sync eventually. Full list of config changes that require a server kick are here: http://www.postgresql.org/docs/9.4/static/runtime-config-connection.html (anything referencing "server start".) Note that the "restart" vs "reload" as submitted is configurable per-config entry as well, something I really like. 

I'd suggest simply setting a top-level param in manifests/server.pp that takes the style of restart expected for a given system and defaults to reload. Unless you think this is a poor design pattern I'm amenable to cleaning this up, adding my end of things, and updating @juniorsysadmin's repo to make the system work for everything in this PR and the original issue.

hunner:
This is how you can do a full restart when certain config entries are changed: https://github.com/puppetlabs/puppetlabs-postgresql/blob/4.1.0/manifests/server/config_entry.pp#L18-L31

bradq:
@hunner Thanks for pointing me in the right way, and apologies for the misdirection; I answered the question rather than the problem. I also didn't review @juniorsysadmin's code very well, so...sorry for wasting everyone's time here. A solid 6 months have passed since I ran into this, and I did indeed forget the details...but in essence, a poorly thought out config change, in conjunction with a poorly implemented architecture, yielded a bit of downtime for us when that full-restart class was triggered. 

This did at least remind me of the specific issue that I *did* have with this...being that both reload and restart should offer booleans to make starting/restarting optional, rather than enforcing them blindly. In general I don't particularly like the idea of effecting DB restarts on production databases due to a config file change...prefer to stage those and document via Puppet but handle manually. I'll send it in in another PR unless I'm again misreading things or there's an ideological reason for this.

mhaskel:
Ok, thanks @juniorsysadmin @bradq . Closing this now due to comments.

PR #545, opened 2014-12-09
Postgis dep

timdaman:
Prevent pick() of postgis_version failure

globals.pp always trys to identify the default what version of postgis even when postgis is not being installed. 

The problem comes when using a custom packaged version of postsql that does not match the default major.minor version of postgresql for the current OS. In that case default_postgis _version is set to undef If postgis_version is also undefined the pick statement tthat set globals_postgis_version fails.

This changes the pick statement in globals.pp to an if statment than can pass undefs safely.
It changes params.pp to pass undef as a package name if postgis_version is equal to undef.
Finally I changed the postgis.pp class to give a meanful error message when no valid package name can be found.

I performed simple testing of this code but I lack the facilities to fullly exercise it. Additionally I do not know rspec. As a result I a submitting the code for a pull request right now incase someone finds it useful or wishes to add the missing test. I plan to learn rspec in the next couple months as I need it for a project and at that time, if the pull request is still waiting, I will add tests then.

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -106,7 +106,11 @@ class postgresql::globals (
     '93'    => '2.1',
     default => undef,
   }
-  $globals_postgis_version = pick($postgis_version, $default_postgis_version)
+  if $postgis_version {
+    $globals_postgis_version = $postgis_version
+  } else {
+    $globals_postgis_version = $default_postgis_version
+  }
 
   # Setup of the repo only makes sense globally, so we are doing this here.
   if($manage_package_repo) {
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -63,15 +63,18 @@ class postgresql::params inherits postgresql::globals {
       $perl_package_name   = pick($perl_package_name, 'perl-DBD-Pg')
       $python_package_name = pick($python_package_name, 'python-psycopg2')
 
-      $postgis_package_name = pick(
-        $postgis_package_name,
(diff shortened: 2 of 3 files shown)


manifests/globals.pp
```diff
@@ -106,7 +106,11 @@
     '93'    => '2.1',
     default => undef,
   }
-  $globals_postgis_version = pick($postgis_version, $default_postgis_version)
+  if $postgis_version {
```
cyberious: What is the reason for the rewrite of this?  This is the exact what pick is for.

manifests/globals.pp
```diff
@@ -106,7 +106,11 @@
...
+    $globals_postgis_version = $postgis_version
+  } else {
+    $globals_postgis_version = $default_postgis_version
+  }
```
daenney: You just reimplemented the behaviour of `pick` in an `if` statement.

manifests/globals.pp
cyberious: Since we assume later on that the minimal version is 2  we should default to that as a nested pick, but we fail if we don't get it either way.

hunner:
Closed because `$postgis_package_name` has a `validate_string()` at https://github.com/puppetlabs/puppetlabs-postgresql/blob/07b8ac4946c1af625b3cec562fa282091d566f7d/manifests/server/postgis.pp#L6 and thus cannot be undef, so pick should be fine. Feel free to reopen it if this is incorrect.

PR #546, opened 2014-12-09
Allow management of logdir

cwood:
Simple request to allow management of the logdir when installing and setting up postgres for the first time. 

I didn't find a way to do this when changing the log dir and on a initial install of postgres it will fail with this class since the log_directory isn't installed.

Sorry if there is easier way to do this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Table of Contents
 2. [Module Description - What does the module do?](#module-description)
 3. [Setup - The basics of getting started with PostgreSQL module](#setup)
     * [PE 3.2 supported module](#pe-32-supported-module)
-    * [Configuring the server](#configuring-the-server) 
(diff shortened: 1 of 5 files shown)


cwood:
Any update on this PR?

igalic:
@cwood can you fix the issues that @hunner pointed out?

cwood:
Just updated

igalic:
@cwood can you please [squash](http://www.git-scm.com/book/en/Git-Tools-Rewriting-History#Squashing-Commits) those 5 commits down to one?

mhaskel:
@cwood ping?

cwood:
Squashed

hunner:
This was merged in 0f4c718

cwood:
yay thanks @hunner :)

PR #547, opened 2014-12-25
Add support for sequence grants

mukaibot:
I needed to be able to grant at the sequence level for my project, so I uncommented the SEQUENCE type and added privilege validation in the same form as the other grant types. Like table and database grant, I also added a wrapper around sequence grant.
Please let me know if this is desirable to be accepted into mainline, and what further changes I should make.

### DIFF
--- a/manifests/server/grant.pp
+++ b/manifests/server/grant.pp
@@ -25,7 +25,7 @@ define postgresql::server::grant (
     #'FUNCTION',
     #'PROCEDURAL LANGUAGE',
     'SCHEMA',
-    #'SEQUENCE',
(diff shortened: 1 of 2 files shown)


igalic:
@mukaibot this will need tests (spec tests, unit tests, and possibly an acceptance tests)
it will also need an update to the README

igalic:
@mukaibot ping

mhaskel:
Thanks for the contribution @mukaibot, but we haven't heard from you in a while so we're closing this. If this is functionality you still want included in the modules, please address the comments in the pull request and resubmit.

PR #550, opened 2015-01-09
Add local_auth_option_postgres to modify postgres and user/database pg_hba rules to be able to pass auth-options like custom maps.

kostecky:
If you specify it, it will append a pg_hba auth-option to the postgres and user/database rules so you can apply auth-options like custom maps

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -10,6 +10,7 @@ class postgresql::server::config {
   $pg_ident_conf_path         = $postgresql::server::pg_ident_conf_path
   $postgresql_conf_path       = $postgresql::server::postgresql_conf_path
   $pg_hba_conf_defaults       = $postgresql::server::pg_hba_conf_defaults
+  $local_auth_option_postgres = $postgresql::server::local_auth_option_postgres
   $user                       = $postgresql::server::user
   $group                      = $postgresql::server::group
   $version                    = $postgresql::server::_version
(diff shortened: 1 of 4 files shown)


kostecky:
Is there any opinion on this modification and is there willingness to merge this?

igalic:
i think it's probably useful, but we haven't seen great engagement from the community so far.

if it is to be merged, it will need spec tests.

kostecky:
Ah ok, I didn't see a spec test for some of the other override options, so
I didn't know if it was required. I can write one up however.

Kristian Kostecky
VM Farms
http://vmfarms.com
kris@vmfarms.com
1 (866) 278-0021 x702
Twitter: @vmfarms

kostecky:
Actually, considering that there is no precedence set for augmenting the current spec for these override options, is there a particular pattern/format that you'd like to see? How would you like to see this done?

mhaskel:
@kostecky could you rename this variable to `$local_auth_option`?

manifests/server/config.pp
```diff
@@ -35,8 +36,8 @@
 
       # Lets setup the base rules
       $local_auth_option = $version ? {
-        '8.1'   => 'sameuser',
-        default => undef,
+        '8.1'   => "sameuser ${local_auth_option_postgres}",
```
bmjen: If `local_auth_option_postgres` is set to `undef`, then `'8.1'` will be set to `"sameuser "` vs `"sameuser"`.

hunner:
Closed due to inactivity. Feel free to reopen it and rebase and address the comments if you would still like it merged.

PR #551, opened 2015-01-11
(MODULES-1153) Add database comment parameter

juniorsysadmin:
This commit adds a comment parameter to postgresql::server::db so that one can add a database description using PostgreSQL's COMMENT command.

Refer to the following links for documentation on the queries used:
http://www.postgresql.org/docs/8.1/static/sql-comment.html
http://stackoverflow.com/questions/11493978/how-to-retrieve-the-comment-of-a-postgresql-database
http://www.postgresql.org/docs/current/interactive/functions-info.html#FUNCTIONS-INFO-COMMENT-TABLE

### DIFF
--- a/spec/acceptance/db_spec.rb
+++ b/spec/acceptance/db_spec.rb
@@ -10,6 +10,7 @@ describe 'postgresql::server::db', :unless => UNSUPPORTED_PLATFORMS.include?(fac
           location => '#{tmpdir}',
         } ->
         postgresql::server::db { 'postgresql_test_db':
+          comment    => 'testcomment',
           user       => 'test',
           password   => 'test1',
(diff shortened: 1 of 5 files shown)


igalic:
"+1:

spec/acceptance/db_spec.rb
```diff
@@ -10,6 +10,7 @@ class { 'postgresql::server': }
           location => '#{tmpdir}',
         } ->
         postgresql::server::db { 'postgresql_test_db':
+	  comment    => 'testcomment',
```
cmurphy: Could you fix the indentation here?

juniorsysadmin:
The comment indentation has been fixed.

PR #553, opened 2015-01-13
Replace require by include

PierreR:
Avoid using the `require` function (which is IMHO a bad practice) by being more explicit in the `require` attribute value.


igalic:
:+1:
***
ping @cmurphy: this seems to be trivial enough to be merged.

PR #554, opened 2015-01-24
FreeBSD PostgreSQL wouldn't start first time

leepa:
It would be enabled, but it wouldn't work properly. This fixes that
issue the same way Puppet does itself; use the onestart/onestop and most
importantly in this case, the onestatus command.

By using this approach it means the Database server will actually start
whereas it would not before. It would enable, but not actually start.
onestatus means the service type gets the right response and behaves
properly.


igalic:
for reference (from [rc.subr(8)](http://www.freebsd.org/cgi/man.cgi?query=rc.subr&sektion=8)):
```
one	Skip the checks	for rcvar being	set to ``YES'',	but
    performs all the other prerequisite tests.
```
:+1: from me.

PR #555, opened 2015-01-29
change example of postgresql::server::role definition to version 3

phaf:
The example shown in the Readme refers to module version 2. Since version 3 roles are renamed to postgresql::server::roles


igalic:
:+1:

PR #557, opened 2015-02-04
Require server package before user permissions

hunner:
The postgresql user is created by the server package, but this file
resource may be evaluated before the package is installed resulting in
permission failures.


igalic:
suddenly realizing how happy i am that postgresql only has one main distributor…

PR #560, opened 2015-02-06
Allowing validation of db connection for more than one user and port

uuid0:
I came across the problem that i couldn't validate the db connections for two different users on the same host, due to the namevar of the exec in validate_db_connection. I've added username and port to the name.

### DIFF
--- a/manifests/validate_db_connection.pp
+++ b/manifests/validate_db_connection.pp
@@ -47,7 +47,7 @@ define postgresql::validate_db_connection(
   # time it takes to run each psql command.
   $timeout = (($sleep + 2) * $tries)
 
-  $exec_name = "validate postgres connection for ${database_host}/${database_name}"
+  $exec_name = "validate postgres connection for ${database_username}@${database_host}:${database_port}/${database_name}"
   exec { $exec_name:
     command     => "echo 'Unable to connect to defined database using: ${cmd}' && false",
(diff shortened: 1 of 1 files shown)


igalic:
sounds reasonable. :+1:

jonnytpuppet:
I could be reading the code wrong, but is there an issue here where the variables that are part of the exec name to be undef?

igalic:
it's just a name, so in theory, unless you somehow manage to duplicate it by leaving stuff out, it shouldn't matter

mhaskel:
@igalic I could be wrong, but sounds like @jonnytpuppet might be concerned about strict variables with the future parser. Are those variables guaranteed to be 'defined' in the strict vars sense (ie, with `$var = undef` `$var` is defined)

manifests/validate_db_connection.pp
```diff
@@ -47,7 +47,7 @@
   # time it takes to run each psql command.
   $timeout = (($sleep + 2) * $tries)
 
-  $exec_name = "validate postgres connection for ${database_host}/${database_name}"
+  $exec_name = "validate postgres connection for ${database_username}@${database_host}:${database_port}/${database_name}"
```
hunner: `database_username` and `database_port` default to undef. We don't want to end up with stray `@` and `:` symbols in the exec

PR #561, opened 2015-02-06
doc links should point to /current/ rather than a specific version

igalic:


igalic:
ping @cmurphy: as discussed at cfgmgmtcamp summit.

PR #564, opened 2015-02-17
Make granting on ALL TABLES IN SCHEMA idempotent

antaflos:
Define a proper SELECT statement to feed into Postgresql_psql's
`unless` parameter that checks if there are any tables in the specified
schema for which the specified role *does not* have the specified
privilege. Only then allow the GRANT statement to be executed. For
details see comments.

Note that this, too, suffers from the problem that there is no feasible
way to check if a role has ALL PRIVILEGES on a table in plain SQL. By
terrible convention the INSERT privilege represents ALL PRIVILEGES here.

### DIFF
--- a/manifests/server/grant.pp
+++ b/manifests/server/grant.pp
@@ -75,10 +75,45 @@ define postgresql::server::grant (
       $on_db = $db
     }
     'ALL TABLES IN SCHEMA': {
-      validate_string($_privilege, 'SELECT', 'INSERT', 'UPDATE', 'REFERENCES',
-        'ALL', 'ALL PRIVILEGES')
-      $unless_function = false # There is no way to test it simply
+      validate_string($_privilege,'SELECT','INSERT','UPDATE','DELETE',
+        'TRUNCATE','REFERENCES','TRIGGER','ALL','ALL PRIVILEGES')
+      $unless_function = 'custom'
       $on_db = $db
+
+      $schema = $object_name
+
+      # Again there seems to be no easy way in plain SQL to check if ALL
+      # PRIVILEGES are granted on a table. By convention we use INSERT
+      # here to represent ALL PRIVILEGES (truly terrible).
+      $custom_privilege = $_privilege ? {
+        'ALL'            => 'INSERT',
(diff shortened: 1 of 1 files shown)


antaflos:
Since I am obviously no DBA that can perform powerful magic with SQL the SELECT statement I devised may not be optimal. But it certainly is better than having the GRANT statement executed on every single Puppet run, which is the current situation.

antaflos:
The failed Travis checks say things like:
> Undefined variable "::is_pe"; Undefined variable "is_pe" at

I don't think this has anything to do with my changes here.

manifests/server/grant.pp
```diff
@@ -103,8 +138,9 @@
   }
 
   $_unless = $unless_function ? {
-      false   => undef,
-      default => "SELECT 1 WHERE ${unless_function}('${role}',
+      false    => undef,
+      'custom' => $custom_unless,
+      default  => "SELECT 1 WHERE ${unless_function}('${role}',
```
igalic: does this end up being
```sql
select 1 where select 1 from ( select …
```
is that correct?

raphink: The `is_pe` bit is caused by recent changes in puppetlabs-concat. The fix is at https://github.com/puppetlabs/puppetlabs-concat/pull/270

antaflos: @igalic, `$custom_unless` contains the SQL query that is passed to `Postgresql_psql` (via `$_unless`). It will be executed as is by `Postgresql_psql`, just like the SQL query constructed in the default case here.

igalic: ack ack ack, didn't see `$custom_privilege` variable name, and thought it was assigned to `$_unless` as `$unless_function`

antaflos:
FWIW, we have this change deployed in our production environments and it works fine.

antaflos:
@cmurphy I am being forward and pinging you so that this PR possibly gets some attention :)

igalic:
@tphoney pinging you also!

raphink:
Great, thanks!

PR #565, opened 2015-02-23
create role before database

gerhardsam:
Creating the role before creating the database is mandatory because otherwise
granting the access right to the database fails.

### DIFF
--- a/manifests/server/db.pp
+++ b/manifests/server/db.pp
@@ -29,6 +29,7 @@ define postgresql::server::db (
   if ! defined(Postgresql::Server::Role[$user]) {
     postgresql::server::role { $user:
       password_hash => $password,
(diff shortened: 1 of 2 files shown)


igalic:
:+1:

AnthonyMastrean:
:+1:

bemosior:
I believe we experience the same issue:
`Postgresql::Server::Database[fakedb]/Exec[/usr/pgsql-9.0/bin/createdb --port='5432' --owner='fakeowner' --template=template0  'fakedb']/returns: createdb: database creation failed: ERROR:  role "fakeowner" does not exist`

After applying these changes, the issue was resolved.

invalidusrname:
Thanks!

PR #567, opened 2015-02-27
Use correct TCP port when checking password

antaflos:
The PostgreSQL server's port is specified in `postgresql::server::port`, so use it in the `unless` clause of `Exec['set_postgres_postgrespw']`. Failing to do so results in repeated invocations of the exec resource during Puppet runs when the server's port is not set to the default of 5432.


igalic:
i was gonna say tests, but this is a private class ;)

PR #568, opened 2015-03-03
(MODULES-661) Add host parameter to psql commands

petems:
This allows you to use `postgres_pqsl` commands to be run against a specified host. The ultimate goal being that you can use `puppetlabs-postsgresql` to run against remote instances, especially things such as Amazons RDS service.

This PR will be further extended to do things like allow remote `table_grant` and the like.

### DIFF
--- a/lib/puppet/provider/postgresql_psql/ruby.rb
+++ b/lib/puppet/provider/postgresql_psql/ruby.rb
@@ -14,6 +14,7 @@ Puppet::Type.type(:postgresql_psql).provide(:ruby) do
     command = [resource[:psql_path]]
     command.push("-d", resource[:db]) if resource[:db]
     command.push("-p", resource[:port]) if resource[:port]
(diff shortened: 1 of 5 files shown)


mhaskel:
:+1: thanks @petems

ColinHebert:
Unless I'm mistaken, this still assumes that there is a user (defined with `:psql_user`) on the machine with passwordless access to the database.

petems:
@ColinHebert Yes, right now it was a very MVP implementation, I'm breaking up the larger "Use the postgres module to fully provision a remote instance" into smaller chunks.

@tomhey's done some awesome work around adding a new `connection_options` hash using Postgres Environment variables (http://www.postgresql.org/docs/current/static/libpq-envars.html) which makes this a lot better, and is probably the best approach to properly implement it. Discussion's here if you're interested https://tickets.puppetlabs.com/browse/MODULES-661

PR #571, opened 2015-03-04
(MODULES-661) Allow remote grants

petems:
You can now create remote databases, grants, and roles

Not sure about the best way to test this, going to do some beaker tests to run against localhost.

I assume undef's are a bad thing in future parser too? I can change that if needed :)

### DIFF
--- a/manifests/server/database.pp
+++ b/manifests/server/database.pp
@@ -7,7 +7,8 @@ define postgresql::server::database(
   $template   = 'template0',
   $encoding   = $postgresql::server::encoding,
   $locale     = $postgresql::server::locale,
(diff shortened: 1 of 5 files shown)


mhaskel:
Hi @petems! `undef`s should be fine, but yeah, will need test and doc updates

igalic:
i'm a big fan of `undef`s, https://github.com/puppetlabs/puppetlabs-concat/pull/270#issuecomment-75937714

mhaskel:
@petems ping?

petems:
This is dependant on the hash conncection settings by @tomhey, so I'll close this :+1:

PR #575, opened 2015-03-09
Remove dependency on puppetlabs/apt since it is not required.

ghoneycutt:
Section 8.1 of the Puppet Labs Style Guide states that soft dependencies
must not be enforced as a hard requirement in the metadata. This will
allow users who do not use systems with apt to no longer manage the
puppetlabs/apt module and its dependencies.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -68,7 +68,6 @@
   ],
   "dependencies": [
     {"name":"puppetlabs/stdlib","version_requirement":"4.x"},
-    {"name":"puppetlabs/apt","version_requirement":">=1.1.0 <2.0.0"},
     {"name":"puppetlabs/concat","version_requirement":">= 1.1.0 <2.0.0"}
   ]
 }


igalic:
:+1:

daenney:
This module does have an explicit dependency, not a "soft" dependency.

ghoneycutt:
@cyberious please merge this

It is a soft dependency as the apt class is called conditionally. If you are on an EL system, you will not call the apt class and so this dependency is not needed in all cases to provide the core functionality.

ghoneycutt:
As someone who uses this module on an EL6 system (which is the primary platform in the enterprise), there is no reason why I should have to maintain a module for apt and its dependencies.

ghoneycutt:
@mhaskel Could you please take a look at this?

mhaskel:
@ghoneycutt sorry for the delay in response here, but we can't get rid of the apt dependency. It is not actually a soft dependency (since it's used in the manifests). There's no way to specify conditional dependencies in the metadata, which is where we'd need to be able to fix this.

ghoneycutt:
@mhaskel The style guide describes a soft dependency as "A soft dependency is a dependency that is only required in a specific set of use cases." The apt module is only required in the specific use case that your system uses apt and is not integral to postgresql itself. Perhaps we are having different views on what a soft dependency is.

mhaskel:
@ghoneycutt yeah. We only consider it to actually be a soft dependency if resources from the dependency aren't used in any of the manifests, so something more like [this](https://github.com/puppetlabs/puppetlabs-rabbitmq#module-dependencies) is what we mean for a soft dependency.

PR #576, opened 2015-03-09
4.2.0 prep

mhaskel:


tomhey:
The "Add host parameter to `postgresql_psql` type (MODULES-661)" ticket is still under active discussion, I'd like to get this release to include the stable API change. Rather than adding a new parameter that may very well be deprecated in the next release.

What sort of timescale are you looking at/ expecting for the 4.2.0 release?

(Maybe slightly controversially, if we don't have time to complete the API discussion for MODULE-661 then I'd prefer it's commit to be reverted and skipped for this release)

mhaskel:
@tomhey the release is going out tomorrow.

mhaskel:
^^ @petems could you give a :+1: :-1: for including host in a release going out tomorrow, since you implemented that?

petems:
I'm ok with it being yanked, with the view to it being eventually re-implemented using the connection_hash @tomhey mentioned, saves having something that would be quickly deprecated.

:-1:

PR #579, opened 2015-03-10
(MODULES-1834) Be less strict when changing template1 encoding

sathieu:
We currently use: UTF8 != UTF-8
We should instead use: pg_char_to_encoding(UTF8) == pg_char_to_encoding(UTF-8)

### DIFF
--- a/manifests/server/initdb.pp
+++ b/manifests/server/initdb.pp
@@ -94,7 +94,7 @@ class postgresql::server::initdb {
         SET encoding = pg_char_to_encoding('${encoding}'), datistemplate = TRUE
         WHERE datname = 'template1'",
       unless  => "SELECT datname FROM pg_database WHERE
-        datname = 'template1' AND pg_encoding_to_char(encoding) = '${encoding}'",
+        datname = 'template1' AND encoding = pg_char_to_encoding('${encoding}')",
     }
   }
 }


antaflos:
:+1: from me!

igalic:
cool!

manifests/server/initdb.pp
```diff
@@ -94,7 +94,7 @@
         SET encoding = pg_char_to_encoding('${encoding}'), datistemplate = TRUE
         WHERE datname = 'template1'",
       unless  => "SELECT datname FROM pg_database WHERE
-        datname = 'template1' AND pg_encoding_to_char(encoding) = '${encoding}'",
+        datname = 'template1' AND encoding = pg_char_to_encoding('${encoding}')",
```
igalic: i'm guessing that postgresql already normalises encoding, which is why you're not passing both of those to the function, as per your comment/commit message, i.e.:
```sql
datname = 'template1' AND pg_encoding_to_char(encoding) = pg_char_to_encoding('${encoding}')
```

sathieu: encoding is an integer here. No need to normalize it.

Example:
```
postgres=# SELECT encoding from pg_database;
 encoding 
----------
        6
        6
        6
        6
        6
        6
        6
```

igalic: ah. duh ._.

mhaskel:
Thanks @sathieu !

PR #580, opened 2015-03-10
Replication type

raphink:
This provides a type to setup replication slots on a Postgresql master server.

### DIFF
--- a/lib/puppet/type/postgresql_replication_slot.rb
+++ b/lib/puppet/type/postgresql_replication_slot.rb
@@ -1,0 +1,16 @@
+Puppet::Type.newtype(:postgresql_replication_slot) do
+  @doc = "Manages Postgresql replication slots.
+
+This type allows to create and destroy replication slots
+to register warm standby replication on a Postgresql
+master server.
+"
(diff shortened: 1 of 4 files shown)


tphoney:
Thanks for the work, this needs a little bit of cleanup, Could you add some documentation then make sure the tests pass and squash your commits.

raphink:
@tphoney I fixed the tests & squashed the commits.

The existing types for this module are not documented in README.md. They'll get their doc in the types tab on the forge anyway.

lib/puppet/type/postgresql_replication_slot.rb
```diff
@@ -0,0 +1,9 @@
+Puppet::Type.newtype(:postgresql_replication_slot) do
+  ensurable
```
hunner: Could you add an `@doc` description, similar to https://github.com/puppetlabs/puppetlabs-mysql/blob/master/lib/puppet/type/mysql_user.rb#L3 (well, more verbose than that would be great).

raphink:
@hunner done

PR #583, opened 2015-03-13
Fix the handling of run_unless_sql_command in puppet 4

hunner:
`Puppet::Util::SUIDManager.run_and_capture` (in puppet < 3.4) returns an
array of the output and the `Process::Status` object.
`Puppet::Util::Execution.execute` (in puppet >= 3.4) returns the output
and saves the `Process::Status` object to `$CHILD_STATUS`

This is handled in the provider, but the type was also trying to handle
it, and was not last updated when the provider was.

### DIFF
--- a/lib/puppet/type/postgresql_psql.rb
+++ b/lib/puppet/type/postgresql_psql.rb
@@ -34,12 +34,7 @@ Puppet::Type.newtype(:postgresql_psql) do
 
     # Return true if a matching row is found
     def matches(value)
-      if Puppet::PUPPETVERSION.to_f < 4
-        output, status = provider.run_unless_sql_command(value)
-      else
-        output = provider.run_unless_sql_command(value)
-        status = output.exitcode
-      end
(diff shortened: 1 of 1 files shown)


igalic:
ooh cool!

joshcooper:
@hunner  I've been stuck in meetings all day, so sorry I didn't comment earlier. Fixing the type makes sense. I think the type was trying to take advantage of this change in puppet https://github.com/puppetlabs/puppet/commit/4ef85c5cb3526bdc627fb561c34f95d6d98a5edf, first released in 3.4.0, but the type was mistakenly comparing against 4 instead of 3.4.

If you ever expect the postgres module to run on windows, you'll also want to change the provider to call `ProcessStatus#exitcode` (when puppet version >= 3.4), instead of relying on `$CHILD_STATUS` as the latter can truncate windows exit codes. More details in http://projects.puppetlabs.com/issues/23124.

PR #587, opened 2015-03-17
Fix acceptance tests for EL

mhaskel:
The return value of fact('selinux') is a stringified bool.


spec/acceptance/z_alternative_pgdata_spec.rb
```diff
@@ -4,7 +4,7 @@
 # location properly.
 
 # Allow postgresql to use /tmp/* as a datadir
-if fact('osfamily') == 'RedHat' and fact('selinux') == true
+if fact('osfamily') == 'RedHat' and fact('selinux') == 'true'
```
cmurphy: Does this depend on the version of facter? Does this change with puppet 4?

PR #589, opened 2015-03-19
Fix setting postgres role password

antaflos:
Discussed in https://tickets.puppetlabs.com/browse/MODULES-1869

It seems env variables passed via `exec`'s `environment` parameter must not be single-quoted, otherwise the single-quotes are interpreted literally in the command strings in `command` and `unless`. For a postgres password of `foobar` this leads to the `unless` code trying to use literally `'foobar'` as password, and the `psql` line in `command` setting literally `'$$foobar$$'` as password.

### DIFF
--- a/manifests/server/passwd.pp
+++ b/manifests/server/passwd.pp
@@ -21,8 +21,8 @@ class postgresql::server::passwd {
       logoutput   => true,
       cwd         => '/tmp',
       environment => [
-        "PGPASSWORD='${postgres_password}'",
-        "NEWPASSWD_ESCAPED='${escaped}'",
+        "PGPASSWORD=${postgres_password}",
+        "NEWPASSWD_ESCAPED=${escaped}",
       ],
       # With this command we're passing -h to force TCP authentication, which
       # does require a password.  We specify the password via the PGPASSWORD
--- a/spec/unit/classes/server_spec.rb
+++ b/spec/unit/classes/server_spec.rb
@@ -43,8 +43,8 @@ describe 'postgresql::server', :type => :class do
         'command'     => '/usr/bin/psql -c "ALTER ROLE \"postgres\" PASSWORD ${NEWPASSWD_ESCAPED}"',
         'user'        => 'postgres',
         'environment' => [
-          "PGPASSWORD='new-p@s$word-to-set'",
-          "NEWPASSWD_ESCAPED='$$new-p@s$word-to-set$$'"
+          "PGPASSWORD=new-p@s$word-to-set",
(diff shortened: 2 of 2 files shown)


igalic:
how will that fare if the password contains spaces or "special" characters?

antaflos:
I trust that the `exec` provider handles that as effectively as it handled making the single quotes a literal part of the environment variable values.

The alternative is going back to *not* using environment variables for setting the `postgres` role password. But the current situation of mangling the user-supplied password *and* messing up idempotentcy is unacceptable.

igalic:
hah. yeah.

mhaskel:
Hi @antaflos, could you update the acceptance tests [here](https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/spec/acceptance/postgresql_psql_spec.rb#L117-L152) to test what happens if the password has a space? Thanks!

antaflos:
@mhaskel Would love to, but I can't seem to get the acceptance tests to run: https://gist.github.com/antaflos/d6b3f88b8f1316b94c34

No idea what I am doing wrong, any ideas?

cmurphy:
@antaflos maybe try running `bundle install` and `bundle update` first?

antaflos:
@cmurphy Thanks, but I followed the instructions in README.md and installed the bundle to `vendor` before trying to run the acceptance tests. The results are what I posted in the gist above.

I am trying something else now.

antaflos:
@mhaskel I just now noticed that the acceptance tests you linked are for the `Postgresql_psql` type, which is not used for setting the `postgres` role password (an `Exec` calling `/usr/bin/psql` is used instead). It really has nothing to do with the changes I propose in this PR.

tphoney:
If the current acceptance tests do not test this change, it would be fantastic if you could add an acceptance test for this change.  This would assure that future code merges do not break the excellent work you have done.

igalic:
@antaflos: shouldn't the exec be changed to use the type instead?

antaflos:
So this still needs attention and I haven't forgotten about it. I just find it curious that apparently nobody else sets the postgres role's password via `postgresql::server::postgres_password` O_o

igalic:
haaah

tphoney:
Awesome work !

rogierslag:
Any idea when a new version including this fix will be released?

tphoney:
@mhaskel is it possible to line up this for a release ?

mhaskel:
@rogierslag @tphoney sorry for the delay in response here, but we have this scheduled for release in the next few weeks.

PR #599, opened 2015-03-26
Fix URLs in metadata.json

raphink:
This is a detail, as github redirects to `puppetlabs-postgresql` anyway in the browser, but it fails when using the Github API.


raphink:
@mhaskel @cmurphy this is a very trivial fix

mhaskel:
Thanks @raphink

PR #601, opened 2015-03-27
Unbreak module: Puppet3.7.4, future parser, ruby 2.1.5, OpenBSD

buzzdeee:
Rename environment -> pgenv parameter of lib/puppet/type/postgresql_psql.rb,
and adapt, where it's used.  Travis passes all tests green.

This parameter is only used internally, and not exposed to the users
of the module.

Without this change, I get following error when running puppet agent --test:

...
Info: Loading facts
Info: Loading facts
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Invalid parameter environment on Postgresql_psql[CREATE ROLE puppetdb ENCRYPTED PASSWORD ****] at /etc/puppet/environments/production/modules/postgresql/manifests/server/role.pp:46 on node puppetdb.srv.intern
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run

don't know if it is ruby version, puppet version, future parser... but 'environment' seems to be some kind of reserved word.

Is pgenv OK? I'm open for suggestions for other names, or any other feedback too.

### DIFF
--- a/lib/puppet/provider/postgresql_psql/ruby.rb
+++ b/lib/puppet/provider/postgresql_psql/ruby.rb
@@ -29,7 +29,7 @@ Puppet::Type.type(:postgresql_psql).provide(:ruby) do
 
   def get_environment
     environment = {}
-    if envlist = resource[:pgenv]
+    if envlist = resource[:environment]
       envlist = [envlist] unless envlist.is_a? Array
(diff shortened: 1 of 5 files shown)


buzzdeee:
disregard this one, the problem is something else, will shortly add another pull request

PR #603, opened 2015-03-30
(MODULES-1761) Provide defined resource for managing recovery.conf

dacrome:
Currently there is no resource to use for creating the recovery.conf

This resource can create a recovery.conf for replication with all
currently supported parameters


jonnytpuppet:
Thanks for your contribution! It looks great!

dacrome:
Nice! When will it be released?

Envek:
Hi, guys!

I've found an issue in this PR: if I set `manage_recovery_conf` to `true` in `postgresql::server` and will not provide `postgresql::server::recovery` at all, empty `recovery.conf` file will be created and PostgreSQL will refuse to start.

Does anyone have any ideas how to fix or avoid this? It seems that problem is how `puppetlabs-concat` module works.

More info: https://tickets.puppetlabs.com/browse/MODULES-4276

PR #607, opened 2015-04-01
Repo priority centos

pooh22:
When using a non-default version of postgresql, the yum.postgresql.org repo is needed on Centos6 in order to get the requested version of postgresql and the postgis extension. The priority of the postgresql repository wasn't set, causing at the least the geos dependency to pick the older package found in epel over the correct version found in the yum.postgresql.org repo.
See issue: https://tickets.puppetlabs.com/browse/MODULES-1913
This pull request does the following:
- set the yum repo priority to 10 when using the manage repo option
- document that epel is required in the README and provide a small example to install postgis

### DIFF
--- a/manifests/repo/yum_postgresql_org.pp
+++ b/manifests/repo/yum_postgresql_org.pp
@@ -1,5 +1,7 @@
 # PRIVATE CLASS: do not use directly
-class postgresql::repo::yum_postgresql_org inherits postgresql::repo {
+class postgresql::repo::yum_postgresql_org ( 
+      $repo_priority = 10 
+    ) inherits postgresql::repo {
   $version_parts   = split($postgresql::repo::version, '[.]')
   $package_version = "${version_parts[0]}${version_parts[1]}"
   $gpg_key_path    = "/etc/pki/rpm-gpg/RPM-GPG-KEY-PGDG-${package_version}"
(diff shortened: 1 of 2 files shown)


manifests/repo/yum_postgresql_org.pp
```diff
@@ -1,5 +1,7 @@
 # PRIVATE CLASS: do not use directly
-class postgresql::repo::yum_postgresql_org inherits postgresql::repo {
+class postgresql::repo::yum_postgresql_org (
+      $repo_priority = 10
```
hunner: This is a private class, and thus should have no parameters.

Either the priority should be hard-coded to some value, or made a parameter on postgresql::globals.

hunner:
Thanks for this PR; feel free to reopen it when updating as per my comment.

pooh22:
On 08/18/15 23:38, Hunter Haugen wrote:
> Thanks for this PR; feel free to reopen it when updating as per my comment.
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/puppetlabs/puppetlabs-postgresql/pull/607#issuecomment-132362270

Dear Hunter

you're probably right about where to set the priority in the module. I'm 
not able to work on this anymore, due to a change in jobs...

Cheers

Simon

PR #609, opened 2015-04-10
Antaflos patch 2

hunner:
Yeah, setting the postgres_password is broken.

Closes #589


antaflos:
Awesome, thank you for taking care of the acceptance test! I don't think I would have solved that quite so elegantly :) Thanks! :+1:

PR #614, opened 2015-04-22
Fixed systemd override for manage_package_repo package versions

cdenneen:
This template is only used by [EL7](https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/manifests/server/config.pp#L160-L177) and EL7 is only supported in >= 9.3
I've confirmed 9.3 and 9.4 available packages have this renamed library:
```
# rpm -qlp postgresql93-server-9.3.5-2PGDG.rhel7.x86_64.rpm | grep systemd
/usr/lib/systemd/system/postgresql-9.3.service
# rpm -qlp postgresql93-server-9.3.6-2PGDG.rhel7.x86_64.rpm | grep systemd
/usr/lib/systemd/system/postgresql-9.3.service
# rpm -qlp postgresql94-server-9.4.0-1PGDG.rhel7.x86_64.rpm |grep systemd
/usr/lib/systemd/system/postgresql-9.4.service
# rpm -qlp postgresql94-server-9.4.1-1PGDG.rhel7.x86_64.rpm |grep systemd
/usr/lib/systemd/system/postgresql-9.4.service
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,5 @@ matrix:
     env: PUPPET_GEM_VERSION="~> 2.7.0" FACTER_GEM_VERSION="~> 1.6.0"
   - rvm: 1.8.7
     env: PUPPET_GEM_VERSION="~> 2.7.0" FACTER_GEM_VERSION="~> 1.7.0"
-  allow_failures:
-    - rvm: 2.1.6
-      env: PUPPET_GEM_VERSION="~> 4.0" STRICT_VARIABLES="yes"
 notifications:
   email: false


mhaskel:
@cdenneen it looks like this file is also used in Fedora. Do you know whether the fixes will work there also?

cdenneen:
@mhaskel yes back to Fedora 17, PG 9.1 this file has the version number, so has been broken when used with managed repo,  in the file name so the include wouldn't work in template:

```
# rpm -qpl postgresql91-server-9.1.4-3PGDG.f17.x86_64.rpm  | grep systemd
/usr/lib/systemd/system/postgresql-9.1.service
```

templates/systemd-override.erb
```diff
@@ -1,4 +1,8 @@
+<% if @managed_package_repo -%>
```
mhaskel: The parameter should be `manage_package_repo` not `managed_package_repo`.

cdenneen: @mhaskel Updated PR

cdenneen:
@mhaskel Updated PR has passed tests

mhaskel:
@cdenneen sorry for the delay in getting back to this, but this looks mostly good. Could you please squash this down to a single commit though? Thanks!

PR #615, opened 2015-04-25
Add default values for more openSUSE and SLES distro versions

tampakrap:


tampakrap:
I don't think the test failures are related to my changes

manifests/globals.pp
```diff
@@ -94,7 +94,13 @@
     'FreeBSD' => '93',
     'OpenBSD' => '9.3',
     'Suse' => $::operatingsystem ? {
-      'SLES' => '91',
+      'SLES' => $::operatingsystemrelease ? {
+        '11.3' => '91',
```
hunner: This should be the `default` case, for backwards compatibility

cyberious: :+1: after default added

tampakrap: the `default => '91'` would be wrong though eg on SLES11SP2 (and maybe on the upcoming SLES12SP1, I need to check it). How about `default => 'undef'`?

tampakrap: I was wrong, SLES11SP2 also has '91', so I changed the default

DavidS: This line can be removed now? Also, how is this working in future versions of SLES? Will 12.1 have 91 again, or does that need another patch?

Using [versioncmp](https://docs.puppetlabs.com/references/latest/function.html#versioncmp) would make that much nicer.

tampakrap: I took the liberty to use 93 as default, and assign the 11.x distros to 91 through a regexp. Now it covers all the 11.x series (along with the upcoming sle11sp4) and also sle12 and the upcoming sle12sp1

PR #616, opened 2015-05-04
(MODULES-2007) Fix Puppet.newtype deprecation warning

roman-mueller:
On Puppet 4 this module is throwing a deprecation warning because it is using Puppet.newtype to create the postgresql_conf type.  
This PR converts it to use Puppet::Type.newtype.


hunner:
Failure due to apt pinning, not this PR's changes.

PR #621, opened 2015-05-08
Ordering fixups

dougneal:
An additional ordering constraint and a bit more robustness around existing ones.
 * If a role is declared along with a database owned by it, make sure the role is created first
 * Use resource collectors to match on the specific parameters of `postgresql::server::role` and `postgresql::server::tablespace`, instead of just matching the resource title (which may be different).

### DIFF
--- a/manifests/server/database.pp
+++ b/manifests/server/database.pp
@@ -90,7 +90,11 @@ define postgresql::server::database(
     }
   }
 
-  # Build up dependencies on tablespace
+  # Build up dependencies
+  if($owner != undef) {
+    Postgresql::Server::Role <| username == $owner |> -> Exec[$createdb_command]
+  }
+
   if($tablespace != undef and defined(Postgresql::Server::Tablespace[$tablespace])) {
     Postgresql::Server::Tablespace[$tablespace]->Exec[$createdb_command]
   }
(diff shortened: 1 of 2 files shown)


hunner:
Ehh, collectors can't be added, but it looks like if the role define is ever declared with a custom title, I see where you would run into issues.

What if the exec resource was converted to a new custom type (copy postgresql_psql and rename to postgresql_createdb, then the provider inherits the postgresql_psql provider) and add an `autorequire` to the new custom type. That's the only way I can think of how to do this without collectors.

How does that sound?

dougneal:
Sounds like that would work, and would be cleaner in some ways too. I've been applying collectors in my profile class in the meantime to stop Puppet trying to create a database before creating its owning role. Are collectors specifically forbidden in modules?

ffrank:
I disbelieve that there is a general rule, but they are problematic in most scenarios. There might be edge cases where the module can adequately predict that collected resources cannot sensibly be created as virtual, but modules should mostly avoid collecting wildly.

Artperkitny:
Fix problem on postgresql 9.3.9 with Db validation occurring before new role is declared.

tphoney:
Hi @dougneal , thanks for the work you have put in to this pull request. However due to inactivity and needs rebased this pull request is being closed. If you want to re-submit this PR with the necessary changes, we will gladly look at the new pull request.

PR #622, opened 2015-05-12
Fixed the metadata of the module so you can install apt 2.0.1

supernovae:
Simple cleanup to accept puppetlabs-apt 2.0.1 as a valid dependency so r10k/librarian and puppet module install won't complain about version dependency


tedivm:
Have you actually tested this? There were some backwards incompatible changes in the apt 2.0 module.

PR #623, opened 2015-05-19
(#2049) Make title of psql resource for schema creation unique

dirkweinhardt:
In order to avoid duplicate resource definitions when defining schemas with the same name in different databases on the same server.


igalic:
:+1:

PR #625, opened 2015-05-20
(#2056) Added 9.4, corrected past versions based on docs

cjestel:
http://www.postgresql.org/docs/9.4/static/auth-pg-hba-conf.html
Doc does not show sha1 as a valid type in any of the versions it was listed for.  9.4 was not in list, so it was populated based on doc as well.


cjestel:
The spec has been adjusted to match the new correct result and all tests are passing.
https://tickets.puppetlabs.com/browse/MODULES-2056

nibalizer:
Can you squash this? @igalic likes all commits to be passing.

PR #626, opened 2015-05-20
Support puppetlabs-concat 2.x

domcleal:
2.0.0 is compatible with current 1.x usage, so permit either 1.x or 2.x
to satisfy the concat dependency.  The fragment `force` parameter is
deprecated, but only results in a warning.  It can be removed at a later
date when 2.x is prevalent.


mcanevet:
:+1:

mhaskel:
Concat 2.0.0 has a major bug, I'm going to hold off on merging this until 2.0.1 is released (next week).

nibalizer:
2.0.1 has released.

PR #628, opened 2015-05-25
Fix the APT handling.

ciaby:
The new APT module changed some parameters, this is a fix for that change.


mhaskel:
Hi @ciaby, thanks for the contribution! I'm going to close this for now though, we're releasing a version of apt (2.1.0) tomorrow (2015-06-15) that is API compatible with the 1.8.x series, so we don't need to make any changes to the module to support the new apt.

PR #629, opened 2015-05-26
Copy snakeoil certificate and key instead of symlinking

mcanevet:
Since postgresql-9.1_9.1.16-0+deb7u1 on wheezy, postgresql can't read
snakeoil certificate as symlink anymore, so server does not restart.
This patch copies cert and key instead of symlinking so that it works
again.

### DIFF
--- a/manifests/server/initdb.pp
+++ b/manifests/server/initdb.pp
@@ -83,19 +83,25 @@ class postgresql::server::initdb {
       require   => File[$require_before_initdb],
     }
     # The package will take care of this for us the first time, but if we
-    # ever need to init a new db we need to make these links explicitly
+    # ever need to init a new db we need to copy these files explicitly
     if $::operatingsystem == 'Debian' or $::operatingsystem == 'Ubuntu' {
       if $::operatingsystemrelease =~ /^6/ or $::operatingsystemrelease =~ /^7/ or $::operatingsystemrelease =~ /^10\.04/ or $::operatingsystemrelease =~ /^12\.04/ {
         file { 'server.crt':
-          ensure  => link,
+          ensure  => file,
           path    => "${datadir}/server.crt",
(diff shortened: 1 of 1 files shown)


mfournier:
Same problem with debian jessie's postgresql-9.4.2-0+deb8u1. It seems that https://www.debian.org/security/2015/dsa-3270 and https://www.debian.org/security/2015/dsa-3269 made something more strict regarding the location of these keys & certs, but it doesn't really strike out in the annoucements.

igalic:
:+1:

bastelfreak:
:+1: for that

dougneal:
:+1:

saimonn:
:+1:

mfournier:
Here are some details about this issue: https://wiki.postgresql.org/wiki/May_2015_Fsync_Permissions_Bug

nb: all distros would be affected, not only Debian/Ubuntu. Not sure how the default install looks like on other distros, maybe no specific action needs to be taken (ie: cert/key are already have the correct ownership/permission/location)

raphink:
Looks like a good fix to me.

igalic:
@mfournier while that may be true, we haven't implemented it on other platforms, so we couldn't have broken it.

what i'm curious about now is: does that mean that a puppet installed/configured postgresql on rhel platforms doesn't support ssl connections?

mcanevet:
We add this really annoying problem more than 3 weeks ago and nobody else complained... I'd be curious to know if we are the only one with this problem...

@DavidS could you please review this PR?

DavidS:
@mcanevet the fix looks straight-forward enough. I'd suggest using 0600 permissions for the key, though, unless you have an overriding need for allowing group members to read the key.

If you're in there, please also fix the comment to not refer to linking.


Please ping me for merge.

mcanevet:
@DavidS ping

DavidS:
brilliant. thanks!

PR #630, opened 2015-05-27
Support Puppet v4

ghoneycutt:


metadata.json
```diff
@@ -59,15 +59,15 @@
...
     },
     {
       "name": "puppet",
-      "version_requirement": "3.x"
+      "version_requirement": ">= 3.0.0 < 5.0.0"
     }
   ],
   "dependencies": [
-    {"name":"puppetlabs/stdlib","version_requirement":"4.x"},
+    {"name":"puppetlabs/stdlib","version_requirement":">= 4.0.0 < 6.0.0"},
```
ghoneycutt: This is because Puppet v3 is compatible with stdlib v3 - v5

https://github.com/puppetlabs/puppetlabs-stdlib#version-compatibility

mhaskel:
@ghoneycutt thanks for the contribution, but we updated the puppet/pe support via #643, closed/commented on the stdlib updates in #642 and any .travis.yml updates should happen via modulesync so I'm closing this.

PR #631, opened 2015-05-27
Drop support for Puppet v2.7 which is way past end of life

ghoneycutt:


bmjen:
@ghoneycutt We will drop 2.7 tests when we release a major version update of the module.

PR #632, opened 2015-05-27
Switched to the 2.0 version of the apt module

tedivm:
This updates the usages of the apt-get module.

### DIFF
--- a/manifests/repo/apt_postgresql_org.pp
+++ b/manifests/repo/apt_postgresql_org.pp
@@ -10,12 +10,12 @@ include ::apt
     priority   => 500,
   }->
   apt::source { 'apt.postgresql.org':
-    location    => 'http://apt.postgresql.org/pub/repos/apt/',
-    release     => "${::lsbdistcodename}-pgdg",
-    repos       => "main ${postgresql::repo::version}",
-    key         => {
+    location => 'http://apt.postgresql.org/pub/repos/apt/',
+    release  => "${::lsbdistcodename}-pgdg",
+    repos    => "main ${postgresql::repo::version}",
+    key      => {
       id         => 'B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8',
-      key_source  => 'https://www.postgresql.org/media/keys/ACCC4CF8.asc',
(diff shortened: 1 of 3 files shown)


mullr:
cc @mullr

mullr:
@tedivm The PR triage team is meeting tomorrow morning, and this is on their list.

tedivm:
Awesome :-)

You should get them to check out #626 too so this module can be up to date with the new concat one as well.

manifests/repo/apt_postgresql_org.pp
```diff
@@ -10,12 +10,17 @@
...
-    location    => 'http://apt.postgresql.org/pub/repos/apt/',
-    release     => "${::lsbdistcodename}-pgdg",
-    repos       => "main ${postgresql::repo::version}",
-    key         => 'B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8',
-    key_source  => 'https://www.postgresql.org/media/keys/ACCC4CF8.asc',
-    include_src => false,
+    location => 'http://apt.postgresql.org/pub/repos/apt/',
+    release  => "${::lsbdistcodename}-pgdg",
+    repos    => "main ${postgresql::repo::version}",
+    key      => {
```
cyberious: This is not backwards compatible with apt 1.8, needs to bump metadata to use apt 2.x

mhaskel:
@tedivm would you be ok with this waiting a few weeks for the new version of apt? Then the change can just be simplified to a metadata change :)

tedivm:
The downside to that is I won't be able to use the puppetdb module, as it depends on this postgres module and I can't install this postgres module without causing dependency conflicts. Waiting several weeks for puppetdb means pushing my whole deployment back.

tedivm:
Frankly, I think you guys should skip this- and everything else you're doing- to try to slip out an apt module sooner than a few weeks agrom now that doesn't break everyone's projects.

mhaskel:
@tedivm thanks, but I'm going to close this PR for now. We're working on the apt 2.1.0 release which will allow this module to use 1.x or 2.x, and it will only require updating the metadata binding. Once apt is released we'll  add a new PR to fix that.

PR #569, opened 2015-03-04
Add classes for plpython and postgresql-docs 

deVIAntCoDE:
Hi, i have just added these features:
1. support for plpython package management, 
2. suport for postgresql-docs management. 
My need was for Rehat (Centos 6), as such i haven't investigated implications for Debian and other providers. 
Kindly review and let me know what you think. 
Thanks

PR #634, opened 2015-05-28
Add plpython and postgresql-docs classes

DavidS:
This is a straight-forward rebase/squash of https://github.com/puppetlabs/puppetlabs-postgresql/pull/569

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -1,55 +1,57 @@
 # Class for setting cross-class global overrides. See README.md for more
 # details.
 class postgresql::globals (
-  $client_package_name  = undef,
-  $server_package_name  = undef,
-  $contrib_package_name = undef,
-  $devel_package_name   = undef,
-  $java_package_name    = undef,
-  $perl_package_name    = undef,
-  $plperl_package_name  = undef,
-  $python_package_name  = undef,
-  $postgis_package_name = undef,
+  $client_package_name    = undef,
+  $server_package_name    = undef,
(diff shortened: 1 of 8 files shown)


manifests/globals.pp
```diff
@@ -1,55 +1,57 @@
...
 
-  $manage_pg_hba_conf   = undef,
-  $manage_pg_ident_conf = undef,
-  $manage_recovery_conf = undef,
+  $manage_pg_hba_conf     = undef,
+  $manage_pg_ident_conf   = undef,
+  $manage_recovery_conf   = undef,
 
-  $manage_package_repo  = undef
+  $manage_package_repo    = undef,
```
cyberious: additional spaces will cause lint issue

cyberious:
I think this might have been more aggressive squash than what we needed, it should have kept all of the previous commit messages within the body of the one commit with a summary on the first line

DavidS:
The other commit messages only referred to changes to the original commit. The only thing I might have dropped inadvertently is that "and tests" part.

manifests/globals.pp
DavidS: `rake lint` passes this locally.

cyberious:
I was talking more about the body of the commit message not the summary line. Similar to https://github.com/puppetlabs/puppetlabs-registry/commit/868e77e4613b7eed2e48b464e19bd546cdced56c were we have a summary and more details as to why the change

DavidS:
Updated the commit description.

PR #638, opened 2015-06-08
the force parameter is not anymore supported by the concat module

duritong:


manifests/server/config.pp
```diff
@@ -139,7 +139,6 @@
     concat { $pg_ident_conf_path:
       owner  => $user,
       group  => $group,
-      force  => true, # do not crash if there is no pg_ident_rules
```
bmjen: Concat 2.0.x should continue to accept the force parameter, but throw a deprecation warning. If you remove the `force` param here, then any postgresql installation that is still running < 2.0.0, will fail... unless you also plan on changing the `metadata.json` to raise the minimum concat version to 2.0.0.

duritong: Concat module exactly raises a warning. This commit is about getting rid of the warning, as it can be confusing to users, as warnings are in red.

mhaskel:
@duritong thanks for this, but since the concat 2.0.x releases have been removed (and #640 has a commit to adjust the dependencies) I don't think this is necessary so I'm closing this PR. Thanks again!

PR #639, opened 2015-06-12
Fixed systemd override for manage_repo package versions

cdenneen:
@mhaskel squashed commit submitted


mhaskel:
Great, thanks @cdenneen !

PR #640, opened 2015-06-12
Update metadata dependencies

mhaskel:
Once puppetlabs-apt 2.1.0 is released (2015-06-16) the postgresql module
will be compatible with apt 1.x (>= 1.8.0) and 2.x (>= 2.1.0). This will
*not* work with puppetlabs-apt 2.0.x.


mhaskel:
Also tightening the concat dependency since concat 2.0.x releases have been removed

PR #642, opened 2015-06-16
Accept compatible versions of stdlib

ghoneycutt:
5.x will also be compatible, so this should accept that, too.

https://github.com/puppetlabs/puppetlabs-stdlib#version-compatibility

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -67,7 +67,7 @@
     }
   ],
   "dependencies": [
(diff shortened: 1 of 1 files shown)


mhaskel:
@ghoneycutt thanks for the contribution, but since stdlib5 does not exist yet we can't be certain it will be compatible with the usage of stdlib4 in this module. Once stdlib5 is more solidified we will adjust dependencies accordingly.

PR #645, opened 2015-06-18
(maint) Add beaker-puppet_install_helper and fix fact bug  

cyberious:
- Issue with fact() was being called against default instead of    using fact_on when iterating against each host, also was installing 3.8.1 despite   specifying PUPPET_VERSION=1.1.1 so updated to use beaker-puppet_install_helper

### DIFF
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -1,5 +1,10 @@
 require 'beaker-rspec/spec_helper'
 require 'beaker-rspec/helpers/serverspec'
+require 'beaker/puppet_install_helper'
+
+run_puppet_install_helper
+
+UNSUPPORTED_PLATFORMS = ['AIX','windows','Solaris','Suse']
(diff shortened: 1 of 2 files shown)


spec/spec_helper_acceptance.rb
```diff
@@ -34,27 +39,6 @@ def psql(psql_cmd, user = 'postgres', exit_codes = [0,1], &block)
...
-  foss_opts = { :default_action => 'gem_install' }
-
-  if default.is_pe?; then install_pe; else install_puppet( foss_opts ); end
-
-  hosts.each do |host|
-    shell("mkdir -p #{host['distmoduledir']}")
-    if ! host.is_pe?
-      # Augeas is only used in one place, for Redhat.
-      if fact('osfamily') == 'RedHat'
-        install_package host, 'ruby-devel'
```
hunner: Is ruby-devel not needed any more?

PR #652, opened 2015-06-29
update apt module version requirement

tobru:
The reworked APT module has some backwards compatible changes.
So the dependency for APT module version can now be more relaxed.


mhaskel:
Thanks @tobru , but this has already been fixed so I'm closing this PR

PR #659, opened 2015-07-01
(MODULES-2181) Fix variable scope for systemd-override

kbarber:
$manage_package_repo wasn't in scope. This was causing all RHEL7 systems with
manage_package_repo on to fail on startup using systemctl, since the template
systemd-override.erb relies on that variable in a conditional, to ensure the
correct PGDG path is set.

Signed-off-by: Ken Barber <ken@bob.sh>


kbarber:
For reference, before & after this patch: https://gist.github.com/kbarber/f61a654f7cc9b74714e2

DavidS:
Travis CI got confused here. I tested it locally against puppet 4.

PR #663, opened 2015-07-02
(MODULES-2185) Fix `withenv` execution under Puppet 2.7

domcleal:
When executing a command with environment variables, the withenv helper
under Puppet 2.7 is on Puppet::Util::Execution and on 3.0 to 3.3, it's
on Puppet::Util.


mhaskel:
@domcleal thanks for the contribution, but can this please be targeted against master? If we end up doing another 4.4.x release we'll pull it over, but likely the next release will be something else.

domcleal:
@mhaskel thanks, if you're doing another 4.4.x, it'd be much appreciated since it's a regression.  I've opened #664 against master to replace this PR.

PR #677, opened 2015-07-29
(MODULES-2211) Fixed systemd-override for RedHat systems with unmanag…

cavaliercoder:
…ed Yum repos

This change lets the params class compute the name of the correct systemd unit file
via $service_name.

I've tested the following scenarios successfully on a CentOS 7.1 vagrant box:
* manage_package_repo = true, version = specified
* manage_package_repo = false, version = specified (with manually configured repo)
* manage_package_repo = false, version = undef

This fixes the issues described in MODULES-2211 where the `.include` path in the systemd-override template doesn't exist if manage_package_repo = false and version is specified.

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -20,6 +20,7 @@ class postgresql::server::config {
   $manage_recovery_conf       = $postgresql::server::manage_recovery_conf
   $datadir                    = $postgresql::server::datadir
   $logdir                     = $postgresql::server::logdir
+  $service_name               = $postgresql::server::service_name
 
   if ($manage_pg_hba_conf == true) {
     # Prepare the main pg_hba file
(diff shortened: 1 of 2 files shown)


jonnytpuppet:
Please forgive me if I am misunderstanding your code here, but isn't $postgresql::server::service_name and $postgresql::params::service_name the same thing?

Thanks

cavaliercoder:
You are correct! There is a reason though.
I've brought `${postgresql::params::service_name}` into scope as `$service_name` so that it can be used in the `systemd-override.erb` template to resolve this issue. This is also consistent with `$port` and `$datadir` which are used in the template.

cavaliercoder:
Is there anything further I can do to progress this PR? I'd love to switch our servers back to the stable branch without breaking the systemd config. Many thanks.

DavidS:
Thank you for catching this.

PR #678, opened 2015-08-04
Decouple pg_hba_rule from postgresql::server

npwalker:
This change allows the pg_hba_rule defined type to be used
even if you have not included postgresql::server.

By default it will continue with the same behavior as before,
however, you can now specify some extra parameters that were
previously coming directly from postgresql::server

### DIFF
--- a/README.md
+++ b/README.md
@@ -700,6 +700,19 @@ This would create a ruleset in `pg_hba.conf` similar to:
     # Order: 150
     host  app  app  200.1.2.0/24  md5
 
+By default, `pg_hba_rule` requires that you include `postgresql::server`, however, you can override that behavior by setting target and postgresql_version when declaring your rule.  That might look like the following.
+
+    postgresql::server::pg_hba_rule { 'allow application network to access app database':
+      description        => "Open up postgresql for access from 200.1.2.0/24",
+      type               => 'host',
+      database           => 'app',
+      user               => 'app',
+      address            => '200.1.2.0/24',
+      auth_method        => 'md5',
+      target             => '/path/to/pg_hba.conf',
+      postgresql_version => '9.4',
+    }
+
 ####`namevar`
 A unique identifier or short description for this rule. The namevar doesn't provide any functional usage, but it is stored in the comments of the produced `pg_hba.conf` so the originating resource can be identified.
 
@@ -730,6 +743,8 @@ An order for placing the rule in `pg_hba.conf`. Defaults to `150`.
 ####`target`
 This provides the target for the rule, and is generally an internal only property. Use with caution.
 
+####`postgresql_version`
+Defaults to the version set in `postgresql::server`.  Use this if you want to manage `pg_hba.conf` without managing the entire PostgreSQL instance.
 
 ###Resource: postgresql::server::pg\_ident\_rule
 This defined type allows you to create user name maps for `pg_ident.conf`. For more details see the [PostgreSQL documentation](http://www.postgresql.org/docs/current/static/auth-username-maps.html).
--- a/manifests/server/pg_hba_rule.pp
+++ b/manifests/server/pg_hba_rule.pp
@@ -12,10 +12,19 @@ define postgresql::server::pg_hba_rule(
 
   # Needed for testing primarily, support for multiple files is not really
   # working.
-  $target      = $postgresql::server::pg_hba_conf_path
+  $target             = $postgresql::server::pg_hba_conf_path,
+  $postgresql_version = $postgresql::server::_version
 ) {
 
-  if $postgresql::server::manage_pg_hba_conf == false {
+  #Allow users to manage pg_hba.conf even if they are not managing the whole PostgreSQL instance
+  if !defined( 'postgresql::server' ) {
+    $manage_pg_hba_conf = true
+  }
+  else {
+    $manage_pg_hba_conf = $postgresql::server::manage_pg_hba_conf
+  }
+
+  if $manage_pg_hba_conf == false {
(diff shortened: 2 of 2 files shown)


hunner:
Could you add documentation for the new parameters?

npwalker:
@hunner I added some documentation to the README let me know if that seems reasonable.

README.md
```diff
@@ -730,6 +744,11 @@ An order for placing the rule in `pg_hba.conf`. Defaults to `150`.
 ####`target`
 This provides the target for the rule, and is generally an internal only property. Use with caution.
 
+####`postgresql_version`
+Defaults to the version set in `postgresql::server`.  Use this if you want to manage `pg_hba.conf` without managing the entire postgreSQL instance.
```
DavidS: Please spell it `PostgreSQL` to be consistent with the rest of the README.

README.md
```diff
@@ -730,6 +744,11 @@ An order for placing the rule in `pg_hba.conf`. Defaults to `150`.
 ####`target`
 This provides the target for the rule, and is generally an internal only property. Use with caution.
 
+####`postgresql_version`
```
DavidS: Is there a (code) reason for the `postgresql_` prefix? Maybe just `version` or, to be more explicit, `server_version` would be better?

README.md
```diff
@@ -730,6 +744,11 @@ An order for placing the rule in `pg_hba.conf`. Defaults to `150`.
...
+Defaults to the version set in `postgresql::server`.  Use this if you want to manage `pg_hba.conf` without managing the entire postgreSQL instance.
+
+####`manage_pg_hba_conf`
```
DavidS: I don't think this is needed here. What would be the use-case of creating a `postgresql::server::pg_hba_rule` with `manage_pg_hba_conf` set to false explicitely?

manifests/server/pg_hba_rule.pp
```diff
@@ -12,10 +12,12 @@
...
   # Needed for testing primarily, support for multiple files is not really
   # working.
-  $target      = $postgresql::server::pg_hba_conf_path
+  $target             = $postgresql::server::pg_hba_conf_path,
+  $postgresql_version = $postgresql::server::_version,
+  $manage_pg_hba_conf = $postgresql::server::manage_pg_hba_conf
 ) {
 
-  if $postgresql::server::manage_pg_hba_conf == false {
+  if $manage_pg_hba_conf == false {
```
DavidS: As mentioned above, this could very well stay like it was.

manifests/server/pg_hba_rule.pp
```diff
@@ -12,10 +12,12 @@
 
   # Needed for testing primarily, support for multiple files is not really
   # working.
-  $target      = $postgresql::server::pg_hba_conf_path
+  $target             = $postgresql::server::pg_hba_conf_path,
+  $postgresql_version = $postgresql::server::_version,
+  $manage_pg_hba_conf = $postgresql::server::manage_pg_hba_conf
```
DavidS: This will create a parse-time order dependency between this code and `postgresql::server`. If for some reason the puppet compiler evaluates a `postgresql::server::pg_hba_rule` before encountering and evaluating the `postgresql::server` inclusion, this will pick up the wrong defaults here.

README.md
npwalker: @DavidS The purpose of the parameter is to allow setting it without needing to include `postgresql::server`.  My use case is to manage pg_hba_conf without managing the entire postgresql instance.  

I don't think this is the best long term solution but I'm hoping it's good enough for now.

manifests/server/pg_hba_rule.pp
npwalker: Doesn't the parse-time order dependency already exist?  I'm just trying to give myself a way to use this defined type without being tightly coupled to `postgresql::server`

README.md
npwalker: I think `version` would be too generic.  As a user, I'd think "version of what"?  `server_version` feels slightly better because I can probably figure it out from context clues.  I have a personal preference for `postgresql_version` but if it really matters I can change it.

DavidS: It's already a given that it's postgresql, as it is attached to a define in the postgresql module. But it's your code, so I'll defer to your decision after having said my piece.

README.md
DavidS: I understand that you want to elide the inclusion, but I don't understand the usecase of "I have not included `postgresql::server` but I want to set `manage_pg_hba_conf` to `false` to cause a compile time error". So what seems to remain is avoiding referencing a undefined variable? Using `defined()` or `pick()` to do so would reduce the required API surface.

manifests/server/pg_hba_rule.pp
DavidS: You're right of course.

README.md
npwalker: @DavidS I don't quite track what you are saying.  Are you saying remove this parameter altogether because the only way you'd use it is to set it to true and then change the if statement to be something like....

```
if defined ( 'postgresql::server' ) {
  if $postgresql::server::manage_pg_hba_conf == false {
       fail('postgresql::server::manage_pg_hba_conf has been disabled, so this resource is now unused and redundant, either enable that option or remove this resource from your manifests')
   } else {
     validate_re($type, '^(local|host|hostssl|hostnossl)$',
     "The type you specified [${type}] must be one of: local, host, hostssl, hostnosssl")
}
```

It hadn't occurred to be to check if `postgresql::server` was already defined and I think if that works the way I think it does then I'm happy to make that change as it's much cleaner.

npwalker: Seems like my original idea was wrong but I've updated the PR with my understanding.  Let me know what you think.

DavidS:
Thank you for this! Excellent work.

PR #679, opened 2015-08-19
removed inherits postgresql::params

vicinus:
inheriting `postgresql::params` creates a cycle, because `postgresql::repo` gets instantiated in `postgresql::globals` which is inherited by `postgresql::params`. Also `postgresql::params` are not needed, because only version is used by `postgresql::repo`, `postgresql::repo::yum_postgresql_org` and `postgresql::repo::apt_postgresql_org` and it get set in `postgresql::globals` by a resource like class instantiation.

The following error is created with puppet 4.2.1 if `manage_package_repo` is set to true in hiera:
```
 puppet apply --modulepath=/etc/puppetlabs/code/environments/production/modules -e 'include postgresql::server'
 Error: Evaluation Error: Error while evaluating a Resource Statement, Could not find scope for postgresql::params at /etc/puppetlabs/code/environments/production/modules/postgresql/manifests/globals.pp:140:5 on node test.local
```

Interestingly if `include postgresql::globals` is added prior to `include postgresql::server` then it works. Probably because `postgresql::repo` and `postgresql::params` are already loaded when `postgresql::server` is loading.

### DIFF
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -2,7 +2,7 @@
 class postgresql::repo (
   $version = undef,
   $proxy = undef,
-) inherits postgresql::params {
+) {
   case $::osfamily {
     'RedHat', 'Linux': {
       if $version == undef {


DavidS:
I've tested this with puppet 3.8.1 on ruby 1.9.3 on Ubuntu 14.04 and I didn't hit this error; although using

```
class { "postgresql::server": ; "postgresql::globals": manage_package_repo => true; }
```

will complain about "Error: Duplicate declaration: Class[Postgresql::Globals] is already declared; cannot redeclare at line 1 on node localhost" independent of this fix applied or not.

Using

```
class { "postgresql::globals": manage_package_repo => true; "postgresql::server": }
```

works without this fix applied.

vicinus:
Sorry, I didn't make myself clear enough in my first report. The problem only occurs if hiera is used to configure `postgresql::globals`. Because if hiera is used, `include postgresql::globals` isn't needed to configure `manage_package_repo` prior to `include postgresql::server`. At the moment my workaround is to use `include postgresql::globals include postgresql::server` instead of only `include postgresql::server`, but I would like to see it fixed in the postgresql module because I believe that other people will also stumble upon this problem. 

I also tested it with puppet 3.8.2 on ruby 1.9.3 on Ubuntu 14.04 and the described error also occurs whether the future parser is used or not (if hiera is used to configure `postgresql::globals::manage_package_repo` and only `include postgresql::server` is used to load the postgresql server).

PierreR:
FWIW I had a similar problem with `globals` (using hiera). It works ok if defined in the same file but cease to work when refactoring some snippet into an  include.

dalen:
have encountered the same issue. And as states including `postgresql::globals` first works as a workaround.

PR #684, opened 2015-08-25
fedora 22 postgresql version 9.4

javierwilson:


DavidS:
Hi Javier,

I'd love to merge the F22 commit, but I'm not so sure about the access method one. Do you have a reference when/where this default changed on the postgres side? I do seem to remember that 8.x still had ident as default and we should match that? If you are of a different (probably better reasoned too) opinion, please add a little explanation to the commit message and/or as comment inline.

Regards, David

javierwilson:
You're right, actually I only wanted to submit the fedora 22 versin 9.4 part, but I don't know how to do a pull request of just one commit. Sorry.

nibalizer:
I've proposed #698 which contains only the f22 commit. :)

tphoney:
I have merged #698 thanks for the work @javierwilson @nibalizer :+1:

PR #687, opened 2015-08-26
Always add version when >= 9.1

PierreR:
Even in the case 'manage_package_repo' is false

By policy, we mirrored all external repos. Systemd won't start at all with a version >= 9.1 in this case.

### DIFF
--- a/spec/unit/classes/server/config_spec.rb
+++ b/spec/unit/classes/server/config_spec.rb
@@ -17,22 +17,12 @@ describe 'postgresql::server::config', :type => :class do
         :path => '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
       }
     end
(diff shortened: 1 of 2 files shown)


PierreR:
Can someone explain the rational behind the failing tests ? Are testing that a plain vanilla RH 7 as

PierreR:
Any comment ?

HelenCampbell:
Hi @PierreR, 
Apologies for the delay on this. If you can look at the conflicts that need resolved I can bring this PR up at the next community meeting on Thursday to get it reviewed and actioned.

PierreR:
@HelenCampbell f929cadd9cd83718894eaa7f981 probably solves the problem already. Unfortunately it was not integrated in the latest `4.6.0` unfortunately even if the fix dates back from July ( #677)

PR #688, opened 2015-08-27
(MODULES-661) Remote DB support

DavidS:
Adds connection-settings (for remote DB support) when creating DB resources.

Connection-settings allows a hash of options that can be used
when connecting the a remote DB (such as PGHOST, PGPORT, PGPASSWORD
PGSSLKEY) and a special option DBVERSION indicating the version
of the remote database.

Including
- Puppet updates
- Documentation updates
- RSpec unit test updates
- RSpec acceptance test updates
- Some test coverage for connection-settings
- Working acceptance test...
    Basic vagrant setup:
    * Two boxes, server and client
    * Runs puppet code to on server to setup a postgres server that allows all connections and md5 connections, creates db puppet to look at
    * Runs puppet code on client to make a server that a psql command can be run against puppet db on other server
    * Does some fancy stuff to get the fact of the IP from the first server to connect to
- Backwards compatible, with deprecation warnings around old parameters


DavidS:
This is a rebase of #656

petems:
:+1:

PR #637, opened 2015-06-04
Fixes MODULES_2059 - adds extension argument

TJM:
This adds the ability to define the extension name separately from the "title" of the resource, which allows you to add the extension to more than one database.

PR #690, opened 2015-08-30
support for extension for several bases

progserega:
Any time need add one extension for many databases. But in postgresql::server::extension postgres extension name st as name of puppet resource. But puppet do not create many resources with one name. 
I move extension name to parameter from resource name. Now we can set up many resource postgresql::server::extension for one extension name for many databases.

### DIFF
--- a/manifests/server/extension.pp
+++ b/manifests/server/extension.pp
@@ -1,6 +1,7 @@
 # Activate an extension on a postgresql database
 define postgresql::server::extension (
   $database,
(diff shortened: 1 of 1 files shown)


DavidS:
This looks like a duplicate of #637 ?

progserega:
Yes. Sorry for duplicate...

DavidS:
No worries! Thank you for your contribution anyways!

PR #693, opened 2015-09-04
Fixes MODULES_2059 - adds extension argument

frconil:
This adds the ability to define the extension name separately from the "title" of the resource, which allows you to add the extension to more than one database.

As per the original ticket, extensions in postgresql can be defined on a per database basis. By using the same name for both the extension and the instance of postgresql::server::extension, you're getting duplicates errors if you try to assign an extension to more than one database.

I tried to expand the current tests to verify we can actually instantiate postgresql::server::extension twice.

Welcoming comments on this as I've never had to do it before :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -679,7 +679,10 @@ OS user for running `psql`. Defaults to the default user for the module, usually
 Hash of environment variable used when connecting to a remote server. Defaults to connecting to the local Postgres instance.
 
 ###Resource: postgresql::server::extension
-Manages a postgresql extension.
+This defined type manages a postgresql extension for a given database.
+
+####`extension`
+The extension to activate. If left blank, will use the name of the resource.
 
 ####`database`
 The database on which to activate the extension.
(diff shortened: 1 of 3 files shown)


tphoney:
Can you please update the README to include documentation for the new parameter.

Thanks for the great work.

frconil:
Updated the documentation as requested.

tphoney:
@mhaskel this pr superceeds #637 and has all the appropriate modifications. It looks good to merge, can you have one final look.

bmjen:
This looks good. Thanks for taking this one over @frconil!

timeapen:
I am new to puppet and am trying to use the postgresql module to install postgres and create several databases and install extensions on each one.  Because of this bug, I can't enable extensions on more than one db.

Currently, when I install the module via puppet, I get the latest release (4.6) which I believe does not have this fix.

When will this fix be released?  Or is there a way for me to install the module from the dev branch?

DavidS:
@timeapen There is no release planned soon. Meanwhile, you can use a git reference in your Puppetfile or download/clone it manually.

timeapen:
Thanks.   I am currently installing the postgres module as follows:

 puppet module install puppetlabs/postgresql

Is there anyway to use the module tool to install from a git-repo?

PR #694, opened 2015-09-04
MODULES-2525 - updated systemd-override to support fedora and CentOS paths for systemd

jbehrends:
https://tickets.puppetlabs.com/browse/MODULES-2525

The path in the current file does not exist on CentOS 7.  Doing some research it seems this is a Fedora path.  So this PR will template the correct path for Fedora or CentOS/Redhat.  Please see the linked ticket for official documentation from RedHat and Fedora.


DavidS:
Thank you for you contribution!

PR #696, opened 2015-09-07
(#MODULES-2438) Implement removal for dbs, roles and grants a.o.

holstvoogd:
Add ensure parameter & Implement removal logic for:
-  postgresql::server::database
-  postgresql::server::database_grant
-  postgresql::server::db
-  postgresql::server::grant
-  postgresql::server::pg_hba_rule
-  postgresql::server::pg_indent_rule
-  postgresql::server::role
-  postgresql::server::table_grant

Add unit tests for removal logic
Add acceptance test for removal via postgresql::server::db

### DIFF
--- a/manifests/server/database.pp
+++ b/manifests/server/database.pp
@@ -9,6 +9,7 @@ define postgresql::server::database(
   $locale           = $postgresql::server::locale,
   $istemplate       = false,
   $connect_settings = $postgresql::server::default_connect_settings,
+  $ensure     = 'present',
 ) {
   $createdb_path = $postgresql::server::createdb_path
   $user          = $postgresql::server::user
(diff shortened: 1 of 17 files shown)


tphoney:
This looks like a big change, i would like others to comment on its usefulness. One thing i would say is that the readme will definitely need updated to reflect the new functionality.

tomhey:
Personally I'd find this immensely useful combined with MODULES-661, this allows a RDS node to provided a service for multiple system tests (you'd just need to ensure each system test creates uniquely names DB objects). The obviously alternative without this patch would be to create and destroy an RDS nodes for each system test.

FYI: In production we are still using a Postgres module based on a branch in 2013 - before this type of functionality was removed - And I'd like to be able to switch back to using the currently release.

kbarber:
My memory could be fussy, but the functionality of 'removal' was reverted somewhat a while ago due to its complexity: https://github.com/puppetlabs/puppetlabs-postgresql/commit/58fe218e91947fb9a0eea44fcdb0f4cccc6c78c3

So @tomhey is roughly correct, I recall adding the basic uninstall functionality but sounds like it was removed due to testing complexity after I handed it over to the module team.

It was slightly brute force to remove packages and items I recall, this snippet gives you a clue: https://github.com/puppetlabs/puppetlabs-postgresql/commit/58fe218e91947fb9a0eea44fcdb0f4cccc6c78c3#diff-8d72d3019b1c5c78aa84d616a0d7257dL17

However, I don't believe this patch is about removing packaging, its about removing other elements (databases, grants and such) so the problem is different. TBH I guess in a perfect world I'd prefer to see this done with real types/providers and using purgeable instead, but hohum.

DavidS:
> in a perfect world I'd prefer to see this done with real types/providers and using purgeable instead, but hohum.

Salt in my wounds, @kbarber

tomhey:
@kbarber correct the removal of **packages** change still makes sense (IMO) - It's not functionality that I see much need for, just discard the current node and build again from scratch. Unfortunately at the time it was removed there was no need to keep support for the removal of **database objects** so that disappear as well.

bmjen:
I merged this branch locally and acceptance tests have passed.

However, due to the invasive nature of this change, we would like to request acceptance tests be written for each define that was modified with an ensure before merging this.

hunner:
Thanks for your contribution! I am closing due to inactivity, but feel free to reopen when you have time.

PR #698, opened 2015-09-23
"fedora 22 postgresql version 9.4"

nibalizer:


nibalizer:
This is a re-up from #684  containing only the fedora22 support patch. Thanks @javierwilson for writing the patch.

PR #699, opened 2015-09-29
Added parameter to disable automatic service restarts on config changes

sw0x2A:
If you run Postgresql on production servers, you might want Puppet not to automatically restart Postgresql service when a config entry change has been applied that requires a restart rather than a reload to become active. This way, Puppet just applies the new config and Ops can restart the service during a scheduled maintenance window.

### DIFF
--- a/manifests/server/config_entry.pp
+++ b/manifests/server/config_entry.pp
@@ -17,9 +17,11 @@ define postgresql::server::config_entry (
 
   case $name {
     /data_directory|hba_file|ident_file|include|listen_addresses|port|max_connections|superuser_reserved_connections|unix_socket_directory|unix_socket_group|unix_socket_permissions|bonjour|bonjour_name|ssl|ssl_ciphers|shared_buffers|max_prepared_transactions|max_files_per_process|shared_preload_libraries|wal_level|wal_buffers|archive_mode|max_wal_senders|hot_standby|logging_collector|silent_mode|track_activity_query_size|autovacuum_max_workers|autovacuum_freeze_max_age|max_locks_per_transaction|max_pred_locks_per_transaction|restart_after_crash|lc_messages|lc_monetary|lc_numeric|lc_time/: {
-      Postgresql_conf {
-        notify => Class['postgresql::server::service'],
-        before => Class['postgresql::server::reload'],
+      if $postgresql::server::service_restart_on_change {
+        Postgresql_conf {
+          notify => Class['postgresql::server::service'],
+          before => Class['postgresql::server::reload'],
+        }
       }
     }
 
--- a/manifests/server/config_entry.pp
+++ b/manifests/server/config_entry.pp
@@ -23,7 +23,12 @@ define postgresql::server::config_entry (
           before => Class['postgresql::server::reload'],
         }
       } else {
-        Postgresql_conf {}
+        Postgresql_conf {
+            before => [
+                Class['postgresql::server::service'],
+                Class['postgresql::server::reload'],
+            ],
+        }
       }
     }
 
--- a/manifests/server/config_entry.pp
+++ b/manifests/server/config_entry.pp
@@ -22,6 +22,8 @@ define postgresql::server::config_entry (
           notify => Class['postgresql::server::service'],
           before => Class['postgresql::server::reload'],
         }
+      } else {
+        Postgresql_conf {}
       }
     }
 
--- a/README.md
+++ b/README.md
@@ -380,6 +380,9 @@ This setting can be used to override the default postgresql service provider. If
 ####`service_reload`
 This setting can be used to override the default reload command for your PostgreSQL service. If not specified, the module will the default reload command for your OS distro.
 
+####`service_restart_on_change`
+This setting can be used to override the default behaviour to restart your Postgresql service when a config entry has been changed that requires a service restart to become active. Defaults to `true`.
+
 ####`service_status`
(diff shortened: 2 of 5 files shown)


DavidS:
This is a very brittle approach, as the module cannot entirely avoid notifications coming from other parts of the manifest. I usually recommend running puppet in `noop` mode outside of maintenance windows to avoid any "accidents". Also that way the configuration on disk and in memory do not diverge, which can be confusing when looking at a machine.

sw0x2A:
Puppet agent runs as daemon and applies changes automatically in our case and it is the default behaviour, which makes me think that the given scenario is a problem for many other Puppet users. Since Puppet is not only managing Postgresql and we want other things to be updated and changed automatically by Puppet, your suggestion to run Puppet only within maintenance windows is not helpful.

kbarber:
@sw0x2A I think what is 'normal' for users is fairly subjective, so I'm qualifying that right up front.

What I see most users do, is run in noop mode out of maintenance windows, and then run without noop during maintenance. This is I feel more inline with the spirit of how puppet has been built.

Prepping a file change automatically, then requiring a manual change later on means that you're only partially automating elements of your configuration management, which I feel goes against the spirit of automation.

Is your concern here, that a particular PostgreSQL DB has a different maintenance window to everything else?

stevie-:
With this option users can choose which way they want to use puppet. Why is this a problem?

dejo1307:
I think that also 'most users' is quite subjective terms. "Most users" I know use Puppet exactly how @sw0x2A explained. However, as mentioned already, why is this a problem to make this an option, if it's just an option and everyone can choose which way to use it?

DavidS:
"It's just an option" that needs to be documented, tested, and maintained in the future. Given the sensitivity of the issue for those who actually want to set this, I'd love to see both a unit test in https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/spec/unit/classes/server_spec.rb verifying that the notification is correctly set up in the catalog, as well as a acceptance test that verifies that both settings work. That is, when it's set to true, that a change in the configuration actually restarts the process and when it's set to false a change in the configuration does NOT restart the process.

A sentence or two with appropriate warnings in the README (https://github.com/puppetlabs/puppetlabs-postgresql#class-postgresqlserver) would also be grand.

DavidS:
PS: I've earmarked this for further discussion in the modules community meeting thursday 19:00 CEST on https://bluejeans.com/280736660 if you want to join!

terrorobe:
Just as food for thought, three examples of modules offering configurable "restart on config change"-behaviour:

* https://github.com/puppetlabs/puppetlabs-mysql#restart
* https://github.com/saz/puppet-memcached#other-class-parameters
* https://github.com/fsalum/puppet-redis#parameters


While I understand that the mechanism doesn't offer 100% control it's a good compromise. Third party component modules mostly don't got handwavy with scheduling restarts for services they don't manage - and what the user does in his profiles/roles is completely up to him.

For clarification, @dejo1307, @sw0x2A and @stevie- and me all work for the same company - ResearchGate - now, but all have a diverse background, having worked at various Web/SaaS companies in the past. 

In web-oriented companies there's usually a strong push in the direction of continuous deployment/continuous integration - you'll find few if any maintenance windows. As soon as something is merged into the master branch it's good to go, but you still want to be able to trigger specific - service interrupting - operations on your own. Restarting datastore services is one of them  - especially if they're not redundant by design ;)

kbarber:
In response to your comment @dejo1307, I was stating my entire follow up comment was subjective which is why I stated that up-front in the manner I did, sorry perhaps I didn't make that clear. In fact this entire conversation can end up in a subjective space, which makes it hard to reason about :-).

Debate around new features is healthy, and it keeps the module at a high-quality, please don't take anything said as a complete disbelief that this feature would be useful to someone/everyone :-).

@terrorobe thanks for pointing out those precedents, I think this is most valuable for stating your case.

manifests/server/config_entry.pp
```diff
@@ -17,9 +17,13 @@
...
-      Postgresql_conf {
-        notify => Class['postgresql::server::service'],
-        before => Class['postgresql::server::reload'],
+      if $postgresql::server::service_restart_on_change {
+        Postgresql_conf {
+          notify => Class['postgresql::server::service'],
+          before => Class['postgresql::server::reload'],
+        }
+      } else {
+        Postgresql_conf {}
```
hunner: This should have:
```puppet
  before => [
    Class['postgresql::server::service'],
    Class['postgresql::server::reload'],
  ],
```

DavidS:
We talked that over at the community meeting and the group concurred that having the capability would be nice. There were a whole list on comments on the actual implementation:

* Needs an entry in the README
* Needs a test in https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/spec/unit/classes/server_spec.rb to check that the behaviour with the setting on and off matches the expectations
* Even when turning notifications off, the ordering between resources should be preserved, so instead of https://github.com/puppetlabs/puppetlabs-postgresql/pull/699/files#diff-50e14c38f0a9be74c9125ca82b18f15aR26, this should add `before` for both `Class['postgresql::server::service']` and `before => Class['postgresql::server::reload']` so that configuration is still deployed before starting the service when provisioning a node.

sw0x2A:
Thanks for all your comments. I will add those changes soon.

sw0x2A:
@DavidS I have added the missing entry in README, wrote unit tests for both conditions of the parameter and fixed the ordering between resources when service restarts are disabled. I think that all comments on the implementation are satisfied now.

DavidS:
Sweet! Thank you for working through this!

PR #702, opened 2015-10-28
Use double qoutes around database name.

stintel:
When using a database name that contains dashes or underscores, the
CREATE DATABASE statement fails with a syntax error. Use double quotes
around the database name to solve this.

### DIFF
--- a/manifests/server/database.pp
+++ b/manifests/server/database.pp
@@ -73,7 +73,7 @@ define postgresql::server::database(
   }
 
   postgresql_psql { "Create db '${dbname}'":
-    command => "CREATE DATABASE ${dbname} WITH OWNER=${owner} ${template_option} ${encoding_option} ${locale_option} ${tablespace_option}",
+    command => "CREATE DATABASE \"${dbname}\" WITH OWNER=${owner} ${template_option} ${encoding_option} ${locale_option} ${tablespace_option}",
     unless  => "SELECT datname FROM pg_database WHERE datname='${dbname}'",
     db      => $default_db,
     require => Class['postgresql::server::service']


antaflos:
:+1: from me, whatever that's worth.

I am surprised that this hasn't been caught much sooner since there are numerous places where double quotes should be used when handling database or role names. Even my very first PR (#32) was about that.

igalic:
/me doesn't have see merge button :|
ping @DavidS

DavidS:
This scares me, as it unveils a dangerous lack of test coverage :-/

igalic:
:|

@DavidS what's a good… or, general way to catch these?

DavidS:
clone https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/spec/acceptance/db_spec.rb for `postgresql::server::database` ? That's also probably the case why it wasn't caught earlier, because everyone uses `::db` and nobody uses `::database`. Which opens *another* can of worms...

igalic:
i really meant, in a ***general*** way.

DavidS:
Quickcheck for puppet? But what is your invariant? Applies-without-error, and applies-second-time-without-changes?

I admit, that would have caught this one.

igalic:
well, in this particular case: all databases, tables, and so on, in strings, are quoted.

DavidS:
Test coverage, code reviews for test coverage, or quickcheck. I know no easy ways.

igalic:
let's go with quickcheck, hypothetically.
what would a quickcheck for puppet need, to provide minimal functionality?
* could we test with puppet apply --noop?
* could we integrate with something *like* rspec-puppet?
* would we have to integrate with something like beaker?
* how important is the difference between `apply` and `agent`? important enough to test for it?

PR #703, opened 2015-11-06
Fix line endings

karyon:
See https://github.com/puppetlabs/puppetlabs-concat/pull/80 for an explanation.

i had no issues with the .rb file, but added it as well because i guessed all shebang lines are probably broken with windows line endings.

it should also be possible to just add the files with shebang lines to the gitattributes file instead of those very inclusive filters, if you prefer that.


igalic:
this is something i'd rather see in [modulesync_configs](https://github.com/puppetlabs/modulesync_configs/)

karyon:
actually, i have a similar PR open in concat: https://github.com/puppetlabs/puppetlabs-concat/pull/373, so... yeah.

should i open a PR there with exactly the same changes?

igalic:
yes, please.

bmjen:
Going to close this in lieu of doing with this modulesync PR.

https://github.com/puppetlabs/modulesync_configs/pull/39

bmjen:
Changed my mind. Probably don't want to wait for us to modulesync this.

PR #704, opened 2015-11-06
Fix default paths for Amazon Linux

downsj2:
The default paths in params are very wrong for Amazon Linux, this is a simple fix which allows the module to function.


tphoney:
Can you please replace the tabs with double spaces. It is causing travis to fail.

Thanks @downsj2

bmjen:
@downsj2 do you mind rebasing off master to pull in some unit test changes? Thanks.

downsj2:
@bmjen It's been rebased to upstream master, looks like the unit tests have kicked off.

bmjen:
@downsj2 Thanks! Looks like those commits need to be squashed down to one. That should be the final thing though.

downsj2:
@bmjen squashed and pushed.

PR #709, opened 2015-12-07
readme update: postgresql::server::db is a defined type not a class

jessereynolds:


igalic:
+1

DavidS:
Ha! Good catch, thanks!

PR #713, opened 2016-01-21
Support postgres-9.5

tnn:
Fixes the following error:
```
Evaluation Error: Error while evaluating a Function Call, pick(): must receive at least one non empty value at /etc/puppetlabs/code/environments/testing/modules/postgresql/manifests/globals.pp:126:30 on node localhost.ts.sv
```

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -134,6 +134,7 @@ class postgresql::globals (
     '9.3'   => '2.1',
     '9.4'   => '2.1',
     '93'    => '2.1',
+    '9.5'   => '2.1',
     default => undef,
   }
   $globals_postgis_version = $postgis_version ? {


mfn:
Awesome, can confirm in my vagrant machine this works!

(note: I tested it only with manually altering the ` manifests/globals.pp` file; I did not simulate completely installing a patched puppetlabs-postgres module).

DavidS:
@tnn @mfn I'm a bit confused about the origin of the error message: can one of you copy me the SHA or version of the module you're running, your exact error message (with line numbers and surrounding messages), how you're configuring your postgresql class, and maybe even the referenced lines from the file specified in the error?

https://github.com/puppetlabs/puppetlabs-postgresql/blob/aab6921e08c1c05fbe3febbd3bdbd58f540806a0/manifests/globals.pp#L122 shouldn't be able to trigger this, and I'm not having luck in reproing it locally.

tnn:
@DavidS We are running 4.5.0 (https://github.com/puppetlabs/puppetlabs-postgresql/tree/4.5.0). 

Something similar to:
```puppet
    encoding     => 'UTF-8',
    locale       => 'en_GB.UTF-8',
    version      => '9.5',
    needs_initdb => false,
  } ->
  class { 'postgresql::server':
    package_name     => 'postgresql-9.5',
    package_ensure   => '9.5.0-1.pgdg14.04+2',
    listen_addresses => '*',
  }
```
This is running Ubuntu 14.04 and we the official package from postgresql.org.

mfn:
@DavidS I was having `puppetlabs-postgresql 4.2.0` installed.

Using that version and trying to install Postgres 9.5 in my vagrant machine resulted in:
`==> default: Error: pick(): must receive at least one non empty value at /etc/puppet/modules/postgresql/manifests/globals.pp:112 on node ...`

However, I've now upgraded to the latest version `4.6.1` and I can also confirm that installing Postgres 9.5 works without altering `manifests/globals.pp` (and the number `9.5` does not appear).

How embarrassing!

dpash:
I'm seeing a similar error message:

    Error while evaluating a Function Call, 'versioncmp' parameter 'a' expects a String value, got Undef at /etc/puppetlabs/code/environments/production/external-modules/postgresql/manifests/params.pp:137:10 on node 

which is similarly caused by $version set to 9.5, as it picks up the default, so $postgis_version is set to undef. This PR would fix this issue for me.

DavidS:
Thanks everyone for their input. As far as I can tell from the responses, upgrading to the newest version of the module already fixes this.

Additionally, looking at https://packages.debian.org/search?suite=stretch&searchon=names&keywords=postgis the current version of postgis for 9.5 is 2.2, not 2.1, so the original PR seems to be factually wrong.

If you still have any problems with the newest version of the module, please feel free to post a ticket to our tracker at  https://tickets.puppetlabs.com/browse/MODULES?_ga=1.176502758.2118929047.1451986136 or submit a PR with a failing test case.

DavidS:
A more encompassing fix for postgis was just merged as #716.

PR #719, opened 2016-01-29
Fix password change failing

matonb:
postgres db user password change failing when postgres is listening on non standard port number.


bmjen:
Thanks @matonb! Can you squash those commits?

matonb:
No problem and sure, I'll try and squash the commit over the weekend.

matonb:
Commits squashed as requested.

matonb:
@bmjen Commits squashed

PR #726, opened 2016-02-01
Escape case where password ends with '$'.

Farzy:
postgresql_escape returns an invalid string if the password end in '$'.

    postgres=# alter role "postgres" password $$foo$$$;   
    ERROR:  syntax error at or near "$"
    LINE 1: alter role "postgres" password $$foo$$$;


tphoney:
Thanks for contributing, the ticket is here https://tickets.puppetlabs.com/browse/MODULES-3025

PR #731, opened 2016-02-04
amazon linux defaults to postgresql92 now

shawn-sterling:
module currently doesn't work with amazon linux

cat /etc/os-release 
NAME="Amazon Linux AMI"
VERSION="2015.09"
ID="amzn"
ID_LIKE="rhel fedora"
VERSION_ID="2015.09"
PRETTY_NAME="Amazon Linux AMI 2015.09"
ANSI_COLOR="0;33"
CPE_NAME="cpe:/o:amazon:linux:2015.09:ga"
HOME_URL="http://aws.amazon.com/amazon-linux-ami/"

yum install postgresql-X currently will install postgresql92-X

it might be better to check if the os[release][major] == 2015 and os[release][minor] >= 9, or a few other options; but I'm not sure when amazon moved to postgresql92 by default.

This works for me, but could break people running on older version of amazon linux.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -30,7 +30,7 @@ class postgresql::params inherits postgresql::globals {
       $version_parts      = split($version, '[.]')
       $package_version    = "${version_parts[0]}${version_parts[1]}"
 
-      if $version == $postgresql::globals::default_version {
+      if $version == $postgresql::globals::default_version and $::operatingsystem != 'Amazon' {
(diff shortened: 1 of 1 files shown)


HelenCampbell:
Hey @shawn-sterling, Thank you for this! I can't seem to find which version moved to default postgres92, however since we don't technically support Amazon Linux I am happy enough to merge in your change as is. Thanks Again :)

PR #732, opened 2016-02-04
Fix markdown error in README

martijn:
This fixes a small error in the README that messes up the markup.


bmjen:
Thanks @martijn . Good catch!

PR #733, opened 2016-02-06
add support for Ubuntu xenial (16.04)

mmoll:
A released version with this change would be nice.


bmjen:
Thanks for the effort @mmoll, but at this time we will have to wait on merging this PR, since Xenial isn't released yet. We will definitely keep this PR in mind when we do our Xenial update across all of the supported modules in April. For now, I'm going to close this PR.

mmoll:
Understandable, but as also older versions of Ubuntu already used systemd would be the change be more acceptable when metadata.json would be untouched?

PR #735, opened 2016-02-09
Correctly set $service_provider

antaflos:
The `$service_provider` variable in params.pp is user-definable via `postgresql::globals::service_provider`. Thus params.pp should reference that variable correctly, like other variables from globals.pp.


bmjen:
Thanks @antaflos !

PR #739, opened 2016-02-15
Fix usage of deprecated apt::source parameters.

ctavan:
The latest puppetlabs-apt module deprecates key_source and include_src
parameters to apt::source which produces the following output on puppetmaster:

```
(Scope(Apt::Source[apt.postgresql.org])) $include_src is deprecated and will be
removed in the next major release, please use $include => { 'src' => false }
instead
(Scope(Apt::Source[apt.postgresql.org])) $key_source is deprecated and will be
removed in the next major release, please use $key => { 'source' =>
https://www.postgresql.org/media/keys/ACCC4CF8.asc } instead.
(Scope(Apt::Key[Add key: B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8 from
Apt::Source apt.postgresql.org])) $key_source is deprecated and will be removed
in the next major release. Please use $source instead.
```

### DIFF
--- a/manifests/repo/apt_postgresql_org.pp
+++ b/manifests/repo/apt_postgresql_org.pp
@@ -13,9 +13,13 @@ include ::apt
     location    => 'http://apt.postgresql.org/pub/repos/apt/',
     release     => "${::lsbdistcodename}-pgdg",
     repos       => "main ${postgresql::repo::version}",
-    key         => 'B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8',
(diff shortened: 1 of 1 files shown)


hunner:
Though this updates the params, the metadata.json still says it works with apt >= 1.8.0 but I believe these parameters do not exist in an apt 1.8.x release, so this is backwards incompatible :(

tphoney:
Hi @ctavan , thanks for the work you have put into this pull request. However due to inactivity and it being backwards incompatible this pull request is being closed. If you want to re-submit this PR were it is no longer backwards incompatible we will gladly look at the new pull request.

PR #720, opened 2016-01-29
Add postgresql_version fact

jyaworski:

PR #740, opened 2016-02-16
This addresses a deficiency in GH-720

jyaworski:
GH-720 introduced a case where only the postgres client was installed on a system, and the fact was returning. This changes it to use the postgres binary instead of psql.

### DIFF
--- a/lib/facter/postgresql_version.rb
+++ b/lib/facter/postgresql_version.rb
@@ -1,8 +1,8 @@
 Facter.add(:postgresql_version) do
   setcode do
-    if Facter::Util::Resolution.which('psql')
-      postgresql_version = Facter::Util::Resolution.exec('psql -V 2>&1')
-      %r{^psql \(PostgreSQL\) ([\w\.]+)}.match(postgresql_version)[1]
+    if Facter::Util::Resolution.which('postgres')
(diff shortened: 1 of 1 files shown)


jyaworski:
Hold on this, there is a problem with the postgresql rpms...

```shell
[root@puppetdb1 ~]# rpm -ql postgresql94-server | grep /bin/postgres$
/usr/pgsql-9.4/bin/postgres
[root@puppetdb1 ~]# which postgres
/usr/bin/which: no postgres in (/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/home/jyaworski/.local/bin:/home/jyaworski/bin)
```

jyaworski:
One idea is to have both psql-client version and psql-server version facts....

jyaworski:
Talking with @hunner there are a couple of options.

1. Clarify in the docs that this is the psql version
2. Also have a postgresql_server_version fact

2 is much more difficult. We can't assume the user is called postgres, and we can't assume trust auth is set up anyway.

jonnytpuppet:
Hi There

Thanks for your submission!

I'm closing this ticket due to inactivity. If you would like to see this feature included, feel free to open a new PR.

Thanks!

PR #742, opened 2016-02-17
Revert "Add postgresql_version fact"

bmjen:
Reverts puppetlabs/puppetlabs-postgresql#720

Due to the complications revealed by #740 , I'm going to revert this features before we can do a 4.7.1 release.

@jyaworski and @hunner can you guys readd this commit to #740 so the feature can be reintroduced after everything is worked out?


jyaworski:
Sure.

PR #751, opened 2016-03-22
Replace depecrated apt::source parameters

amateo:
include_src and key_source parameters are deprecated in favour of
include and key hash parameters

### DIFF
--- a/manifests/repo/apt_postgresql_org.pp
+++ b/manifests/repo/apt_postgresql_org.pp
@@ -10,12 +10,16 @@ include ::apt
     priority   => 500,
   }->
   apt::source { 'apt.postgresql.org':
-    location    => 'http://apt.postgresql.org/pub/repos/apt/',
-    release     => "${::lsbdistcodename}-pgdg",
-    repos       => "main ${postgresql::repo::version}",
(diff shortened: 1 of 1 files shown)


igalic:
this will need an upgrade to .fixtures.yml and metadata.json

it will also mean an major upgrade of the module.

jonnytpuppet:
@igalic in your opinion, why does this need a major upgrade bump to this module? I understand that we need to up the version to support that latest apt, but does that impact functionality so much to postgres that we need to make a major version bump?

Perhaps I'm missing something

Thanks!

jonnytpuppet:
This is also a duplicate of #739

igalic:
@jonnytpuppet because we suddenly depend on a module that did a major bump, because of incompatibilties… we should signal that move.

hunner:
@jonnytpuppet Yeah, migrating a module to a major-bump of apt is a major bump of that module. :/

jonnytpuppet:
Hi There

Thanks for your contribution!

I'm closing this PR due to inactivity and failing tests. Feel free to open a new PR if you would like to see this feature, although I see there is another PR open with the same functionality.

Thanks

PR #753, opened 2016-03-29
The $force parameter is deprecated in concat. Remove it.

jyaworski:
Concat 2.x has deprecated <code>$force</code>, so I was seeing messages like this on my server:

```syslog
2016-03-29 16:36:45,946 WARN  [puppet-server] Scope(Concat[/var/lib/pgsql/9.3/data/pg_ident.conf]) The $force parameter to concat is deprecated and has no effect.
```

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -10,7 +10,7 @@
   "dependencies": [
     {"name":"puppetlabs/stdlib","version_requirement":"4.x"},
     {"name":"puppetlabs/apt","version_requirement":">=1.8.0 <3.0.0"},
-    {"name":"puppetlabs/concat","version_requirement":">= 1.1.0 <3.0.0"}
+    {"name":"puppetlabs/concat","version_requirement":"2.x"}
   ],
   "data_provider": null,
   "operatingsystem_support": [
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -141,7 +141,6 @@ class postgresql::server::config {
     concat { $pg_ident_conf_path:
       owner  => $user,
       group  => $group,
-      force  => true, # do not crash if there is no pg_ident_rules
       mode   => '0640',
       warn   => true,
       notify => Class['postgresql::server::reload'],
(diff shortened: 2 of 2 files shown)


domcleal:
It's deprecated in concat 2.x, but it's valid and used in concat 1.x I believe.  In 1.x I think it defaults to false, while 2.x had the equivalent behaviour of true.

In short, I think you'd need to raise the minimum concat version to 2.x too.

jyaworski:
Is that something that we would be willing to accept? Concat 2.x has been out for some time.

jyaworski:
@domcleal @hunner I've bumped the concat dependency.

metadata.json
```diff
@@ -10,7 +10,7 @@
   "dependencies": [
     {"name":"puppetlabs/stdlib","version_requirement":"4.x"},
     {"name":"puppetlabs/apt","version_requirement":">=1.8.0 <3.0.0"},
-    {"name":"puppetlabs/concat","version_requirement":">= 1.1.0 <3.0.0"}
+    {"name":"puppetlabs/concat","version_requirement":"2.x"}
```
hunner: This would cause this PR to be backwards incompatible (major range bump).

jyaworski: Really?

hunner:
We have an epic ticket for tying all of the "postgresql features we'd like in 5.0.0" together https://tickets.puppetlabs.com/browse/MODULES-3282 so this could be added to that.

jonnytpuppet:
@hunner @bmjen Is this something we'd be looking to merge, given it's a backwards-incompatible change?

Thanks!

bmjen:
@jonnytpuppet No. We can't merge this. When we create a branch/roadmap towards 5.0.0, this can go into that.

tphoney:
@jyaworski 
I have added this PR to the modules epic, so it will be picked up when we start https://tickets.puppetlabs.com/browse/MODULES-3282. For now i will close this PR, and it can be reopened or amended when the work starts. Thanks again for the effort you have put into this !

PR #754, opened 2016-03-30
Adds timestamps into logs by default

xprazak2:
This adds timestamps into logs by default which allows easier debugging.

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -20,6 +20,7 @@ class postgresql::server::config {
   $datadir                    = $postgresql::server::datadir
   $logdir                     = $postgresql::server::logdir
   $service_name               = $postgresql::server::service_name
+  $log_line_prefix            = $postgresql::server::log_line_prefix
 
   if ($manage_pg_hba_conf == true) {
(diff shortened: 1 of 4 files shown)


manifests/server/config.pp
```diff
@@ -115,6 +116,10 @@
     }
 
   }
+  # Allow timestamps in log by default
+  postgresql::server::config_entry {'log_line_prefix':
```
DavidS: Can you shield this by a `if $log_line_prefix` ? I'd expect that this creates an invalid config entry when $log_line_prefix is undef; also some people might want to configure that somewhere else.

DavidS:
If you address the comment, please make sure to rebase on top of https://github.com/puppetlabs/puppetlabs-postgresql/pull/755 , which will clear out the travisci issues.

xprazak2:
I rebased and added an `if` guard.

DavidS:
https://travis-ci.org/puppetlabs/puppetlabs-postgresql/jobs/119995312 has built this successfully, but due to today's github hiccup it didn't register properly.

PR #756, opened 2016-04-05
Adding grant_role

paulhamby:
This adds the ability to grant a group role to a user role. This was mentioned in https://tickets.puppetlabs.com/browse/MODULES-1165 and this is basically the code from the comment in that ticket. It appears to work for me just fine, so hopefully it will work for others as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,6 +323,7 @@ The postgresql module comes with many options for configuring the server. While 
 * [postgresql::server::database_grant](#postgresqlserverdatabase_grant)
 * [postgresql::server::db](#postgresqlserverdb)
 * [postgresql::server::extension](#postgresqlserverextension)
+* [postgresql::server::grant_role](#postgresqlservergrant_role)
 * [postgresql::server::pg_hba_rule](#postgresqlserverpg_hba_rule)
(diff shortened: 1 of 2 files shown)


tphoney:
Hi @paulhamby thanks for putting this PR forward for approval, the code looks good. However there are a few things that are needed before we can merge this work. 
We definitely need tests for this, if you look at #344 it was rejected because there were no tests. 

If you have any questions, please feel free to reach out. Thanks again for the work !!!

jonnytpuppet:
@paulhamby I was just wondering if you had had a chance to look at this?

Thanks!

paulhamby:
I haven't yet, I just need to find the time to write some tests.

thunderkeys:
A simpler 'unless' condition might be something like this:

```
SELECT 1 WHERE pg_has_role('${user}', '${role}', 'MEMBER')
```
We've been using the code from PR #344 for quite some time now, which had this conditional and it's worked great. Also, although we never needed to use the "admin option" that was available in that PR, may be worth looking at including with this PR as well.

thunderkeys:
This can probably be closed with the merging of #762

PR #761, opened 2016-04-28
Don't overwrite or calculate the environment twice

janfabry:
#688 passed the `environment` parameter, but did not notice that it was already calculated. And we should duplicate the `connect_settings`, otherwise we add to it.

Fixes https://tickets.puppetlabs.com/browse/MODULES-3311

### DIFF
--- a/lib/puppet/provider/postgresql_psql/ruby.rb
+++ b/lib/puppet/provider/postgresql_psql/ruby.rb
@@ -30,7 +30,7 @@ Puppet::Type.type(:postgresql_psql).provide(:ruby) do
   private
 
   def get_environment
-    environment = resource[:connect_settings] || {}
+    environment = (resource[:connect_settings] || {}).dup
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/postgresql_psql/ruby.rb
```diff
@@ -16,21 +16,19 @@ def run_sql_command(sql)
...
-      run_command(command, resource[:psql_user], resource[:psql_group], environment)
+      run_command(command, resource[:psql_user], resource[:psql_group])
     end
   end
 
   private
 
   def get_environment
-    environment = resource[:connect_settings] || {}
+    environment = resource[:connect_settings].dup || {}
```
janfabry: This doesn't work when `resource[:connect_settings]` is empty. I guess `environment = environment.dup` on the next line is better, but I'll verify that when I have a test environment set up.

PR #762, opened 2016-04-30
Grant role with unit tests

janfabry:
Added unit tests to `grant_role`, created in #756.

This is the first time I wrote tests for a Puppet module, so all feedback welcome (including on code style). Do you need higher level tests as well?

Further enhancements I want to add to this branch:
 
 - [x] support `ensure => present/absent`, so they can be removed as well (similar to `postgresql::server::extension`)
 - [x] support `$connect_environment`

### DIFF
--- a/manifests/server/grant_role.pp
+++ b/manifests/server/grant_role.pp
@@ -13,4 +13,11 @@ define postgresql::server::grant_role (
     unless    => "select pg_roles.rolname,usename from pg_user join pg_auth_members on (pg_user.usesysid = pg_auth_members.member) join pg_roles on (pg_roles.oid = pg_auth_members.roleid) where pg_roles.rolname ='${group}' and pg_user.usename = '${role}'",
     require   => Class['postgresql::server'],
   }
+
+  if defined(Postgresql::Server::Role[$role]) {
+    Postgresql::Server::Role[$role]->Postgresql_psql["GRANT '${group}' TO '${role}'"]
+  }
+  if defined(Postgresql::Server::Role[$group]) {
+    Postgresql::Server::Role[$group]->Postgresql_psql["GRANT '${group}' TO '${role}'"]
+  }
 }
--- a/manifests/server/grant_role.pp
+++ b/manifests/server/grant_role.pp
@@ -8,6 +8,14 @@ define postgresql::server::grant_role (
   $port             = $postgresql::server::port,
   $connect_settings = $postgresql::server::default_connect_settings,
 ) {
+  validate_string($group)
+  validate_string($role)
+  if empty($group) {
+    fail('$group must be set')
+  }
+  if empty($role) {
+    fail('$role must be set')
+  }
   case $ensure {
     'present': {
       $command = "GRANT \"${group}\" TO \"${role}\""
--- a/manifests/server/grant_role.pp
+++ b/manifests/server/grant_role.pp
@@ -1,0 +1,16 @@
+define postgresql::server::grant_role (
+  $group,
+  $role,
+  $psql_db          = $postgresql::server::default_database,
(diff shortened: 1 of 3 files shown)


jonnytpuppet:
@janfabry Thank you so much for your contribution, it's very much appreciated!

Before we can merge this, we would need the following things:

- Please make sure the tests pass. Are you able to run these locally on your system?
- Please make sure to squash all of your commits in to a single commit

Thanks!

janfabry:
Darn, I only ran `rake spec`, not `rake validate lint` which would have caught that. And I did not notice the final failure on Travis, sorry.

If this passes I'll rebase and squash them in one commit.

manifests/server/grant_role.pp
```diff
@@ -0,0 +1,43 @@
+# Define for granting membership to a role. See README.md for more information
+define postgresql::server::grant_role (
+  $group,
```
bmjen: Thanks for the contribution @janfabry! Can you add some validation for `$group` and `$role` so that they don't get passed in as undef or invalid values? Looks like if there are invalid values passed into either of these parameters, then we will continue to execute a bad psql command.

Thanks!

manifests/server/grant_role.pp
```diff
@@ -0,0 +1,43 @@
...
+  $role,
+  $ensure           = 'present',
+  $psql_db          = $postgresql::server::default_database,
+  $psql_user        = $postgresql::server::user,
+  $port             = $postgresql::server::port,
+  $connect_settings = $postgresql::server::default_connect_settings,
+) {
+  case $ensure {
+    'present': {
+      $command = "GRANT '${group}' TO '${role}'"
```
janfabry: Big mistake from me: the single quotes should be double quotes, as these are identifiers, not strings!

kenazk: Hi there, just got a note from Lindsey to look into this PR. Thanks for contributing! It looks like there are still some open issues that @bmjen raised around validation. Could you address those in the PR?

thunderkeys: Is it just the correction of the single quotes to double quotes (identified by @janfabry), and the validation of $group and $role that this PR is pending on? This diff (untested, but pretty straightforward) might clean up those two remaining issues.

```
--- grant_role.pp.orig 	2016-08-30 19:15:53.000000000 -0700
+++ grant_role.pp      	2016-08-30 19:19:54.000000000 -0700
@@ -8,13 +8,15 @@
   $port             = $postgresql::server::port,
   $connect_settings = $postgresql::server::default_connect_settings,
 ) {
+  validate_string($group)
+  validate_string($role)
   case $ensure {
     'present': {
-      $command = "GRANT '${group}' TO '${role}'"
+      $command = "GRANT \"${group}\" TO \"${role}\""
       $unless_comp = '='
     }
     'absent': {
-      $command = "REVOKE '${group}' FROM '${role}'"
+      $command = "REVOKE \"${group}\" FROM \"${role}\""
       $unless_comp = '!='
     }
     default: {
```

janfabry:
I added the individual code fixes, but I think this really needs acceptance tests, and I don't have any experience with those in Puppet modules. I don't expect myself to have time to invest in learning these in the (near) future, so if somebody else wants to take this over and finish the job, please do so.

janfabry:
O man, learning too much about Puppet and its history with this tiny pull request :-)

janfabry:
@kenazk I updated this PR, not sure whether you got notified. As I said, at the code level I think this PR is okay, but personally I think it needs acceptance tests, which I can't write at the moment. So I'll leave it up to you to accept this or not.

bmjen:
Thanks for the contribution @janfabry ! We will work in a ticket to get the acceptance tests written!

PR #763, opened 2016-05-17
Silence unnecessary warning messages.

blackknight36:
These warnings cause puppet agents to send out a report every time
that puppet runs.  Removed warning messages in order to silence
these reports.

### DIFF
--- a/lib/puppet/provider/postgresql_psql/ruby.rb
+++ b/lib/puppet/provider/postgresql_psql/ruby.rb
@@ -37,13 +37,6 @@ Puppet::Type.type(:postgresql_psql).provide(:ruby) do
         if setting =~ /^(\w+)=((.|\n)+)$/
           env_name = $1
           value = $2
-          if environment.include?(env_name) || environment.include?(env_name.to_sym)
-            if env_name == 'NEWPGPASSWD'
-              warning "Overriding environment setting '#{env_name}' with '****'"
-            else
-              warning "Overriding environment setting '#{env_name}' with '#{value}'"
-            end
-          end
           environment[env_name] = value
         else
           warning "Cannot understand environment setting #{setting.inspect}"


DavidS:
From my reading of the code, these warnings should only be triggered if you have a duplicate entries in the `connect_settings` and the `environment` properties. I do think that in this case a warning is appropriate, as having the same variable defined at different places is really dangerous and confusing. I'll leave this open for now, should I have overlooked something.

blackknight36:
Thanks.  It's possible I am misreading the code however I do not have 
any connect_settings defined in my puppet manifests and the environment 
is using the default production environment.  My biggest issue is that 
these warnings show up every time the agent runs which causes tagmail to 
send out e-mail alerts every hour.

tomhey:
Can you provide an example of one of the warning message produced? I'd be interested to see what the variable change is changing from/ to.

blackknight36:
Sure.  I end up with messages that look like this.

 > 2016-05-11 08:57:48 -0400 Postgresql_psql[CREATE ROLE cats_reader 
ENCRYPTED PASSWORD ****](provider=ruby) (warning): Overriding 
environment setting 'NEWPGPASSWD' with '****'
2016-05-11 08:57:48 -0400 Postgresql_psql[ALTER ROLE cats_reader 
ENCRYPTED PASSWORD ****](provider=ruby) (warning): Overriding 
environment setting 'NEWPGPASSWD' with '****'
2016-05-11 08:57:48 -0400 Postgresql_psql[CREATE ROLE cats ENCRYPTED 
PASSWORD ****](provider=ruby) (warning): Overriding environment setting 
'NEWPGPASSWD' with '****'
2016-05-11 08:57:49 -0400 Postgresql_psql[ALTER ROLE cats ENCRYPTED 
PASSWORD ****](provider=ruby) (warning): Overriding environment setting 
'NEWPGPASSWD' with '****'
2016-05-11 08:57:49 -0400 Postgresql_psql[CREATE ROLE postgres ENCRYPTED 
PASSWORD ****](provider=ruby) (warning): Overriding environment setting 
'NEWPGPASSWD' with '****'
2016-05-11 08:57:49 -0400 Postgresql_psql[ALTER ROLE postgres ENCRYPTED 
PASSWORD ****](provider=ruby) (warning): Overriding environment setting 
'NEWPGPASSWD' with '****'
2016-05-11 08:57:49 -0400 Puppet (notice): Finished catalog run in 7.55 
seconds

This happens every time the agent runs which is also strange since the 
password shouldn't be changing between runs.

tomhey:
It's strange you're getting this warning, as to me it seems to indicate that NEWPGPASSWD is provided both in the environment and (somehow) the connect_settings when Postgresql_psql is called from Postgresql::Server::Role. The code in Role adds NEWPGPASSWD to the environment, but I can't see anywhere the connect_settings are modified.

The fact you get this on every run (even when the password isn't changed) is because the Postgresql_psql resource is used to check the password is already in the correct state, which generates this warning.

I've tried reproducing this on the latest master version (commit id: f040dd2d722141933fba8aac2a9ebb2e9c10eb22) with the following setting and it works correctly for me.

```bash
root@e0301e511e1b:/tmp/test# cat apply postgres-role-test.pp 

include postgresql::server

postgresql::server::role{'a-user':
  password_hash => 'foobar',
}

root@e0301e511e1b:/tmp/test# puppet apply postgres-role-test.pp 
Notice: Compiled catalog for e0301e511e1b in environment production in 1.66 seconds
Notice: /Stage[main]/Postgresql::Client/File[/usr/local/bin/validate_postgresql_connection.sh]/content: content changed '{md5}20301932819f035492a30880f5bf335a' to '{md5}625f6b9f027fbae935499443d1bf69f6'
Notice: /Stage[main]/Main/Postgresql::Server::Role[a-user]/Postgresql_psql[CREATE ROLE a-user ENCRYPTED PASSWORD ****]/command: command changed 'notrun' to 'CREATE ROLE "a-user" ENCRYPTED PASSWORD '$NEWPGPASSWD' LOGIN NOCREATEROLE NOCREATEDB NOSUPERUSER  CONNECTION LIMIT -1'
Notice: Finished catalog run in 2.39 seconds

root@e0301e511e1b:/tmp/test# puppet apply postgres-role-test.pp 
Notice: Compiled catalog for e0301e511e1b in environment production in 1.63 seconds
Notice: Finished catalog run in 2.21 seconds
```

Can you create a self contained example that demonstrates this problem?

Reamer:
@blackknight36 https://github.com/puppetlabs/puppetlabs-postgresql/commit/47ab296f36f66075554b4ccdb7553b0154bedfd0 should fix the problem. With https://github.com/puppetlabs/puppetlabs-postgresql/commit/90c9c670cd1b35fce447399652a9102db08d1427
this warnings are still present

blackknight36:
Thanks.  I will use that commit instead.

PR #765, opened 2016-05-24
(MODULES-3319) Hash argument for create_resources

jcpunk:
Without this patch, seperate modules and classes need to be built
to declare the provided defines for the postgresql server.
This patch permits the passing of a hash into the postgresql::server
function for the various resources.

The puppetlabs-mysql module provides similar behavior whereby
databases can be managed as values from Hiera or an ENC.  This
commit brings the puppetlabs-postgresql module into feature
parity with the puppetlabs-mysql module.


jcpunk:
I believe I'll need some help with the tests...

spec/unit/classes/server_spec.rb
```diff
@@ -157,4 +157,78 @@ class { 'postgresql::globals':
       is_expected.to contain_class('postgresql::repo').with_version('99.5')
     end
   end
+
+  describe 'create config entries' do
+    let(:params) {{ :config_entries => {:config_entry => {:ensure => 'present', :name => 'port_spec', :value => '5432'}}}}
```
domcleal: To fix the tests, try changing the symbols (`:config_entry`, `:ensure` etc) to strings, or replace the whole value assigned to `:config_entries` to a string that's compatible with Puppet's DSL. rspec-puppet passes it through with little editing, and the Ruby symbol syntax isn't valid in a manifest.

jcpunk:
hmmmm, I can't figure out why those two remaining tests are failing....

spec/unit/classes/server_spec.rb
```diff
@@ -157,4 +157,78 @@ class { 'postgresql::globals':
       is_expected.to contain_class('postgresql::repo').with_version('99.5')
     end
   end
+
+  describe 'create config entries' do
+    let(:params) {{ 'config_entries' => {'config_entry' => {'ensure' => 'present', 'name' => 'port_spec', 'value' => '5432'}}}}
+    it 'stops postgresql and changes the port' do
+      is_expected.to contain_file('systemd-override')
```
domcleal: Are these expectations correct?  Shouldn't it check that it's `expected.to contain_postgresql__server__config_entry`?

spec/unit/classes/server_spec.rb
```diff
@@ -157,4 +157,78 @@ class { 'postgresql::globals':
...
+  end
+
+  describe 'create pg_ident_rule entries' do
+    let(:params) {{ 'pg_ident_rules' => {'test' => {'map_name' => 'thatsmymap', 'system_username' => 'systemuser', 'database_username' => 'dbuser', }} }}
+
+    it { is_expected.to contain_concat__fragment('pg_ident_rule_test') }
+  end
+
+  describe 'create recovery entries' do
+    let(:params) {{ 'recovery' => {'test' => {'restore_command' => 'restore_command', 'recovery_target_timeline' => 'recovery_target_timeline' }} }}
```
domcleal: I think you also need to add `'manage_recovery_conf' => true` to the params hash for the recovery feature to activate.

jcpunk:
I picked a less interesting config option (copied from the config test) and still seem to be not right.....

spec/unit/classes/server_spec.rb
```diff
@@ -157,4 +157,77 @@ class { 'postgresql::globals':
       is_expected.to contain_class('postgresql::repo').with_version('99.5')
     end
   end
+
+  describe 'create config entries' do
+    let(:params) {{ 'config_entries' => {'ensure' => 'present', 'name' => 'data_directory_spec', 'value' => '/var/pgsql'}}}
```
domcleal: The layout of this isn't quite right, it's missing a hash in the middle with the resource names. It should probably be:

    let(:params) {{ 'config_entries' => {'data_directory_spec' => {'ensure' => 'present', 'value' => '/var/pgsql'}}}}

jcpunk:
Any ideas why my two remaining tests have failed?

domcleal:
No, but I don't think it's because of this PR, they're failing anyway.

domcleal:
I filed https://github.com/puppetlabs/ruby-hocon/issues/75 for the issue currently causing tests to fail (since the release of specinfra 2.59.1).

jcpunk:
Interesting...  I'd expected the self tests to locate those sorts of things...

Anyway, I've updated the branch with (hopefully) viable fixes

jcpunk:
Something went wrong with my rebase, canceling PR.  I'll make a new one once I've got my stuff fixed....

PR #766, opened 2016-05-26
(MODULES-3416) Add Fedora 24 with PostgreSQL 9.5

domcleal:


domcleal:
I think the test failures on Puppet 4.5.0 are just https://tickets.puppetlabs.com/browse/PUP-6336.

domcleal:
I shouldn't have assumed. I checked into it a little more, and filed https://tickets.puppetlabs.com/browse/PUP-6385 as it appears to be a regression in Puppet 4.5.0.

PR #768, opened 2016-06-01
Avoid unnecessary newlines in recovery.conf template and add .erb suffix

tdevelioglu:
Avoid unnecessary newlines in recovery.conf template and add .erb suffix.

### DIFF
--- a/manifests/server/recovery.pp
+++ b/manifests/server/recovery.pp
@@ -32,7 +32,7 @@ define postgresql::server::recovery(
     # Create the recovery.conf content
     concat::fragment { 'recovery.conf':
       target  => $target,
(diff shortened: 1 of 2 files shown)


bmjen:
Thanks @tdevelioglu ! For the most part, this update looks ok. There appears to be some test failures caused by some template changes. Do you think you can take a look?

tphoney:
@tdevelioglu have you had a chance to look at the issues that @bmjen raised ? it also now needs a rebase. Sorry about that.

bmjen:
Thanks @tdevelioglu . We're going to close this PR due to inactivity. Please reopen this if you want to revisit it and the test failures are addressed. Thanks!

PR #770, opened 2016-06-02
MODULES-3107 fail when datadir is mount point

k2patel:
If datadir is a mount point ( in my case NFS ), module fails during manifests/server/initdb.pp.
Why
it try to create datadir – which already exists because it's mount point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,6 +409,13 @@ Overrides the default PostgreSQL data directory for the target platform. Default
 
 **Warning:** If datadir is changed from the default, Puppet does not manage purging of the original data directory, which causes it to fail if the data directory is changed back to the original.
 
+##### `manage_datadir`
+
+Manage datadir as file resource by this module, default true.
+If false, it will not create the folder, and expect 'datadir' exists and own by the 'user'.
+
+**Warning** If manage_datadir is false, Puppet will still initialize database.
+
 ##### `default_database`
 
 Specifies the name of the default database to connect with. On most systems, this is 'postgres'.
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -55,7 +55,7 @@ class postgresql::globals (
   $manage_recovery_conf     = undef,
 
   $manage_package_repo      = undef,
-  $ismount                  = false,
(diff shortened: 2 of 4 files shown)


tphoney:
@k2patel thanks for putting this forward for approval, the issue you are running into could be worked around by having your postgresql resource depend on the NFS mount: https://docs.puppet.com/puppet/latest/reference/lang_relationships.html

If you want to amend the code:
- The variable name should follow the naming conventions https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/manifests/server.pp#L53-L55
-

tphoney:
In your manifest you could create a dependency chain like (this is a shot in the dark, without the error message:
`File["/data"]-> 
mount { "/data":
        device  => "/dev/sdb1",
        fstype  => "ext4",
	ensure  => "mounted",
	options => "defaults",
        atboot  => "true",
    }->Exec [ 'postgresql_initdb']

postgresql::server{ 'mount_example':
  datadir => '/data'
  othervars => 'uyfiyg'
}`

regarding your code, you can change the variable name in this pr.

k2patel:
i tried to think over it. but can not find easy fix on the situation.
- Since server::initdb has datadir fined as file resource, is always going to conflict with my file resource unless i use switch to control it.

tphoney:
@k2patel if you use references `File["/data"]- it will work IE  https://docs.puppet.com/puppet/latest/reference/lang_resources_advanced.html#amending-attributes-with-a-resource-reference

k2patel:
That make sense now. let me do that... let me close this PR for now.

Thank you.

k2patel:
@tphoney Does this Amending attributes with a resource reference works with puppet 3.x ?

Because i do not find any reference in documentation for that.
So in order to support 3.8 i might still need to have this patch done ?

Please advise.

k2patel:
nvm my bad.. i have found the solution.

Thanx for pointing out.
i had to do is instead of "Before => Class[Postgresql::Server::Initdb]" i have to do Exec['postgresql_initdb'] ---- it was easy than i thought but i just missed it.

PR #774, opened 2016-06-21
Allow username and password as integers

sathieu:
Fixes: String can't be coerced into Fixnum

### DIFF
--- a/lib/puppet/parser/functions/postgresql_password.rb
+++ b/lib/puppet/parser/functions/postgresql_password.rb
@@ -13,6 +13,6 @@ module Puppet::Parser::Functions
     username = args[0]
     password = args[1]
 
-    'md5' + Digest::MD5.hexdigest(password + username)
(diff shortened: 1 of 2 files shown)


sathieu:
See https://github.com/2ndquadrant-it/puppet-barman/pull/29#issuecomment-227210810

bmjen:
Thanks for the contribution @sathieu! Can you fix the acceptance tests failures? It appears that the service is not listening on the port after the change.

sathieu:
The test fail for another reason:
```
Gem::InstallError: mime-types-data requires Ruby version >= 2.0.
```
Nothing to do with my PR.

bmjen:
@sathieu I believe that error was fixed in master by pinning json_pure if ruby < 2.0. Can you rebase from master and push again? Sorry!

sathieu:
@bmjen. Done. CI in progress.

bmjen:
Thanks @sathieu . The unit test failures on puppet 4 are a known issue.

PR #779, opened 2016-07-09
(MODULES-3247) Enable schema and database ownership change

ntpttr:
Currently postgresql::server::schema, postgresql::server::db and
postgresql::server::database allow creating new schemas and
databases if they don't already exist and assigning owners to
them. This patch enables changing the owner of a schema or database
that already exists if the change_ownership variable is set to true.

### DIFF
--- a/README.md
+++ b/README.md
@@ -792,7 +792,7 @@ Defines the value for the setting.
 
 #### postgresql::server::db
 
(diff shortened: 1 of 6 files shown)


ntpttr:
CI failure is unrelated, mime-types/ruby dependency error.

PR #780, opened 2016-07-12
fix include_src deprecation

b4ldr:
fix WARNING: $include_src is deprecated and will be removed in the next major release, please use $include => { 'src' => false } instead


hunner:
Yeah, we need to do this, but it would shift the lower bound of the apt requirement to >= 2.0.0 and that is a breaking change I think. It's not technically specified as part of semver about how dependency changes impact versions, though.

What do you think? Also, this PR will need rebasing as it includes two merge commits.

krzyzakp:
This is also not complete fix of problem:
Warning: Scope(Apt::Source[apt.postgresql.org]): $key_source is deprecated and will be removed in the next major release, please use $key => { 'source' => https://www.postgresql.org/media/keys/ACCC4CF8.asc } instead.


I know it's breaking change for APT dependency, but as many repos are switching to new module version, this should follow as well I think.

PR #784, opened 2016-08-03
mime-types ruby 2.0 dependency

matonb:
mime-types 3.0 introduces a dependency on mime-types-data which requires ruby >= 2.0
Pin mime-types to 2.99.2 if using ruby 1.x

Fixes failing ruby 1.9.3 tests in PR #777


eputnam:
Thanks for the PR! Please resolve conflicts.

eputnam:
this needs a rebase but it looks like these changes have already taken place on master. For future reference, the Gemfile is managed by modulesync which we use to make changes across all of our supported modules. Changes to individual files are therefore not usually merged. You can find the list of managed modules here: https://github.com/puppetlabs/modulesync_configs/blob/master/managed_modules.yml

eputnam:
recovery.pp changes already on master and Gemfile changes won't be accepted as stated previously.

PR #785, opened 2016-08-15
handle case where default database name is not the same as user

hdeadman:
This is to resolve issue: [MODULES-3739](https://tickets.puppetlabs.com/browse/MODULES-3739) 

Without this change you get an error if you override postgresql::globals::user with something other than the postgresql::globals::default_database value.

### DIFF
--- a/manifests/server/passwd.pp
+++ b/manifests/server/passwd.pp
@@ -5,6 +5,14 @@ class postgresql::server::passwd {
   $group             = $postgresql::server::group
   $psql_path         = $postgresql::server::psql_path
   $port              = $postgresql::server::port
+  $database          = $postgresql::server::default_database
+
+  # psql will default to connecting as $user if you don't specify name
+  $_datbase_user_same = $database == $user
+  $_dboption = $_datbase_user_same ? {
+    false => " --dbname ${database}",
+    default => ''
+  }
 
   if ($postgres_password != undef) {
     # NOTE: this password-setting logic relies on the pg_hba.conf being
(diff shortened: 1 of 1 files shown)


manifests/server/passwd.pp
```diff
@@ -5,6 +5,15 @@
   $group             = $postgresql::server::group
   $psql_path         = $postgresql::server::psql_path
   $port              = $postgresql::server::port
+  $database          = $postgresql::server::default_database
+
+  # psql will default to connecting as $user if you don't specify name
+  # dbname option could probably be used unconditionally assuming it is in all supported versions of psql
+  $datbase_user_same = $database == $user
+  $dboption = $datbase_user_same ? {
```
DavidS: * You can shorten that by putting `($database == $user)` directly into the selector.
* Using `$_dboption` will make it a private variable, and avoid unintentional linkage later.

spec/unit/classes/server/config_spec.rb
```diff
@@ -34,8 +34,9 @@
           class { 'postgresql::globals':
             manage_package_repo => true,
             version => '9.4',
-          }->
-          class { 'postgresql::server': }
+            before => Class['Postgresql::Server'],
+          }
+          include postgresql::server
```
DavidS: Those are probably all related to https://tickets.puppetlabs.com/browse/PUP-6608 and should be reverted before merging. 4.6.1 release should be out this week.

manifests/server/passwd.pp
hdeadman: I had ($database == $user) in the selector before but I think it made the travis tests fail on Ruby 1.9. I didn't see any other selectors in the module that used expressions. The expression was working for me and in the other non 1.9 tests.

spec/unit/classes/server/config_spec.rb
hdeadman: @DavidS - OK after 4.6.1 is released I will re-submit the pull request with a single commit and make the changes you suggest. Let me know if there is a way to inline the expression in the selector that will work in Ruby 1.9 & Puppet 3.

DavidS:
Thanks for your patience.

I checked selectors, and you're correct that puppet3 only allows a single variable and not expressions there.

hdeadman:
Would the unit tests re-run with Puppet 4.6.1 instead of 4.6.0 if I closed and re-opened this pull request?

DavidS:
Yes.

PR #786, opened 2016-08-19
Allow working directory to be configurable via globals (instead of always using /tmp)

hdeadman:
When running on a server with /tmp mounted with "noexec" option in fstab (to satisfy security) I was getting an error like: 

Error: Error executing SQL; psql returned 256: 'could not change directory to "/root"

which was resolved for most people back in version 2.2.1 by setting /tmp as the working directory for psql. I need to set the folder to something else now that /tmp is secure so this pull request makes /tmp configurable. 

This pull also has a commit from another pull request I have open that is waiting for 4.6.1 to come out so the tests will pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -453,6 +453,10 @@ Set a prefix for the server logs. Default: `'%t '`
 
 Sets up official PostgreSQL repositories on your host if set to true. Default: false.
 
+##### `module_workdir`
+
+Specifies working directory under which the psql command should be executed. May need to specify if /tmp is on volume mounted with noexec option. Default: /tmp 
+
 ##### `needs_initdb`
 
 Explicitly calls the initdb operation after the server package is installed and before the PostgreSQL service is started. Default: OS dependent.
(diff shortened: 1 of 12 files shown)


hdeadman:
Opened ticket [MODULES-3759](https://tickets.puppetlabs.com/browse/MODULES-3759)

eputnam:
thanks for the PR! please add the new param to the readme and add some tests.

hdeadman:
I updated the documentation and added a couple tests. Let me know if you want tests that cover every class or define where the the attribute is used.

eputnam:
merging. out of curiosity, is there a reason psql needs to be run in a particular dir? does the dir need to be writable?

hdeadman:
@eputnam - I was having problems on a server with /tmp mounted in /etc/fstab with noexec option. There is a security check in the [scap security guide](https://github.com/OpenSCAP/scap-security-guide/blob/8cd9fb09b1fbde47c7376bdf117452d427db787c/RHEL/7/input/xccdf/system/permissions/partitions.xml#L111) that calls for that. I don't recall exactly which part of module was breaking but when I google'd error message it brought me to the release notes for the change that made psql use /tmp to begin with.

PR #788, opened 2016-09-12
(#3858) Fix unless check in grant_role to work with roles as well as users

thunderkeys:
This PR corrects the unless check for postgresql::server::grant_role as well as adding acceptance tests.  Documented originally in https://tickets.puppetlabs.com/browse/MODULES-3858

Please let me know if you have any questions.


eputnam:
Thanks for the PR @thunderkeys !

PR #789, opened 2016-09-13
WIP: (MODULES-3864) Escape shell characters in SQL cmd

kivoli:
This is the possible patch for [#3864](https://tickets.puppetlabs.com/browse/MODULES-3864). This PR is still missing tests, thus I marked it as work in progress.


DavidS:
Changing it at this place breaks backwards compatibility and will upset many existing configurations. Therefore this is not a viable solution.

PR #792, opened 2016-09-21
Improvements around changing owners

hasegeli:


hasegeli:
Any luck looking at this?  It would be better to do so, before releasing the `change_ownership` argument.  The changes only look complicated because of re-indentation.

manifests/server/database.pp
```diff
@@ -2,14 +2,13 @@
...
   $dbname           = $title,
-  $owner            = $postgresql::server::user,
+  $owner            = undef,
   $tablespace       = undef,
   $template         = 'template0',
   $encoding         = $postgresql::server::encoding,
   $locale           = $postgresql::server::locale,
   $istemplate       = false,
   $connect_settings = $postgresql::server::default_connect_settings,
-  $change_ownership = false,
```
DavidS: Removing this parameter is a breaking change that might cause existing configurations to break. It is better for users to accept the parameter, but not act on it, except for emitting a warning that it will go away. Newer versions of stdlib have a `deprecation` method to do so in a friendly way.

hasegeli: I don't think this argument has been released yet.  Do you think it is still necessary to accept it?

hasegeli:
I made the commits really clean by first reverting the existing changing owner implementation.

DavidS:
This looks SO MUCH better! :tada: 

You do seem to have missed one place in the tests where the expected title has changed.

I fully expect @ntpttr be satisfied with the changes too, but he should be aware of them.

hasegeli:
I fixed it.  Thank you for the review.

hasegeli:
Thank you for looking at it.

manifests/server/database.pp
DavidS: For reference, this is a revert of https://github.com/hasegeli/puppetlabs-postgresql/commit/82c6716d4845bde01a03c58b2123b3009c0d97ca which indeed was never released.

PR #794, opened 2016-09-23
Fix SQL style on role.pp

hasegeli:


eputnam:
thanks for the PR @hasegeli !

PR #796, opened 2016-09-29
Move to RHEL repo to download.postgresql.com from yum.postgresql.com

chrisfu:
Hey,

I've switched over to using download.postgresql.org on the advice of Stephen Frost (via Twitter). This solves the 403's you'll be seeing when hitting repomd.xml in any part of the yum.postgresql.org structure right now.

Thanks,

Chris Merrett


chrisfu:
Just FYI, they've added a redirect for now for yum.postgresql.org for now, but it's only temporary.

https://twitter.com/net_snow/status/781525615715418113

devrimgunduz:
Hi,

Please use https instead, as http may also be deprecated.

kbarber:
Hey @DavidS @hunner this was brought to my attention by @mhagander ... who works over at the PGDG (so represents PostgreSQL) and says this is urgent. Sounds like it :-).

Magnus, can the old URL's still work for a while? We have no way of changing this setting in config for older versions of the module, thats all. So this might live on for a bit. Of course, users can still 'decompose' this, and provide their own repo - this module doesn't force you to use our repo class.

kbarber:
@mhagander @chrisfu is this affecting debian also? I mean, is there a new authoritative location for Debian we should be using also here?

manifests/repo.pp
```diff
@@ -8,7 +8,7 @@
       if $version == undef {
         fail("The parameter 'version' for 'postgresql::repo' is undefined. You must always define it when osfamily == Redhat or Linux")
       }
-      class { 'postgresql::repo::yum_postgresql_org': }
+      class { 'postgresql::repo::download_postgresql_org': }
```
kbarber: Hehe, maybe its just best to say 'redhat' or something of that nature, changing the class name each time the download URI changes, might get annoying. I say 'redhat', because it matches the case statement: https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/manifests/repo.pp#L7 ... likewise apt, might be come debian.pp. Nit pick, really.

manifests/repo/download_postgresql_org.pp
```diff
@@ -17,14 +17,14 @@
     $label2 = 'rhel'
   }
 
-  yumrepo { 'yum.postgresql.org':
```
kbarber: This is going to leave behind old yum repo files, unless someone has a purge policy on yum repo:

    resources { "yumrepo": ensure => purged }

Umm, consider a) not renaming the resource - or b) explicitly removing the old resource, by making this old one ensure => absent? Anyone have any other ideas?

kbarber:
Oh, you already pasted the tweet. Still, wouldn't mind a link to an official wording @mhagander

manifests/repo/download_postgresql_org.pp
DavidS: In this case I would favor explicitly removing the old resource.

DavidS: .. or call the repo just "postgresql.org".

mhagander:
There's a redirect in place, yes, so the old URL will keep working for a while. There has been no decision on exactly how long that will be - it was dumped in there really quickly when it was noticed there were a lot of repo configs that had not updated when the repo RPMs updated (and clearly some of them, though probably not all, were puppet-module maintained ones). So there's no real end-date set for that one. However, with the redirects things will of course also be less efficient for the users (extra roundtrips).

For one official link, see https://people.planetpostgresql.org/devrim/index.php?/archives/92-Lots-of-changes-at-yum.PostgreSQL.org.html. The first bulletpoint being the reference to the change being done long ago but many scripts not updated yet.

I notice there is nothing actually on the yum.postgresql.org site. @devrimgunduz - can we get a paragraph in at the top of that page directing people to use the new URLs?

Finally - debian is not affected. Similar changes were made to that repository already at the time we created the postgresql.org hosted repository, so the only old references there will be from people who were testers of that system, never in production.

manifests/repo/download_postgresql_org.pp
kbarber: @DavidS +1 to both suggestions, setting a repo name thats less prone to change like 'postgresql.org' is a good idea, while we ensure the old is purged also ...

DavidS:
I've created https://tickets.puppetlabs.com/browse/MODULES-4051 to get some time dedicated from us to get this cleaned up and merged "really soon now".

hunner:
@DavidS @kbarber @chrisfu #818 has been merged to handle the breaking URL (also see #797 for similar changes). I disagree with changing the name of the repo and/or having an extra ensure absent resource, but would welcome a parameter that allows the URL to be customized. Thanks!

DavidS:
Closing this as the core change has been merged and #820 will be merged soon with the additional functionality.

PR #797, opened 2016-09-30
Allow user to override repository URL

ardichoke:
This enables the user to utilize a different mirror for postgresql downloads. Useful if you host a local mirror.


kbarber:
Looks like you're adding configurability to this branch here: https://github.com/puppetlabs/puppetlabs-postgresql/pull/796 ... which is a great idea. Will also allow users to provide their own mirror, which is generally a good practice these days.

There is still a pending review on 796, so expect some changes of course :-). Let me take a quick pass/review on this though, since this is something I'm curious about anyway.

spec/unit/classes/globals_spec.rb
```diff
@@ -43,21 +43,21 @@
...
       end
-      
+
       it 'should pull in class postgresql::repo' do
         is_expected.to contain_class("postgresql::repo")
       end
 
       it do
-        should contain_yumrepo('yum.postgresql.org').with(
+        should contain_yumrepo('download.postgresql.org').with(
```
kbarber: So, adding a test that passes in a different global for base_repourl, and checks to see if the repo_url is indeed different will prove your solution properly.

manifests/repo/download_postgresql_org.pp
```diff
@@ -16,15 +16,21 @@
     $label1 = 'redhat'
     $label2 = 'rhel'
   }
+  $default_baseurl = "https://download.postgresql.org/pub/repos/yum/${postgresql::repo::version}/${label1}/${label2}-\$releasever-\$basearch"
 
-  yumrepo { 'yum.postgresql.org':
+  $baseurl_real = pick($postgresql::repo::baseurl,$default_baseurl,'undefined')
+  if ( $baseurl_real == 'undefined' ){
```
kbarber: Nit pick really, but there has to be a better value to match against, maybe `undef` itself (https://docs.puppet.com/puppet/latest/reference/lang_data_undef.html)? Merely that a string feels like a bad signal to use.

README.md
```diff
@@ -510,6 +510,10 @@ Path to your `recovery.conf` file.
 
 Sets the proxy option for the official PostgreSQL yum-repositories only. Debian is currently not supported. This is useful if your server is behind a corporate firewall and needs to use proxy servers for outside connectivity.
 
+##### `repo_baseurl`
+
+Sets the baseurl for the PostgreSQL repository. Useful if you host your own mirror of the repository.
```
kbarber: Maybe mention the default will be the official PostgreSQL repos for apt and yum depending on OS? I wouldn't say what they are, just leave that vague perhaps to save future unnecessary edits.

manifests/repo/apt_postgresql_org.pp
```diff
@@ -5,12 +5,18 @@
   #
   # http://www.postgresql.org/download/linux/debian/
   #
+  $default_baseurl = 'http://apt.postgresql.org/pub/repos/apt/'
+  $baseurl_real = pick($postgresql::repo::baseurl,$default_baseurl,'undefined')
+  if ( $baseurl_real == 'undefined' ){
```
kbarber: Same here re: undef

manifests/repo/apt_postgresql_org.pp
```diff
@@ -5,12 +5,18 @@
...
+    fail('No repo baseurl defined or automatically detected.')
```
kbarber: I'm guessing this condition, would really only occur if there was a bug in the pick code. I'm just curious if this check is even necessary, the pick chooses a variable you set directly above it ($default_baseurl) as a backup to the user provided one ... I'm just pondering if this is even just overkill doing an undefined match.

Bugs in the code can be (probably already are) testing with rspec-puppet, at least if we test for a yumrepo resource, with an expected repo url ... we're done here, no runtime test needed.

manifests/repo/download_postgresql_org.pp
```diff
@@ -16,15 +16,21 @@
     $label1 = 'redhat'
     $label2 = 'rhel'
   }
+  $default_baseurl = "https://download.postgresql.org/pub/repos/yum/${postgresql::repo::version}/${label1}/${label2}-\$releasever-\$basearch"
```
kbarber: Hmmm, the tricky bit here will be that a user will have to have a static URL, or their own way of templating it on their own, if this is passed as a string way back in globals ... I guess I'm just not sure if one should allow the full static URL, or ask for a prefix, and we apply these late URL path bits (I mean the label1, label2 magic URL generation). I'm not that sure about how PostgreSQL recommends to mirror, it might be they say preserve their URL's ... or are less specific.

I'm really okay with this as a current strawman and probably would be fine enough to see it merged as it is (more because doing it with prefix only, is less flexible ultimately), just wondering if this has been considered.

ardichoke:
I have submitted a new commit to address your review notes.  I went ahead and removed the runtime undefined checks, as I agree, they were not useful outside of a strange bug.  Added spec tests and updated documentation.

hunner:
Should we leave the title of the yumrepo at `yum.postgresql.org` instead of `download.postgresql.org` ? This introduces a lot of code changes and could result in two yumrepos present on a box, one that points at the old url and one that points at the new one. It doesn't make sense to have `yumrepo { 'yum.postgresql.org': ensure => absent, }` either, because that's poor puppet code style.

ardichoke:
@hunner that change was made in #796, not in my pull request. You should post your feedback on that PR.

hunner:
@ardichoke the change may have also been made in #796, but it is present in this PR as well so you are also responsible for the changes. We merged #818 in the mean time to fix the repo URL without renaming the repository, but would still welcome the ability to override the URL!

ardichoke:
So why didn't you just ask me to rebase my commit off the accepted PR in the first place instead of requesting that I make a change to something that came from the upstream commit to my PR? What purpose would it serve for me to make the requested change when a different PR than my upstream was accepted for this? Making this change now would be pointless as it's based off a different branch of the code, I might as well just rewrite this PR off of the accepted branch instead which I really don't have time to do right now.

hunner:
I understand your frustration at having to essentially rewrite the PR as a rebase would be difficult. 

#818 fixed the download url problem with minimal other impacts so I went with that route to fix the problem and avoid breaking anything in the mean time. We really do appreciate the work that you all do on the modules!

PR #798, opened 2016-10-05
Remove duplicate RedHat default section

jesusaurus:
A duplicate default_version default section was added in
https://github.com/puppetlabs/puppetlabs-postgresql/commit/1eb45d07f163ab3e3fd117d9de4bc4351b228853.

This removes the duplicate default section.


hunner:
How is this not an error o.o

hunner:
Filed https://tickets.puppetlabs.com/browse/PUP-6930

PR #801, opened 2016-10-12
Fixed unless check in grant_role so that it can grant roles to superusers

crispygoth:
This is my fix for MODULES-3967, an issue where grant_role did not work with superusers.

### DIFF
--- a/manifests/server/grant_role.pp
+++ b/manifests/server/grant_role.pp
@@ -32,7 +32,7 @@ define postgresql::server::grant_role (
 
   postgresql_psql { "grant_role:${name}":
     command          => $command,
-    unless           => "SELECT t.count FROM (SELECT count(*) FROM pg_roles AS r_role JOIN pg_auth_members AS am ON (r_role.oid = am.member) JOIN pg_roles AS r_group ON (r_group.oid = am.roleid) WHERE r_group.rolname = '${group}' AND r_role.rolname = '${role}') AS t WHERE t.count ${unless_comp} 1",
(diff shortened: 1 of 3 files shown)


manifests/server/grant_role.pp
```diff
@@ -32,7 +32,7 @@
 
   postgresql_psql { "grant_role:${name}":
     command          => $command,
-    unless           => "SELECT 1 WHERE pg_has_role('${role}', '${group}', 'MEMBER') ${unless_comp} true",
+    unless           => "SELECT t.count FROM (SELECT count(*) FROM pg_roles AS r_role JOIN pg_auth_members AS am ON (r_role.oid = am.member) JOIN pg_roles AS r_group ON (r_group.oid = am.roleid) WHERE r_group.rolname = '${group}' AND r_role.rolname = '${role}') AS t WHERE t.count ${unless_comp} 1",
```
hasegeli: Something like `SELECT 1 WHERE EXISTS(...) ${unless_comp} true` as before would look nicer in here.

No need for parentheses on ON clauses.

spec/acceptance/server/grant_role_spec.rb
```diff
@@ -59,4 +59,63 @@ class { 'postgresql::server': }
...
+        class { 'postgresql::server': }
+
+        # Since we are not testing pg_hba or any of that, make a local user for ident auth
+        user { $user:
+          ensure => present,
+        }
+
+        postgresql::server::role { $user:
+          password_hash => postgresql_password($user, $password),
+	  superuser     => true,
```
hasegeli: Missing indentation

PR #805, opened 2016-10-19
Fix deprecated apt::source parameters

adepretis:
`puppetlabs/apt` 2.0.0 has been release well over a year ago and deprecated some parameters for `apt::source` (see [Changelog](https://forge.puppet.com/puppetlabs/apt/changelog#aptsource)).

In `puppetlabs/apt` version 2.0.0 these parameters (`key_source`, `include_src`, ...) have been completely removed, in version >= 2.1.0 using these results in a deprecation warning.

Isn't it time to finally implement these changes and require `puppetlabs/apt` >= 2.0.0? ;-)


hunner:
This is a great change, but being backwards compatible would want to go in when the other changes for "postgresql next" go in (as per https://tickets.puppetlabs.com/browse/MODULES-3282). And we just don't have the bandwidth to handle this :(

bmjen:
@adepretis This change will get slotted for the next major release of this module. Increasing the lower bound to a dependency creates an backwards incompatibility for users still on puppetlabs-apt <2.0 and puppetlabs-postgresql 4.x

hunner:
Closes #780

PR #806, opened 2016-10-19
(MODULES-1850) replace validate_string by validate_re for real verification on input

jhooyberghs:
(MODULES-1850) replace validate_string by validate_re for real verification on input


DavidS:
The last cell is not failing due to anything in this PR, and it is very unlikely that anything here would break on ruby 1.9.

PR #809, opened 2016-10-21
Fixed unless check in grant_role so that it can grant roles to superusers 

crispygoth:
Updated version of PR#801 (sorry, I had a bit of a git fail and accidentally closed the last PR)

This is my fix for MODULES-3967, an issue where grant_role did not work with superusers.


tphoney:
great work, thanks for this @crispygoth !

PR #810, opened 2016-10-26
Add support for Gentoo-based operatingsystems

optiz0r:
- Add params.pp support for osfamily=Gentoo
- Make `postgresql::client` management of `postgresql-client` package
  optional. (Gentoo does not split client/server/devel/contrib packages).
  Add spec test for this new behaviour
- Ensure ::server::contrib and ::lib::devel fail on gentoo for same reason
- Extend the systemd-override to Gentoo to propagate port and datadir changes

Fixes https://tickets.puppetlabs.com/browse/MODULES-4011

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -171,6 +171,30 @@ class postgresql::params inherits postgresql::globals {
       $psql_path              = pick($psql_path, '/usr/bin/psql')
     }
 
+    'Gentoo': {
+      $user                = pick($user, 'postgres')
+      $group               = pick($group, 'postgres')
+
+      $client_package_name  = pick($client_package_name, 'UNSET')
+      $server_package_name  = pick($server_package_name, 'postgresql')
+      $contrib_package_name = pick_default($contrib_package_name, undef)
+      $devel_package_name   = pick_default($devel_package_name, undef)
+      $java_package_name    = pick($java_package_name, 'jdbc-postgresql')
+      $perl_package_name    = pick($perl_package_name, 'DBD-Pg')
+      $plperl_package_name  = undef
+      $python_package_name  = pick($python_package_name, 'psycopg')
+
+      $service_name         = pick($service_name, "postgresql-${version}")
+      $bindir               = pick($bindir, "/usr/lib/postgresql-${version}/bin")
+      $datadir              = pick($datadir, "/var/lib/postgresql/${version}_data")
+      $confdir              = pick($confdir, "/etc/postgresql-${version}")
+      $service_status       = pick($service_status, "systemctl status ${service_name}")
+      $service_reload       = "systemctl reload ${service_name}"
+      $psql_path            = pick($psql_path, "${bindir}/psql")
+
+      $needs_initdb         = pick($needs_initdb, true)
+    }
+
     'FreeBSD': {
       $link_pg_config       = true
       $user                 = pick($user, 'pgsql')
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -166,27 +166,28 @@ class postgresql::server::config {
     }
   }
 
-  if $::osfamily == 'RedHat' {
-    if $::operatingsystemrelease =~ /^7/ or $::operatingsystem == 'Fedora' {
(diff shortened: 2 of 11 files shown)


manifests/params.pp
```diff
@@ -171,6 +171,30 @@
       $psql_path              = pick($psql_path, '/usr/bin/psql')
     }
 
+    'Gentoo': {
+      $user                = pick($user, 'postgres')
+      $group               = pick($group, 'postgres')
+
+      $client_package_name  = pick($client_package_name, 'UNSET')
```
hunner: Why is this `pick(... 'UNSET')` instead of `pick_default(... undef)`?

templates/systemd-override.erb
```diff
@@ -1,8 +1,14 @@
-<%- if scope.lookupvar('::operatingsystem') == 'Fedora' -%>
+<%- if scope.lookupvar('::osfamily') == 'Gentoo' -%>
+.include /usr/lib64/systemd/system/<%= @service_name %>.service
+<%- elsif scope.lookupvar('::operatingsystem') == 'Fedora' -%>
 .include /lib/systemd/system/<%= @service_name %>.service
-<% else -%>
```
hunner: Why remove the blank line here?

manifests/server/config.pp
```diff
@@ -166,27 +166,28 @@
...
-        content => template('postgresql/systemd-override.erb'),
-        notify  => [ Exec['restart-systemd'], Class['postgresql::server::service'] ],
-        before  => Class['postgresql::server::reload'],
-      }
-      exec { 'restart-systemd':
-        command     => 'systemctl daemon-reload',
-        refreshonly => true,
-        path        => '/bin:/usr/bin:/usr/local/bin'
-      }
+  if $::osfamily == 'Gentoo' or
```
bmjen: @optiz0r IMO a better way, although not particularly DRY, would be to leave the RedHat block as is, and copy/paste the content into a separate Gentoo block. This way it prevents any RedHat maintenance from impacting Gentoo and vice versa.

templates/systemd-override.erb
optiz0r: Simple oversight, will fix this.

manifests/params.pp
optiz0r: I couldn't persuade puppet (4.6) do the right thing with `pick_default(..., undef)`.

```puppet
class client (
  $package_name   = $postgresql::params::client_package_name,
) {
   if $package_name {                                                 
     package { 'postgresql-client':
       name   => $package_name,
     }
   }
```
This always lead the package to be included in the catalog with a null name. I'm guessing this has something to do with the way it's passed through the default argument from `params.pp`?

I've seen the $x='UNSET', if $x != 'UNSET' idiom used in relation to `params.pp` in the past.

While we're on the subject, can anyone tell me how this pattern is supposed to be working?
```puppet
$server_package_name  = pick($server_package_name, 'postgresql')
```
If I understand the variable scoping correctly `$server_package_name` will never have a value before params.pp is evaluated (unless set in top or node scope, neither of which is considered best practice), so the first case of the `pick()` expression should always be null. And if it did have a value, this line of code would be attempting to change the value of a variable which is forbidden because all variables in puppet DSL are really constants? If anyone wants to override the value of these variables, they would do so one the `postgres::server` or `postgres::globals` classes, so is this left over from a refactoring, or have I missed something about how this really works?

optiz0r:
I'll rebase squash these commits once everyone's happy with the final code, in the meantime I'll leave them separate if that helps with seeing what's changed as a result of the reviews.

manifests/params.pp
DavidS: To answer the easy question first: the `$server_package_name` within the pick call is resolving to `$::postgresql::globals::server_package_name` as that is inherited into this class. The assignment happens in the local scope, setting `$::postgresql::params::server_package_name`, hiding the value from the globals. This chaining allows the user to override values at any of the levels they desire.

The issue with using `undef` is that it will reset an argument to the parameter default. That's why you need to use something like `UNSET`.

PR #813, opened 2016-11-07
Fix grant documentation

hasegeli:
The previous list of possible privileges were wrong.  Also "ALL" is not a synonym for "CREATE".  While fixing them, point out that the arguments accept a comma-seperated list.  Also, add the missing "grant" resource to the table of contents.


wilson208:
@hasegeli can you be able to outline why these changes are necessary in the description?

Thanks :)

hasegeli:
Done

PR #816, opened 2016-11-12
Change default user, group and data directory to proper value on Free…

olevole:
…BSD platform for PostgresQL 9.6: see https://svnweb.freebsd.org/ports?view=revision&revision=r421360 and https://github.com/freebsd/freebsd-ports/blob/master/UPDATING#L314 for details

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -172,10 +172,20 @@ class postgresql::params inherits postgresql::globals {
     }
 
     'FreeBSD': {
-      $link_pg_config       = true
-      $user                 = pick($user, 'pgsql')
-      $group                = pick($group, 'pgsql')
+      case $version {
+        '96': {
+          $user                 = pick($user, 'postgres')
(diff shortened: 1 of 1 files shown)


binford2k:
A better reference link, https://github.com/freebsd/freebsd-ports/blob/66ea117334e2d54d5746d0099f8c087259c929b6/UPDATING#L318-L333

hunner:
@olevole The first line of a git commit message should be about 50 characters, and no more than 72 characters: https://github.com/phonegap/phonegap/wiki/Git-Commit-Message-Format . Any longer description should be in the 3rd or further lines.

Also, linking to the freebsd-ports UPDATING link should be to a specific hash as @binford2k gave, not master (because master branch moves lines around)

The code looks good so this is just for future reference. Thanks!

olevole:
Ok, I took note of comments. Thank you!

PR #808, opened 2016-10-21
Added the functionality to set the timezone and fixed the test setup (Circle-Ci is running successfully)

Faffnir:
https://api.travis-ci.org/Faffnir/puppetlabs-postgresql.svg?branch=master

PR #817, opened 2016-11-17
Added timezone as an additional configuration feature

hunner:
Pick of https://github.com/puppetlabs/puppetlabs-postgresql/pull/808


DavidS:
@hunner There was a bit of docs in the original PR, was that not enough?

PR #823, opened 2016-11-29
Support granting SELECT and UPDATE permission on sequences (MODULES-4158)

chris-reeves:
Add support for granting SELECT and UPDATE permissions on sequences (MODULES-4158).
    
This requires a significant change to the unless statement used to determine whether the permission has already been granted. The previous statement used the standard SQL information_schema tables, however these provide limited information on sequence permissions. In order to retrieve the current permission set with the necessary level of detail we must query the pg_class system catalog and parse out the permissions from there.

As part of this PR, I've added additional acceptance tests to verify both old and new functionality. All tests pass on my local system (using supplied Gemfile).

### DIFF
--- a/spec/acceptance/server/grant_spec.rb
+++ b/spec/acceptance/server/grant_spec.rb
@@ -82,6 +82,40 @@ describe 'postgresql::server::grant:', :unless => UNSUPPORTED_PLATFORMS.include?
         end
       end
     end
+
+    it 'should grant update on a sequence to a user' do
+      begin
+        pp = pp_setup + <<-EOS.unindent
+
+          postgresql_psql { 'create test sequence':
+            command   => 'CREATE SEQUENCE test_seq',
+            db        => $db,
+            psql_user => $owner,
+            unless    => "SELECT 1 FROM information_schema.sequences WHERE sequence_name = 'test_seq'",
+            require   => Postgresql::Server::Database[$db],
+          }
+
+          postgresql::server::grant { 'grant update on test_seq':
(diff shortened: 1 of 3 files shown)


spec/acceptance/server/grant_spec.rb
```diff
@@ -0,0 +1,190 @@
+require 'spec_helper_acceptance'
+
+describe 'postgresql::server::grant:', :unless => UNSUPPORTED_PLATFORMS.include?(fact('osfamily')) do
+
+  db = 'grant_priv_test'
+  owner = 'psql_grant_priv_owner'
+  user = 'psql_grant_priv_tester'
+  password = 'psql_grant_role_pw'
+
+  pp_setup = <<-EOS.unindent
```
hunner: Variables should not be declared outside of tests. If it is needed by multiple tests then a `let()` should be used. Yes, our code does the same thing for `db`/`owner`/`user`/`password` and it's wrong and needs fixing :)

hunner:
This looks great! Could you add the additional values to https://github.com/puppetlabs/puppetlabs-postgresql#privilege-1 as well?

spec/acceptance/server/grant_spec.rb
chris-reeves: This was done for consistency with other spec tests (as you spotted).

Happy to change to something more sane if you want though - just let me know.

chris-reeves:
As for the docs, those additional values fall under "`object_type` dependent string". Unless we break out the privileges available for each object type.

PR #827, opened 2016-12-12
Change - Ensure that the postgres data dir has proper selinux context

blackknight36:
The postgresql data directory needs to be set to the postgresql_db_t
type.  The selinux context was being set to 'etc_t' by default.

### DIFF
--- a/manifests/server/initdb.pp
+++ b/manifests/server/initdb.pp
@@ -22,30 +22,43 @@ class postgresql::server::initdb {
     cwd        => $module_workdir,
   }
 
+  if $::osfamily == 'RedHat' and $::selinux == true {
+    $seltype = 'postgresql_db_t'
+    $logdir_type = 'postgresql_log_t'
+  }
+
+  else {
+    $seltype = undef
+    $logdir_type = undef
+  }
+
   # Make sure the data directory exists, and has the correct permissions.
   file { $datadir:
-    ensure => directory,
-    owner  => $user,
-    group  => $group,
-    mode   => '0700',
+    ensure  => directory,
+    owner   => $user,
+    group   => $group,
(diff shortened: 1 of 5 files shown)


tphoney:
Can you please have a look at the failing tests, we cant merge if the tests are failing.

blackknight36:
Are tests being run with SELinux enabled?  Based on the errors it looks like puppet isn't able to access the postgres data directory.

tphoney:
the acceptance tests work, there is code to enable selinux. it is only failing on unit tests for puppet 4 and above. the ::selinux variable is not set. This could be fixed by explicitly setting it in the tests.

blackknight36:
Does rspec puppet treat boolean facts as a string?  The spec_helper_acceptance.rb file has an if statement that appears to be treating $selinux as a string.

```
# Set up selinux if appropriate.
    if fact('osfamily') == 'RedHat' && fact('selinux') == 'true'
      pp = <<-EOS
        if $::osfamily == 'RedHat' and $::selinux == 'true' {
          $semanage_package = $::operatingsystemmajrelease ? {
            '5'     => 'policycoreutils',
            default => 'policycoreutils-python',
          }

          package { $semanage_package: ensure => installed }
          exec { 'set_postgres':
            command     => 'semanage port -a -t postgresql_port_t -p tcp 5433',
            path        => '/bin:/usr/bin/:/sbin:/usr/sbin',
            subscribe   => Package[$semanage_package],
          }
        }
      EOS

      apply_manifest_on(agents, pp, :catch_failures => false)
    end
```
I believe true should not be quoted here since it is a boolean value.

blackknight36:
I've updated the PR and tests are passing now.

PR #830, opened 2017-01-02
Remove unnecessary `client_package_name` parameter from server.

tdevelioglu:
This was missed during some re-factor a long time back.


tphoney:
Nice catch ! thanks for the fix.

PR #831, opened 2017-01-05
Remove deprecated force parameter from concat

ekohl:
Prevents warnings like:

Scope(Concat[/var/lib/pgsql/data/pg_ident.conf]): The $force parameter to concat is deprecated and has no effect

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -154,7 +154,6 @@ class postgresql::server::config {
     concat { $pg_ident_conf_path:
       owner  => $user,
       group  => $group,
-      force  => true, # do not crash if there is no pg_ident_rules
       mode   => '0640',
       warn   => true,
       notify => Class['postgresql::server::reload'],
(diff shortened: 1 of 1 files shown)


eputnam:
@ekohl thanks for the PR! Since the postgresql module still supports concat as low as 1.1.0, we're hesitant to merge this immediately. I should know more very soon, though, so hang tight.

ekohl:
I didn't consider that so it's fine to keep this open until it's dropped.

eputnam:
We'll probably leave this open until we drop Puppet 3 support from puppetlabs-concat, which shouldn't actually be that long.

mdular:
+1 👍

dpash:
This is now an error.

    Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Concat[/etc/postgresql/9.5/main/pg_ident.conf]: has no parameter named 'force' at /etc/puppetlabs/code/environments/production/external-modules/postgresql/manifests/server/config.pp:154 on node host.example.com

Vinclame:
Nice. I updated puppetlabs/concat to 4.0.0 and I'm using postgres from puppetforge (with this fix apparently). When can I expect an update for this on puppetforge?

PR #833, opened 2017-01-11
Fix validation script to work with dash

sathieu:
Followup fro #760.


sathieu:
Error with dash is:
```
1++: not found
```

hunner:
This will thankfully go away with https://tickets.puppetlabs.com/browse/MODULES-1394

PR #843, opened 2017-02-13
Change - Remove deprecated force parameter on concat resources

blackknight36:
The $force parameter to concat is deprecated and has no effect.  This commit silences warning messages in the puppet server logs.  For example:

> 2017-02-13 08:32:21,798 WARN  [qtp1532802229-36820] [puppetserver] Scope(Concat[/var/lib/pgsql/data/pg_ident.conf]) The $force parameter to concat is deprecated and has no effect.


wilson208:
Doesn't change current functionality since the param has no affect within the concat module. All this will affect are logs. LGTM 👍

PR #846, opened 2017-02-21
Allowo to disable managing passwords for users

bjoernhaeuser:


bjoernhaeuser:
The problem is that `pg_shadow` and `pg_authid` is not accessible inside RDS postgres instance. Nonetheless creating roles with a passwords works, but a second run would fail because it cannot verify that it has to change the password. 

The other option would be to always alter the password, but this sounds even more sketchy to me.

bjoernhaeuser:
Rebased to current master.

README.md
```diff
@@ -1445,6 +1445,9 @@ Default value: `true`.
 ##### `password_hash`
 Sets the hash to use during password creation. If the password is not already pre-encrypted in a format that PostgreSQL supports, use the `postgresql_password` function to provide an MD5 hash here, for example:
 
+##### `refresh_password`
+If set to true, refreshes the password on changes. Set this to false if your user is not allowed to view current passwords of user.
```
hunner: I think the name of `update_password` would make more sense, as setting this to false would stop puppet from trying to update the password even when it IS able to retrieve the password hashes from the database.

Also, reflecting that this stops puppet from changing the password even if it *can* retrieve them, could you update the docs to reflect a use case of something like "Set this to false to not modify the role's password after creation."

bjoernhaeuser:
@hunner changed! Thank you for reviewing.

PR #848, opened 2017-02-28
Update $allowed_auth_methods for newer PostgreSQL versions 9.5 and 9.6

ardrigh:
New auth-method `bsd` was added with PostgreSQL 9.6:
https://www.postgresql.org/docs/9.6/static/auth-pg-hba-conf.html


eputnam:
thanks @ardrigh !

PR #851, opened 2017-03-11
fix default params for SUSE family OSes

tampakrap:
- remove the old openSUSE 13.2
- add default value '94' to cover all the recent and supported openSUSE distros
- use the old initscript for SLE 11 SPx, systemd commands for anything else


wilson208:
👍 LGTM, thanks!

PR #854, opened 2017-03-20
Align postgis default version with postgres version

georgehansper:
As the values for the postgis version is hard-coded, this module currently installs by default older, unsupported versions of postgis.
This commit updates the module to align with current default postgis versions for each related postgres version.


wilson208:
👍 LGTM

PR #855, opened 2017-03-21
(MODULES-4558) Add support for --data-checksums on initdb

atmoz:
> We want to use checksums on data pages. This has to be enabled when running initdb. Currently, the puppetlabs/postgresql module does not seem to support this.

https://tickets.puppetlabs.com/browse/MODULES-4558

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,6 +411,12 @@ Overrides the default PostgreSQL data directory for the target platform. Default
 
 **Warning:** If datadir is changed from the default, Puppet does not manage purging of the original data directory, which causes it to fail if the data directory is changed back to the original.
 
+##### `data_checksums`
+
+Use checksums on data pages to help detect corruption by the I/O system that would otherwise be silent. Valid values: 'true' or 'false'. Default: initdb's default ('false').
+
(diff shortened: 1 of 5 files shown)


polpot78:
Could you rebase this into a single commit?

atmoz:
Will do. I also added the new option `data_checksums` to `README.md`

mmoll:
is this the same functionality as #878?

atmoz:
@mmoll Yes

So this PR have been open almost half a year (also used in production), and then this similar PR (#878) gets merged after just 3 days?

Looks like the other PR misses the [version check](https://github.com/atmoz/puppetlabs-postgresql/blob/2137844cb54f644f1fb7a8a0c56c740f0a16a588/manifests/server/initdb.pp#L78-L81) as well.

Come on guys :thumbsdown:

mmoll:
@atmoz could you rebase/rework that PR so the additional check can get in?

mmoll:
got in via GH-915, so this PR can get closed

tphoney:
Sorry about the slowness. Thanks for your patience @atmoz and apologies.

PR #861, opened 2017-03-29
Fix placement of chaining arrow for style guide/linting

domcleal:
Corrects lint warning under puppet-lint 2.2.0.


domcleal:
Fixed in f385d2cf.

PR #862, opened 2017-04-10
(MODULES-4682) Pass default_connect_settings to validate service

vinzent:
If using CentOS or RedHat Enterprise Linux with PostgreSQL from
SoftwareCollections the validation of the service fails because
it needs the `LD_LIBRARY_PATH` environment variable to work properly.

To pass the env for example for postgresql_psql one can only set

```
class {'postgresql::globals':
  default_connect_settings => {
    'LD_LIBRARY_PATH' => '/opt/rh/rh-postgresql95/root/usr/lib64',
}
```

This patch passes the default_connect_settings to the
`validate_db_connection` resource in `postgresql::service`

### DIFF
--- a/manifests/server/service.pp
+++ b/manifests/server/service.pp
@@ -10,6 +10,7 @@ class postgresql::server::service {
   $port             = $postgresql::server::port
   $default_database = $postgresql::server::default_database
   $psql_path        = $postgresql::params::psql_path
+  $connect_settings = $postgresql::server::default_connect_settings
 
   anchor { 'postgresql::server::service::begin': }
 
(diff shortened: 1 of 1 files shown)


manifests/server/service.pp
```diff
@@ -33,6 +34,7 @@
         run_as          => $user,
         database_name   => $default_database,
         database_port   => $port,
+        connect_settings => $connect_settings,
```
eputnam: @vinzent , looks like the linter is complaining about the hash rocket alignment in this resource. please make sure they all line up.

vinzent:
@eputnam rebased and updated

eputnam:
gah, now a rebase and it should be good @vinzent

PR #865, opened 2017-05-11
Let listen_addresses be defined independently

hasegeli:
Postgresql module is often extended by other shared modules.  Shared modules should not use resource-like class definitions to let the callers use them when necessary.  It is still a reasonable behaviour for the shared module to set max_connections.  Making it optional would allow this to happen.

Removing the default value "localhost" should not break anything as it is already the default of PostgreSQL and on every distribution I know of.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -2,7 +2,7 @@
 class postgresql::params inherits postgresql::globals {
   $version                    = $postgresql::globals::globals_version
   $postgis_version            = $postgresql::globals::globals_postgis_version
-  $listen_addresses           = 'localhost'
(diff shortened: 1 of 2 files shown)


hunner:
Of note, if the `listen_address` is not configured at all then postgresql will only create a unix socket. If it is present, it will create a tcp socket as well. That makes this change backwards-incompatible.

hasegeli:
> Of note, if the listen_address is not configured at all then postgresql will only create a unix socket. If it is present, it will create a tcp socket as well. That makes this change backwards-incompatible.

I don't think so.  It is defaulted to "localhost" when not set.

PR #867, opened 2017-05-11
(MODULES-4906) Add support for concat 3.0.0 and 4.0.0

dhollinger:
Prevent `puppet module list` and `puppet module install` from complaining about invalid installations of postgresql and concat due to dependency mismatch. Goal is to reduce confusion among users that may get confused by those messages.


bastelfreak:
Can we get this merged rather sooner? Several of our newly written acceptance tests at Vox Pupuli fail because the puppet module tool can't resolve the dependencies.

bastelfreak:
also related: https://github.com/puppetlabs/puppetlabs-postgresql/pull/869

PR #868, opened 2017-05-13
Maintain config entries for PostgreSQL 10

hasegeli:


eputnam:
@hasegeli `unix_socket_directory` is valid until 9.3 (https://www.postgresql.org/docs/9.2/static/runtime-config-connection.html#GUC-UNIX-SOCKET-DIRECTORY). Since we will probably be supporting 9+ in the new release, that should stay. otherwise, this looks great!

hasegeli:
Good catch.  I adjusted the commits.

HelenCampbell:
Thanks @hasegeli !

PR #873, opened 2017-05-25
(maint) Fix docs for #870

hunner:


hasegeli:
I am sorry for not maintaining the documentation.

Maybe we should silently deprecate those options by not document anymore.  Is there any advantage of using them over `postgresql::server::config_entry`?

PR #874, opened 2017-05-26
(maint) Add 'os' facts used by apt module

domcleal:
The 'os' facts are used by puppetlabs/puppetlabs-apt@33d17cf in the apt
module, so need populating in addition to legacy facts.


eputnam:
@domcleal sorry for the churn on this, it's been fixed. thanks for the PR!

PR #877, opened 2017-05-26
Cleanup README

hasegeli:
This silently deprecates recently made not-required arguments by not documenting them anymore.  config_entry define is better than using them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -495,12 +495,6 @@ Overrides the default PostgreSQL log directory.
 
 Default value: initdb's default path.
 
-##### `log_line_prefix`
-
-Set a prefix for the server logs.
-
-Default value: `undef`.
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -496,12 +496,6 @@ Overrides the default PostgreSQL log directory.
 
 Default value: initdb's default path.
 
-##### `log_line_prefix`
```
eputnam: there's still a `log_line_prefix` param in globals, did you  mean to remove this here?

hasegeli: Yes.  It is still there, but now it is possible to use `server::config_entry` instead of setting it in here like all other configuration variables.  I meant to silently deprecate this usage by not documenting it anymore.

PR #878, opened 2017-06-01
add data_checksums option to initdb

tjikkun:


manifests/server/initdb.pp
```diff
@@ -82,11 +83,17 @@
       $require_before_initdb = [$datadir]
     }
 
-    $initdb_command = $locale ? {
+    $ic_locale = $locale ? {
```
tphoney: what is the meaning of this variable name, it is a little terse. was there any reason to rename it ?

tphoney:
Looks pretty good otherwise.

manifests/server/initdb.pp
tjikkun: Well, it was like this:
```
$ic_base = "${initdb_path} --encoding '${encoding}' --pgdata '${datadir}'"
$ic_xlog = $ic_base + optional xlogdir
$initdb_command = $ic_xlog + optional locale
```
Now I changed it so:
```
$ic_base = "${initdb_path} --encoding '${encoding}' --pgdata '${datadir}'"
$ic_xlog = $ic_base + optional xlogdir
$ic_locale = $ic_xlog + optional locale
$initdb_command = $ic_locale + optional data-checksums
```
Basically ic_ (probably short for initdb_command) was there before, I just kept the same convention.

tphoney:
thanks for the work @tjikkun :)

PR #879, opened 2017-06-08
(MODULES-1394) replace validate_db_connection type with custom type

eputnam:
This replaces the defined type `validate_db_connection` with a new
custom type called `postgresql_conn_validator`. It's functionality is
nearly a copy of its predecessor with the exception of the
`create_db_first` parameter which is taken care of with a collector in
service.pp. The old type is still intact but all docs have been removed
and a warning has been attached to it announcing its deprecation.


eputnam:
@hasegeli , it would be awesome to have your input on this

manifests/validate_db_connection.pp
```diff
@@ -63,6 +65,9 @@
     $env = $pass_env
   }
 
+  warning($validate_cmd)
+  warning($env)
+
```
hasegeli: Debugging?

lib/puppet/type/postgresql_conn_validator.rb
```diff
@@ -0,0 +1,82 @@
...
+    desc 'The DNS name or IP address of the server where PostgreSQL should be running.'
+  end
+
+  newparam(:port) do
+    desc 'The port that the PostgreSQL server should be listening on.'
+
+    validate do |value|
+      if value
+        value =~ /[0-9]+/
+      end
```
hasegeli: Maybe it is better to check this as an integer.

lib/puppet/util/postgresql_validator.rb
```diff
@@ -0,0 +1,64 @@
+module Puppet
+  module Util
+    class PostgresqlValidator
+      attr_reader :resource
+
+      def initialize(resource)
+        @resource = resource
+      end
+
+      def build_psql_cmd
```
hasegeli: Postgres has the `PQping()` command on its binary protocol to issue a connection check.  It can be used with `pg_isready` binary.  I think it would make a good default provider.  It would be more efficient and easier to use, because you don't need to pass authentication.  It also supports a custom timeout.

lib/puppet/util/postgresql_validator.rb
```diff
@@ -0,0 +1,64 @@
...
+        final_cmd = []
+
+        cmd_init = "#{@resource[:psql_path]} --tuples-only --quiet "
+
+        final_cmd.push cmd_init
+
+        cmd_parts = {
+          :host => "-h #{@resource[:host]}",
+          :port => "-p #{@resource[:port]}",
+          :db_username => "-U #{@resource[:db_username]}",
```
hasegeli: Some short and some long arguments don't please my eyes.  I prefer `--xxx=yyy` format on scripts when possible to avoid argument parsing issues.

lib/puppet/util/postgresql_validator.rb
```diff
@@ -0,0 +1,64 @@
...
+      end
+
+      private
+
+      def execute_command
+        Execution.execute(build_validate_cmd, :uid => @resource[:run_as])
+      end
+
+      def build_validate_cmd
+        "/bin/echo 'SELECT 1' | #{parse_connect_settings.join(' ')} #{build_psql_cmd} "
```
hasegeli: I wouldn't spawn another process but use `--command='SELECT 1'`.

ruby.rb
```diff
@@ -0,0 +1,58 @@
+Puppet::Type.type(:validate_db_connection).provide(:ruby) do
+
+  commands :psql => 'psql'
+
+  def psql_cmd
+    final_cmd = []
+
+    cmd_init = "#{resource[:psql_path]} --tuples-only --quiet "
```
hasegeli: `--no-psqlrc` is also always a good idea on scripts.  Otherwise the client's configuration can interfere.

PR #888, opened 2017-06-22
use https for apt.postgresql.org repo

hex2a:


eputnam:
thanks for the PR @hex2a !

PR #893, opened 2017-06-30
Allow order parameter to be string value

matonb:
concat::fragment accepts either a string or integer as order parameter value.
Numeric ordering doesn't work once converted to string filenames:

20 sequential fragments for example

| Numeric (no padding) | String (0 Zero Padded) |
|----------------------|------------------------|
| 10_fragment          | 001_fragment           |
| 11_fragment          | 002_fragment           |
| ...                  |                        |
| 19_fragment          | 010_fragment           |
| 1_fragment           | 011_fragment           |
| 20_fragment          | 012_fragment           |
| 2_fragment           | 013_fragment           |
| 3_fragment           | 014_fragment           |

Order has gone to pot...

### DIFF
--- a/manifests/server/pg_hba_rule.pp
+++ b/manifests/server/pg_hba_rule.pp
@@ -5,10 +5,10 @@ define postgresql::server::pg_hba_rule(
   String $database,
   String $user,
   String $auth_method,
-  Optional[String] $address     = undef,
-  String $description           = 'none',
-  Optional[String] $auth_option = undef,
-  Integer $order                = 150,
+  Optional[String] $address       = undef,
+  String $description             = 'none',
+  Optional[String] $auth_option   = undef,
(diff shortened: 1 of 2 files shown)


derbious:
Yes please.
This is also causing an issue creating the `ipv4acls`. Also, the internal function `postgresql_acls_to_resources_hash` returns a zero-padded string for order. https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/lib/puppet/parser/functions/postgresql_acls_to_resources_hash.rb#L48

matonb:
@derbious Hadn't noticed the ordering 'fix' in that file.

I opened this PR because the current release didn't work with my existing configuration.

RFC: Making order an integer deviates from concat::fragment allowing both integer and string?

HelenCampbell:
@matonb Merged, thank you!

PR #896, opened 2017-07-07
add parameter "version" to postgresql::server::extension to update the extension version

georgehansper:
[MODULES-5635](https://tickets.puppetlabs.com/browse/MODULES-5635)
This change adds the parameter "version" to the defined type postgresql::server::extension

If unspecified, the module behaviour is unchanged.

If specified, the module will invoke the PostgreSQL-specific `ALTER EXTENSION...`

`version` may be set to `latest`, in which case the SQL `ALTER EXTENSION "extension" UPDATE` is applied to this database (only).

`version` may be set to a specific version, in which case the extension is updated using `ALTER EXTENSION "extension" UPDATE TO 'version'`

When an extension package is updated, this does not automatically change the effective version in each database. The `ALTER EXTENSION...` takes care of this on a per-database basis.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1177,6 +1177,21 @@ Valid options: 'present' or 'absent'.
 
 Specifies the extension to activate. If left blank, uses the name of the resource.
 
+#### `version`
+
+Specifies the version of the extension which the database uses.
+When an extension package is updated, this does not automatically change the effective version in each database.
+
+This needs be updated using the PostgreSQL-specific SQL `ALTER EXTENSION...`
+
+`version` may be set to `latest`, in which case the SQL `ALTER EXTENSION "extension" UPDATE` is applied to this database (only).
+
+`version` may be set to a specific version, in which case the extension is updated using `ALTER EXTENSION "extension" UPDATE TO 'version'`
+
+eg. If extension is set to `postgis` and version is set to `2.3.3`, this will apply the SQL `ALTER EXTENSION "postgis" UPDATE TO '2.3.3'` to this database only.
+
+`version` may be omitted, in which case no `ALTER EXTENSION...` SQL is applied, and the version will be left unchanged.
+
 ##### `package_name`
 
 Specifies a package to install prior to activating the extension.
--- a/manifests/server/extension.pp
+++ b/manifests/server/extension.pp
@@ -1,11 +1,12 @@
 # Activate an extension on a postgresql database
 define postgresql::server::extension (
   $database,
-  $extension        = $name,
-  String[1] $ensure = 'present',
-  $package_name     = undef,
-  $package_ensure   = undef,
-  $connect_settings = $postgresql::server::default_connect_settings,
+  $extension                   = $name,
+  Optional[String[1]] $version = undef,
+  String[1] $ensure            = 'present',
+  $package_name                = undef,
+  $package_ensure              = undef,
+  $connect_settings            = $postgresql::server::default_connect_settings,
 ) {
   $user             = $postgresql::server::user
   $group            = $postgresql::server::group
@@ -42,7 +43,10 @@ define postgresql::server::extension (
     db               => $database,
     command          => $command,
     unless           => "SELECT t.count FROM (SELECT count(extname) FROM pg_extension WHERE extname = '${extension}') as t WHERE t.count ${unless_comp} 1",
-    require          => Postgresql::Server::Database[$database],
+  }
+
+  if($database != undef and defined(Postgresql::Server::Database[$database])) {
+    Postgresql::Server::Database[$database]->Postgresql_psql["Add ${extension} extension to ${database}"]
   }
 
   if $package_name {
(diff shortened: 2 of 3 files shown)


sammcj:

sammcj:
Hi @puppetlabs, just wondering if this could please be merged in as we're trying not to run off long-existing forks of official modules.

Thanks,

georgehansper:
Issue is [MODULES-5634](https://tickets.puppetlabs.com/browse/MODULES-5634)

georgehansper:
Issue is [MODULES-5635](https://tickets.puppetlabs.com/browse/MODULES-5635)

tphoney:
Fantastic PR many thanks @georgehansper  !!!

georgehansper:
Thank you very much for merging this PR.

antgel:
@georgehansper Hi, I think that this PR is causing the behaviour seen at https://github.com/example42/psick/issues/193, specifically:
```
Debug: Executing with uid=postgres gid=postgres: '/usr/bin/psql -d puppetdb -t -c "SELECT COUNT(*) FROM (SELECT t.count FROM (SELECT count(extname) FROM pg_extension WHERE extname = 'pg_trgm') as t WHERE t.count = 1) count"'
Error: /Stage[main]/Puppetdb::Database::Postgresql/Postgresql::Server::Extension[pg_trgm]/Postgresql_psql[Add pg_trgm extension to puppetdb]: Could not evaluate: Error evaluating 'unless' clause, returned pid 9730 exit 1: ''
```
It seems that it's quite possible for the `psql` command to run before the postgresql server is installed, let alone running. Downgrading from 5.2.0 to 5.1.0, which didn't contain this PR, fixed the problem. Any thoughts?

georgehansper:
It is _possible_ for the psql command to run before postgres is started, but I don't think that the code in PR #896 is the cause of this problem.
For this to happen, it would be necessary to specify:
     `database => undef`
in the type declaration of postgresql::server::extension

This is definitely not the case here, because the psql command which is failing explicitly includes the DB puppetdb.
```
/usr/bin/psql -d puppetdb -t -c "SELECT COUNT(*) FROM (SELECT t.count FROM (SELECT count(extname) FROM pg_extension WHERE extname = 'pg_trgm') as t WHERE t.count = 1) count"
```

The question I would ask is: Does the puppet run succeed because the version was rolled back, or just because it was run a second time (regardless of the version)?

I have tried to reproduce the issue using:
```
include stdlib

  encoding => 'UTF-8',
  locale   => 'en_AU.UTF-8',
  version  => '9.6',
  datadir  => '/var/lib/pgsql/9.6/data',
  user     => 'postgres',
  group    => 'postgres',
} ->
class { 'postgresql::server':
  listen_addresses => '*',
}

postgresql::server::database { 'db_with_pg_trgm':
}

postgresql::server::extension { 'add extension pg_trgm to db_with_pg_trgm':
  extension => 'pg_trgm',
  database  => 'db_with_pg_trgm',
}
```
But puppet always gets the ordering right.
The ordering is derived from the dependancies, which state that the extension requires the nominated database to exist (and be managed under puppet).
For puppet to be able to check that the DB exists, postgresql needs to be running.
ie puppet starts postgresql before the extension is applied, or the unless clause is executed.

I suspect the problem is more likely related to the subsequent error message:
```
  Error: /Stage[main]/Postgresql::Server::Reload/Exec[postgresql_reload]: Failed to call refresh: invalid byte sequence in US-ASCII
```
which might account for postgresql not being started, or failing soon after being started.
Note that applying an extension does not require a reload, so the reload is due to something else.

FYI: Specifying `database=>undef` allows puppet to apply an extension to the DB postgres.

justinstoller:
So, I'm seeing a similar issue when running the PuppetDB module now, but only on hosts that have pulled in the new postgresql module. I think the problem is that the check
```puppet
and defined(Postgresql::Server::Database[$database])
```
is dependent on the manifest order.

For example the puppetdb module defines the extension to be installed and then defines the database to install them into [here](https://github.com/puppetlabs/puppetlabs-puppetdb/blob/master/manifests/database/postgresql.pp#L27-L37).

The PDB ordering is admittedly weird, but I do think that this causes a change in behavior by not complete auto-requiring dependent databases.

antgel:
Thanks @justinstoller for clarifying. @georgehansper Any thoughts?

georgehansper:
I suspect @justinstoller is right, and that using `defined(Postgresql::Server::Database[$database])` may not be the right approach. I will try something else.

georgehansper:
Hello @justinstoller @antgel

I can reproduce the error message using the module puppetlabs-puppetdb

I have created a new branch `extension_upgrade_fixdep` in the repository https://github.com/infoxchange/puppetlabs-postgresql
which I believe will resolve this issue.

Before I create a pull-request, could you please verify that this branch also resolves the issue in your environment?

Regards,
George Hansper

antgel:
@georgehansper Works for me! :)

justinstoller:
Hi! Sorry, I've gotten a bit swamped, the change in your branch looks fine to me, but I don't have time to functionally test it. I've pinged the modules folks to see if they can work with you to get this across the line.

georgehansper:
Hi @justinstoller , @antgel 

I have submitted PR #932 to merge the code from https://github.com/infoxchange/puppetlabs-postgresql branch `extension_upgrade_fixdep`

PR #897, opened 2017-07-07
add ensure=>absent to postgresql::server::role

georgehansper:
[MODULES-5636](https://tickets.puppetlabs.com/browse/MODULES-5636)
This change introduces the parameter 'ensure' to the defined type postgresql::server::role

ensure => present (default) leaves the current behaviour unchanged.

ensure => absent invokes `DROP ROLE...` instead of `CREATE ROLE...`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1444,7 +1444,16 @@ Provides the target for the rule, and is generally an internal only property.
 **Use with caution.**
 
 #### postgresql::server::role
-Creates a role or user in PostgreSQL.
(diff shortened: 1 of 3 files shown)


sammcj:

georgehansper:
Issue is [MODULES-5636](https://tickets.puppetlabs.com/browse/MODULES-5636)

sammcj:
Hey @tphoney, thanks for merging #896 - would you mind taking a look this as well?

PR #900, opened 2017-07-12
Default contcat order

matonb:
Remove numeric order override from concat, use default alpha sort.


tphoney:
@matonb do you foresee any backwards compatibility issues ?

matonb:
I don't think there are any tests for this specific case,  I'll run some 'real-world' cases and get back to you.

tphoney:
many thanks !

tphoney:
Thanks @matonb !

PR #913, opened 2017-08-21
Change - Add support for Fedora 26

blackknight36:
Fedora 26 provides postgresql-server version 9.6 by default.  Added
support to avoid puppet failures on Fedora 26 nodes.

See https://rpmfind.net/linux/RPM/fedora/updates/26/x86_64/p/postgresql-server-9.6.4-1.fc26.x86_64.html for details.


tphoney:
sweet. Thanks @blackknight36

PR #914, opened 2017-08-27
use postgresql 9.6 for the newest SLES and openSUSE releases

tampakrap:


tphoney:
Awesome @tampakrap thanks for the PR !!!

PR #917, opened 2017-09-01
Remotely manage Redshift servers, users, groups, and grants

y-trobinso:
Some months back, we forked puppetlabs-postgresql to provide the ability to remotely manage Redshift servers in AWS. This includes connections to the remote cluster, users, groups, and grants, but _not_ cluster-level changes at this time (which are better managed in modules aware of AWS management details, such as [terraform](https://www.terraform.io/)).

To do this, we extended some of the plumbing made available in MODULES-661 (https://github.com/puppetlabs/puppetlabs-postgresql/pull/636), repurposing several classes to provide options for Redshift's dialect and user model differences with vanilla postgresql. We've also provided for newer keywords made available by Redshift's recent inclusion of GetClusterCredentials (eg, `PASSWORD DISABLE`, as per https://docs.aws.amazon.com/redshift/latest/dg/r_ALTER_USER.html).

That all said, this branch was forked some time ago, and it may lack a requisite level of polish to make it immediately suitable for inclusion in the main postgresql module. The purpose of this PR, then, is to open a dialog -- to determine if this feature is suitable upstream in the puppet module instead of just our fork. If it is indeed suitable, the second goal of this PR is to iron out any wrinkles in ancillary design issues, code, docs, and tests prior to inclusion.

I'll be responsive to discussion here as the primary owner of the Yelp fork. Please feel welcome to open issues and PRs on the fork directly, should that make it easier to serialize changes.


y-trobinso:
We have a PR in internal review (https://github.com/Yelp/puppetlabs-postgresql/pull/18) that should fix the build failures. I'll update once those are cleared.

y-trobinso:
Release checks are now clean and updated downstream.

y-trobinso:
Ping for review. This one is fairly lengthy, so please let us know if we can chop it down or make changes to make this viable.

FransUrbo:
I reported a few issues a few months back that haven't been looked at, but they're also about AWS. RDS, granted, but how "Redshifty" is this PR?

https://tickets.puppetlabs.com/browse/MODULES-5068
https://tickets.puppetlabs.com/browse/MODULES-5069
https://tickets.puppetlabs.com/browse/MODULES-5072

y-trobinso:
@FransUrbo In order to facilitate the differences between postgresql (>=8.4) and Redshift, this review adds a `$dialect` variable (of default `'postgres'`). Anything Redshift-specific should fall behind `$dialect == 'redshift'` so as to not break existing compatibility.

Addressing the specific issues you raise:
* MODULES-5068: pg_shadow exists in Redshift at this writing, but it does not appear to be accessible even to superusers or the master user account for a cluster. For this reason, checking the password hash as part of an `UNLESS` statement is incompatible with the redshift dialect: https://github.com/Yelp/puppetlabs-postgresql/blob/master/manifests/server/role.pp#L211
* MODULES-5069: A conditional has been added [here](https://github.com/Yelp/puppetlabs-postgresql/blob/master/manifests/server.pp#L77) for the redshift dialect to skip local server installation. To connect, we've been passing in `$connection_settings` directly.
* MODULES-5072: To prevent class namespace collisions when managing multiple clusters, this change prefixes `${title}` when calling `[Pp]ostgresql_psql`. For example, [here](https://github.com/Yelp/puppetlabs-postgresql/blob/master/manifests/server/role.pp#L68).

Perhaps the most "Redshifty" change in this review is in how the user model works. Postgresql uses the `ROLE` keyword to identify a collection of permissions granted to users, groups, groups of groups, etc, where roles can apply to each other recursively. By comparison, Redshift splits this into `USER` and `GROUP`, limiting recursion and requiring the support of two distinct keywords. This review extends that support to role.pp, adding two additional classes (dbgroup.pp and dbgroupmember.pp) to handle the differences in the user model. However, we are aware of one outstanding issue regarding grants on group entities in Redshift, which I have a WIP fix in progress for here: https://github.com/Yelp/puppetlabs-postgresql/pull/19/files

To summarize: this change _MUST NOT_ regress existing functionality (ie, first do no harm). Beyond that guarantee, support for Redshift contained in this PR (enabled by passing in `$dialect == 'redshift'`) can be considered beta to late beta, insofar as there are some known issues and rough edges that would benefit from more developer coverage and time. We're PRing in case this is useful to maintainers of the Puppetlabs module, but we're more than happy to have this remain a fork if needed.

FransUrbo:
MODULES-5068 (pg_shadow issue): I see that you're not running the `unless` at all (commented out). My solution was to select from `pg_user` instead. Is that something you can do?

MODULES-5069 + MODULES-5072 (must install local psql server): Nice!! THanx!


So your addition of `skip_install` is basically what I need. My question is really: "Can I use `dialect => 'redshift'` to do this for RDS instead of RedShift? Or do you think there's differences between the two?

y-trobinso:
> MODULES-5068 (pg_shadow issue): I see that you're not running the unless at all (commented out). My solution was to select from pg_user instead. Is that something you can do?

At least when I tested this, the contents of `pg_user` contained literal star characters, intead opting to store the password in pg_shadow. Others' mileage may vary, but since the comparison is made between hashed passwords, I was unable to write a comparable check here.

> So your addition of `skip_install` is basically what I need. My question is really: "Can I use `dialect => 'redshift'` to do this for RDS instead of RedShift? Or do you think there's differences between the two?

The two are different enough (eg, in information_schema and keywords) that I would not rely on the Redshift dialect for RDS. However, the approach taken here might be able to be extended/forked to meet your needs, perhaps by providing an "rds" dialect.

Please feel welcome to fork from our downstream branch [here](https://github.com/Yelp/puppetlabs-postgresql) in the meantime.

FransUrbo:
On 30 Sep 2017, at 01:11, Thomas Robinson <notifications@github.com> wrote:

> At least when I tested this, the contents of pg_user contained literal star characters, intead opting to store the password in pg_shadow. Others' mileage may vary, but since the comparison is made between hashed passwords, I was unable to write a comparable check here.

Yes, I’ve managed to get my RDS working with https://github.com/puppetlabs/puppetlabs-postgresql/pull/922
thanx to your PR giving me some pointers and discovered the same in RDS. Maybe it would have been
better just to leave the unless (in my PR) commented out like you did…

The pg_authid is also not readable by anyone in RDS (as I think you mention it was the same in RedShift), so
can’t use that either for RDS..

Other than that unless (and the pg_user/pg_shadow thing), it seems that RDS/PostgreSQL is exactly the same
as “normal” PSQL, so I wouldn’t need much for RDS. And all that might be a PSQL v9 thingie, not sure..

> The two are different enough (eg, in information_schema and keywords) that I would not rely on the Redshift dialect for RDS. However, the approach taken here might be able to be extended/forked to meet your needs, perhaps by providing an "rds" dialect.

Yes, the only part that needs an “dialect = rds” would be that unless - to have it not run on RDS I mean.

However, I figured out a … “better” (?) way than the “skip_install”. Using the already existing ‘package_ensure’.
Not sure if that could be called with ‘noinstall’ instead, but I used the most obvious ‘purged’…


If you could rebase/rewrite your PR in more logical commits (at least one/the first with only the ‘dialect’ thing,
which then everywhere else just uses the default value of ‘postgresql’ and then the others where you add your
‘redshift’ additions), it would probably be easier for the maintainers of this module to review and I could then
“piggy-back” on your PR with mine…


But I’ll do a PR on your PR in your repo for my minute little change for RDS, but everywhere where your
have “if dialect = postgresql’ needs to be ‘if dialect = postgresql or dialect = rds’, so it would’t be pretty :)

y-trobinso:
Ignore the current build failures. This same traceback is occurring on a clean master pull of Puppetlabs/puppetlabs-postgresql as well, and appears to be dependency-related: https://travis-ci.org/puppetlabs/puppetlabs-postgresql/jobs/282494595

@FransUrbo:
> If you could rebase/rewrite your PR in more logical commits (at least one/the first with only the ‘dialect’ thing, which then everywhere else just uses the default value of ‘postgresql’ and then the others where you add your ‘redshift’ additions), it would probably be easier for the maintainers of this module to review and I could then “piggy-back” on your PR with mine…

I'll consider doing this. Although, it's possible to loosely follow the commit history as improvements were made, from postgres::server, to postgres::role, to postgres::grant, and onward. Splitting by file would not be wrong.

> But I’ll do a PR on your PR in your repo for my minute little change for RDS, but everywhere where your have `if dialect = postgresql` needs to be `if dialect = postgresql or dialect = rds`, so it would’t be pretty :)

Nice PR. :)

And yeah, I'm sure there's a cleaner representation to be found for this. If not `dialect`, then perhaps a new variable pointing out the hosting framework.

eputnam:
@y-trobinso ,

Thanks for the colossal amount of work you put in here! My immediate thought was this: would it make more sense to perhaps have a separate Redshift module? We'd like to keep the scope of this module to just vanilla(-ish) postgresql and maybe Redshift is enough extra stuff that it merits its own module? @hunner @tphoney

y-trobinso:
> Thanks for the colossal amount of work you put in here!

No prob!

> My immediate thought was this: would it make more sense to perhaps have a separate Redshift module? We'd like to keep the scope of this module to just vanilla(-ish) postgresql and maybe Redshift is enough extra stuff that it merits its own module?

That seems reasonable to me, given our fork/this PR alters the contract of `postgresql::server` to allow the management of remote clusters other than vanilla postgres. Several of the information schema options made available to vanilla are also not available (or are located differently in) redshift, which is another good case for a separate module to reduce the amount of logic necessary to maintain each lookup strategy (especially in `grant.pp`).

That said, much of the plumbing, particularly for psql client connectivity, remains the same. So while a separate module would be appropriate, some forking, subclassing, and/or library strategy seems reasonable here to prevent duplication. Which, for now, is served just fine with our downstream (`Yelp/puppetlabs-postgresql`) fork.

I think in the ideal world, any new module will fork from vanilla puppetlabs-postgresql initially, until enough is pulled out and abstracted for it to exist independently (pulling from common dependencies). But I think it's ultimately repository maintainers' call on what that should look like.

Thank you so much for the reply, and do let me know if there's anything more I can provide here.

davinhanlon:
Hi @y-trobinso . As per @eputnam 's comment above, thanks for doing all this work and submitting the PR. We appreciate the submission and the conversation that it sparked. We reviewed this and feel that it would probably be better served in a separate Redshift module. That said, I appreciate your concern about duplicate code and realise it's not ideal either.

In any event, we're going through our Puppet supported modules currently and performing widespread updates on them ("Rubocopping" them). This means that this PR will need to be rebased afterwards. For now I will leave this PR open, but we're likely to request re-basing shortly. If you feel this Redshift functionality is best served in the main Postgresql module then we'd appreciate it if you could re-base after our Rubocop release and we can re-review this - it would be best done as part of the Community PR triage (open session on a Thursday at 0900 US PST: https://etherpad.openstack.org/p/puppet-community-triage).

y-trobinso:
Thanks @davinhanlon . I chatted with interested parties on my end, and we'll keep a lookout for the need to rebase this PR in the near future.

As for moving this into a separate module, I agree that that is likely the cleanest way to do this. To that end, please let us know if it makes better sense to leave this in PR form (for an eventual hard fork) or to close the PR and instead base this module on our downstream (Yelp/puppetlabs-postgresql) instead.

Either way, we're happy to continue to maintain our downstream fork and this PR for the time being and chase any merges or rebases that need to happen from the main puppetlabs-postgresql module. If it's easier to reconcile any additional work that needs to happen (ie, rebasing, module migration) in real time during PR triage, we're also happy to commit to that in the coming weeks.

y-trobinso:
For simplicity's sake, we (on our end) have decided to continue to maintain this PR in its separate fork (Yelp/puppetlabs-postgresql). We'll continue to keep the Redshift pieces there, until such a time as it makes sense to migrate them into their own module.

Thank you for all the help. Closing this PR accordingly.

PR #919, opened 2017-09-21
Updates test to reflect new error message

HelenCampbell:


spec/unit/defines/server/grant_spec.rb
```diff
@@ -233,7 +233,11 @@
       "class {'postgresql::server':}"
     end
 
-    it { is_expected.to compile.and_raise_error(/parameter 'object_name' variant 0 expects size to be 2, got 1/) }
+    if Puppet::Util::Package.versioncmp(Puppet.version, '5.2.0') >= 0
```
tphoney: or 
it { is_expected.to compile.and_raise_error(/parameter 'object_name' variant \d expects size to be 2, got 1/) }

PR #922, opened 2017-09-29
Do not require installation of server packages

FransUrbo:
This PR solves

MODULES-5068: pg_shadow does not exist in RDS @ AWS
MODULES-5069: Can't manage a remote PostgreSQL instance without installing server locally

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -71,14 +71,20 @@ class postgresql::server (
     warning('Passing "createdb_path" to postgresql::server is deprecated, it can be removed safely for the same behaviour')
   }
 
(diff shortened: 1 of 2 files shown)


FransUrbo:
This should probably be "piggy-backed" of off #917 with a `dialect ?= redshift` to avoid doing that `unless` in the second commit (which is failing CI tests).

KedneckInc:
Any chance of getting this working?  I'm in a situation where I'd like to use the pg_hba rule management features, but the database is already installed in a Docker image.

david22swan:
@FransUrbo Is there any movement on this pr?

PR #932, opened 2017-11-07
add parameter "version" to postgresql::server::extension - fix dependency on database

georgehansper:
The previous change to the module postgresql::server::extension introduced a conditional dependency on the database which contains the extension.

This dependency was implemented in a way that did not work properly in all cases.

This update addresses this,  and the dependency on the database should now apply consistently.

### DIFF
--- a/manifests/server/extension.pp
+++ b/manifests/server/extension.pp
@@ -32,6 +32,12 @@ define postgresql::server::extension (
     }
   }
 
+  if( $database != 'postgres' ) {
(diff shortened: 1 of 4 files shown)


eputnam:
Thanks @georgehansper !

antgel:
Indeed thanks @georgehansper ! @eputnam Is a release on the horizon?

eputnam:
@antgel there is a release in progress. you can follow it here: https://tickets.puppetlabs.com/browse/MODULES-5955

PR #942, opened 2017-12-06
According to the puppet doc, Pattern should be a list of regex.

PierreR:
Link to the doc:
https://puppet.com/docs/puppet/5.3/lang_data_abstract.html#pattern


PierreR:
I am using[ a tool ](https://github.com/bartavelle/language-puppet/issues/227) that implements the puppet grammar. This kind of "outside the spec" declaration are really annoying specially in popular, supported puppetlabs module such as `puppetlabs-postgresql`.

I hope you accept this PR.

Thanks,

PR #943, opened 2017-12-11
{WIP} Manage postgresql.conf directory

tphoney:


manifests/server/config.pp
```diff
@@ -106,6 +108,15 @@
     }
   }
 
+ if ($manage_postgresql_conf == true) {
+    # Prepare the main pg_hba file
+    file { $postgresql_conf_path:
+      owner  => $user,
+      group  => $group,
+      mode   => '0600',
+      }
```
tphoney: @hunner looking for a neat way to properly infer dependance. Ie the directory should be set before we attempt to change the postgresql.conf file.

PR #947, opened 2018-01-10
defaulted psql_path to postgresql::server::psql_path

crayfishx:
The `psql_path` variable declared within the  `postgresql::server::service` class defaults by calling the default directly from `params` rather than from the parent service class like other variables.  Since these are not parameters, this makes `psql_path` impossible to override using class params/hiera.   It looks like this was done in error as `$postgresql::server::psql_path` exists and defaults to the params entry.

This fix defaults it to `$postgresql::server::psql_path` and should give the same default behavour but allow you to override `postgresql::server::psql_path: foo` in Hiera.

### DIFF
--- a/manifests/server/service.pp
+++ b/manifests/server/service.pp
@@ -9,7 +9,7 @@ class postgresql::server::service {
   $user             = $postgresql::server::user
   $port             = $postgresql::server::port
   $default_database = $postgresql::server::default_database
-  $psql_path        = $postgresql::params::psql_path
+  $psql_path        = $postgresql::server::psql_path
   $connect_settings = $postgresql::server::default_connect_settings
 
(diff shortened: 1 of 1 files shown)


ubellavance:
I don't know if it's related, but I'm trying to use the module with the postgresql package from SCL and it has weird paths.  For example, psql is at `/opt/rh/rh-postgresql96/root/usr/bin/psql`. If I try to use the module, the connection validator fails, most likely because it cannot find the psql binary. I don't know how to override it.

ubellavance:
Nevermind, it looked it worked OK.  The problem is that when it calls psql from the SCL package, it cannot find a shared object.  I installed the non-SCL psql and removed my psql_path entry and it works ok.

PR #949, opened 2018-01-15
Use puppet4 functions-api

juliantodt:
The legacy puppet3 functions-api should be avoided.
This migrates the existing functions to the new api and uses it in the module. It also adds the tests for the new functions based on those for the olds.
We should keep the legacy functions for now to allow compatibility for those using the functions outside out this module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ class { 'postgresql::server':
 
 postgresql::server::db { 'mydatabasename':
   user     => 'mydatabaseuser',
-  password => postgresql_password('mydatabaseuser', 'mypassword'),
+  password => postgresql::password('mydatabaseuser', 'mypassword'),
 }
 ```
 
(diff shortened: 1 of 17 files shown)


hasegeli:
Copying the functions is not good for maintainability.  Can't we make the old functions call the new ones?

juliantodt:
Yes, I can see that. The idea is to remove the old functions sooner or later while allowing compatibility for now as we cannot be sure that no other modules might use the functions provided here.
What do you think about adding deprecation-warnings like the ones in puppetlabs-stdlib [here](https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/lib/puppet/parser/functions/validate_array.rb#L23) to the old functions and removing them in the next-ish major release?
Simply calling the new functions is not possible afaik and also would not be the best solution inmho because we want people to move to the new functions anyway.

david22swan:
@juliantodt You will need to rebase your changes. In addition I am pretty sure that implementing puppet 4 will make this change backwards incompatible and so you will likely have to wait for it to be put in.

david22swan:
@juliantodt Apologies, but due to the amount of time that this PR has been left to lay fallow I am afraid that I must now close it. If at a later time you wish to return to and complete it, or submit an entirely new pr, we will of course be more than happy to review them.

PR #950, opened 2018-01-18
Allow adding roles, config entires and hba rules via hiera

ekohl:
This includes refactoring like using contain rather than the anchor pattern.

The autorequires should ensure the sql is always executed after the server is configured. This was needed because the hiera entries would create a cyclic dependency. For some reason the test is failing so it might not be working.

### DIFF
--- a/lib/puppet/type/postgresql_psql.rb
+++ b/lib/puppet/type/postgresql_psql.rb
@@ -120,6 +120,8 @@ Puppet::Type.newtype(:postgresql_psql) do
     newvalues(:true, :false)
   end
 
+  autorequire(:class) { ['Postgresql::Server::Service'] }
+
   def should_run_sql(refreshing = false)
     onlyif_param = @parameters[:onlyif]
     unless_param = @parameters[:unless]
--- a/spec/acceptance/overridden_settings_spec.rb
+++ b/spec/acceptance/overridden_settings_spec.rb
@@ -1,0 +1,53 @@
+require 'spec_helper_acceptance'
+
+# These tests are designed to ensure that the module, when ran overrides,
+# sets up everything correctly and allows us to connect to Postgres.
+describe 'postgresql::server', unless: UNSUPPORTED_PLATFORMS.include?(fact('osfamily')) do
+  pp = <<-MANIFEST
+    class { 'postgresql::server':
+      roles          => {
+        'testusername' => {
+          password_hash => postgresql_password('testusername', 'supersecret'),
+          createdb      => true,
+        },
+      },
+      config_entries => {
+        max_connections => 200,
+      },
+      pg_hba_rules   => {
+        'from_remote_host' => {
+          type        => 'host',
+          database    => 'mydb',
+          user        => 'myuser',
+          auth_method => 'md5',
+          address     => '192.0.2.100/32',
+        },
+      },
+    }
+
+    postgresql::server::database { 'testusername':
+      owner => 'testusername',
+    }
+  MANIFEST
+
+  it 'with additional hiera entries' do
+    apply_manifest(pp, catch_failures: true)
+    apply_manifest(pp, catch_changes: true)
+  end
+
+  describe port(5432) do
+    it { is_expected.to be_listening }
+  end
+
+  it 'can connect with psql' do
+    psql('--command="\l" postgres', 'postgres') do |r|
+      expect(r.stdout).to match(%r{List of databases})
+    end
+  end
+
+  it 'can connect with psql as testusername' do
+    shell('PGPASSWORD=supersecret psql -U testusername -h localhost --command="\l"') do |r|
+      expect(r.stdout).to match(%r{List of databases})
+    end
+  end
+end
--- a/spec/unit/classes/server_spec.rb
+++ b/spec/unit/classes/server_spec.rb
@@ -163,4 +163,58 @@ describe 'postgresql::server', type: :class do
       is_expected.to contain_class('postgresql::repo').with_version('99.5')
     end
   end
+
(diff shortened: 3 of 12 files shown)


eputnam:
@ekohl autorequires are not evaluated at compile time so there's no way to test them with unit tests. `that_requires` does not test autorequire.

ekohl:
@eputnam how should we cover that then? Ignore it in the tests? More unit tests?

lib/puppet/type/postgresql_psql.rb
```diff
@@ -119,6 +119,8 @@ def matches(value)
     newvalues(:true, :false)
   end
 
+  autorequire(:class) { 'postgresql::server::service' }
```
ekohl: I don't know if this actually works. I had to add some ```require => Class['postgresql::server::service']``` so it might not.

ekohl:
I believe this should be ready for review. There's an acceptance test that verifies the behavior.

lib/puppet/type/postgresql_psql.rb
ekohl: Turns out it didn't but rspec-puppet was able to test for it. Now updated so it does work.

ekohl:
The build now fails with `Current textdomain ("master_domain") was not added, use FastGettext.add_text_domain !` but I don't know how to actually solve that. Was that something I changed?

eputnam:
@ekohl nothing you did. all better now!

mmoll:
@ekohl this needs a rebase :)

ekohl:
I did rebase it, but somehow I'm now getting an error that it attempts to create the user before the server is started despite there being an autorequire for this that the unit tests prove exists. Maybe the ```require -> undef``` is causing unexpected behavior?

ekohl:
Rebased. I also opened https://github.com/puppetlabs/puppetlabs-postgresql/pull/967 to get some preparation patches merged and reduce the size of this PR.

ehelms:
@ekohl Your cleanup was merged (https://github.com/puppetlabs/puppetlabs-postgresql/pull/967) so I think this can be rebased now?

spec/unit/classes/server_spec.rb
```diff
@@ -163,4 +163,58 @@ class { 'postgresql::globals':
...
+            user: 'myuser',
+            auth_method: 'md5',
+            address: '192.0.2.100',
+          },
+        },
+      }
+    end
+
+    it { is_expected.to compile.with_all_deps }
+    it do # rubocop:disable RSpec/ExampleLength
```
hunner: Rubocop complains about this.

spec/unit/defines/server/reassign_owned_by_spec.rb
```diff
@@ -32,12 +32,13 @@ class {'postgresql::server':}
...
 
+  it { is_expected.to compile.with_all_deps }
   it { is_expected.to contain_postgresql__server__reassign_owned_by('test') }
 
   it {
     is_expected.to contain_postgresql_psql('reassign_owned_by:test:REASSIGN OWNED BY "test_old_role" TO "test_new_role"')
       .with_command('REASSIGN OWNED BY "test_old_role" TO "test_new_role"')
       .with_onlyif(%r{SELECT tablename FROM pg_catalog.pg_tables WHERE\s*schemaname NOT IN \('pg_catalog', 'information_schema'\) AND\s*tableowner = 'test_old_role'.*}m)
-      .that_requires('Class[postgresql::server]')
+      .that_requires('Class[Postgresql::Server::Service]')
```
hunner: Ooo, dependency refs are case insensitive. TIL

spec/unit/classes/server_spec.rb
ekohl: Updated

ekohl:
Thanks! This will need some monitoring because even though the tests pass, I have seen some cases where the actual dependency isn't enforced. I'm starting to suspect there could be a bug in puppet types.

spec/acceptance/overridden_settings_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+  pp = <<-MANIFEST
+    class { 'postgresql::server':
+      roles          => {
+        'testusername' => {
+          password_hash => postgresql_password('testusername', 'supersecret'),
+          createdb      => true,
+        },
+      },
+      config_entries => {
+        max_connections => 200,
```
hunner: This seems to be causing issues on some of our test nodes:
```
09:04:41   Apr 03 09:04:39 k8ytf5mtkdzr708 pg_ctl[14776]: FATAL:  could not create shared memory segment: Invalid argument
09:04:41   Apr 03 09:04:39 k8ytf5mtkdzr708 pg_ctl[14776]: DETAIL:  Failed system call was shmget(key=5432001, size=36954112, 03600).
09:04:41   Apr 03 09:04:39 k8ytf5mtkdzr708 pg_ctl[14776]: HINT:  This error usually means that PostgreSQL's request for a shared memory segment exceeded your kernel's SHMMAX parameter.  You can either reduce the request size or reconfigure the kernel with larger SHMMAX.  To reduce the request size (currently 36954112 bytes), reduce PostgreSQL's shared memory usage, perhaps by reducing shared_buffers or max_connections.
09:04:41   Apr 03 09:04:39 k8ytf5mtkdzr708 pg_ctl[14776]: If the request size is already small, it's possible that it is less than your kernel's SHMMIN parameter, in which case raising the request size or reconfiguring SHMMIN is called for.
```

ekohl: Should I lower it to 20?

PR #953, opened 2018-02-13
Rubocop Process Complete

david22swan:


spec/acceptance/alternative_port_spec.rb
```diff
@@ -2,27 +2,23 @@
...
+    MANIFEST
 
-    apply_manifest(pp, :catch_failures => true)
-    apply_manifest(pp, :catch_changes => true)
+    apply_manifest(pp, catch_failures: true)
+    apply_manifest(pp, catch_changes: true)
   end
 
-  describe port(55433) do
+  describe port(55_433) do
```
pmcmaw: Is this port number correct?

david22swan: Rubocop states that Numeric Literals should be separated into lengths of no more than 3 digit's each, with underscores in the place of decimals.

pmcmaw: I would suggest changing it back and disabling the rule for this line.

spec/acceptance/postgresql_psql_spec.rb
```diff
@@ -1,153 +1,143 @@
...
-      EOS
-
-      apply_manifest(pp, :catch_failures => true)
-      apply_manifest(pp, :expect_changes => true)
-    end
-  end
-
-  it 'should not run some SQL when the onlyif query returns no rows' do
-    pp = <<-EOS
+    pp_seven = <<-MANIFEST.unindent
```
pmcmaw: Do we need to use .unindent?
According to https://zaiste.net/posts/heredoc_in_ruby/  it will 'remove blank characters up to the column where heredoc starts'. I cannot see any blank characters.

david22swan: From what I understand it is to unindent the heredoc to the point were at least one line of it has no indents.

david22swan: That or the first line.

david22swan: This is due to certain heredocs failing when applied if there are excess indents

PR #957, opened 2018-02-22
WIP - (MODULES-6330) - PDK convert 1.3.2

pmcmaw:
Using PDK to convert module so that it is compliant with the PDK.

The second commit `additional workaround` has been done manually, this is addressed in the next version of the pdk template.


pmcmaw:
Out of date. 
Closing.

PR #963, opened 2018-03-01
Resolve dependencies to the databases automatically

hasegeli:


david22swan:
@hasegeli Your changes are currently causing failures in the grant_spec acceptance test's on lines 98 and 110. Please fix this before it can be merged.

hasegeli:
Could you please retrigger the tests?

david22swan:
@hasegeli I've closed and opened the PR. The test's should show fine now.

manifests/server/role.pp
```diff
@@ -4,7 +4,6 @@
   $password_hash    = false,
   $createdb         = false,
   $createrole       = false,
-  $db               = $postgresql::server::default_database,
```
hunner: This parameter is considered public API and removing it would be breaking, even if it doesn't actually do anything. If the db param is not needed then we could just make it unused and deprecated it, and remove it later.

Does the database specified not affect the `ALTER ROLE` commands below? Are roles always defined globally and not db-specific?

ekohl:
This needs a rebase now.

hasegeli:
Have you seen the comment
```
+  # XXX: We should probably do everything in this class before.  Postgresql_psql resource could
+  # specify this better, but currently it is impossible to use "autorequire" against custom
+  # resources.  We can improve this when a better way is available.
```

Do you know how to implement it inside `Postgresql_psql` rather than `postgresql::server::database`.  I cannot read Ruby code.  I might be missing something.

ekohl:
> Do you know how to implement it inside Postgresql_psql rather than postgresql::server::database. I cannot read Ruby code. I might be missing something.

I think I implemented this in https://github.com/puppetlabs/puppetlabs-postgresql/commit/d69ff1ec91d02337415911fac35d8a1fa5a906fa but please verify this.

hasegeli:
Yes, we need the same but against a resource not a class.  Do you know how to do it?  This is how its done on Puppet:

manifests/server/role.pp
hunner: Interesting that it was not documented. Yes, I think it should still be kept and just emit a warning and do nothing. It has been there ever since `postgresql::server::role` was created in 2013 and even before that for `postgresql::role`

hunner:
> `Postgresql_psql["CREATE DATABASE \"${dbname}\""] -> Postgresql_psql <| db == $dbname |>`

I think this would be in autorequire like:
```ruby
autorequire(:postgresql_psql) do
  ["CREATE DATABASE #{self[:db]}"]
end
```

hasegeli:
This would be the same as the current coding.  Although it would be better if we could rely on the whole `postgresql::server::database` resource.  Can we make something like the following work:

```
autorequire(:postgresql::server::db) do
  [#{self[:db]}]
end
```

mattman578:
Hello, we have been using your submodle for a while and we are currently having an issue 
when I run puppet parser validate on server.pp I get the following error message 
Error: Could not parse for environment production: Syntax error at 'Hash'; expected ')' at /home/mayres/git/puppetlabs-postgresql/manifests/server.pp:59
Can you please point me in the correct direction to fix ?

david22swan:
@hasegeli I am know closing this as you had previously mentioned in another PR that you had finished working with Puppet. I would like to once again state how sorry I am to hear that, and hope to one day hear from you again in the future.

PR #965, opened 2018-03-02
(MODULES-6709) - Update to release branch to remove docs changes

pmcmaw:
Cherry picking the docs revert from master and putting into release, then updating the CHANGELOG.


pmcmaw:
No longer releasing this right now. Release process will be restarted when necessary.

PR #966, opened 2018-03-06
(maint) Fix to grant_spec.rb regarding database test on earlier machines

david22swan:
The database test has issues on machines with postgresql version's prior to '9.1.24'.
This is due to the fact that the functions implement by 'https://github.com/puppetlabs/puppetlabs-postgresql/pull/891', i.e. 'ensure => absent uses REVOKE instead of GRANT' was not implemented in postgresql databases until this version. No issue has arisen with the use of revoke on separate tables however so the test's involving them have been left alone.
This PR introduces an exemption on the specified test that prevent's it from being run against machines that do not support it and a warning has been included within the readme file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1154,7 +1154,7 @@ Manages grant-based access privileges for users, wrapping the `postgresql::serve
 
 ##### `ensure`
 
-Specifies whether to grant or revoke the privilege. Default is to grant the privilege.
+Specifies whether to grant or revoke the privilege. Revoke or 'absent' works only in PostgreSQL version 9.1.24 or later.
 
 Valid values: 'present', 'absent'.
 * 'present' to grant the privilege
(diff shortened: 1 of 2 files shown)


pmcmaw:
Can you update the title to a more meaningful title and description and I will get this merged for you?
The current one would make it very difficult to write a changelog entry.

README.md
```diff
@@ -1154,7 +1154,7 @@ Manages grant-based access privileges for users, wrapping the `postgresql::serve
 
 ##### `ensure`
 
-Specifies whether to grant or revoke the privilege. Default is to grant the privilege.
```
pmcmaw: This may require more information that stating it will create problems. Jean will give you some help writing this section or ensure to get it reviewed by Jean.

pmcmaw: Great update. We will get it reviewed by Jean to ensure its ok. I would add that revoke works on tables, the limit is only on revoke permissions on the database or something along them lines.

pmcmaw:
Hey @jbondpdx,

Jus wondering if you could provide some guidance on the README update for this change. 
Long story short, Revoke permissions work on tables but it was only implemented to the database in a later version. The docs associated can be found [here](https://www.postgresql.org/docs/9.1/static/sql-revoke.html).

david22swan:
@jbondpdx Is this fine???

david22swan:
@jbondpdx Forgot about that, fixed :)

pmcmaw:
Nice one David.

PR #967, opened 2018-03-07
Various cleanups

ekohl:
While writing https://github.com/puppetlabs/puppetlabs-postgresql/pull/950 I needed verious cleanups. Because that PR is hard to finish, it's best to get these (safe) cleanups in so there's less changes in mid air. This reduces the change of merge conflicts.

See individual commits for the exact changes.


spec/unit/defines/server/role_spec.rb
```diff
@@ -59,12 +59,13 @@
     end
 
     it { is_expected.to contain_postgresql__server__role('test') }
-    it 'has create role for "test" user with password as ****' do
+    it 'has create role for "test" user with password as ****' do # rubocop:disable RSpec/ExampleLength
```
smortex: The [regression test suite failure](https://travis-ci.org/puppetlabs/puppetlabs-postgresql/jobs/350248274) seems related to this change

PR #980, opened 2018-04-26
(MODULES-7023)   removing duplication in .sync.yml

david22swan:


pmcmaw:
Title needs updated to include the actual ticket eg no space - instead and capitals.
Also failing Travis.

david22swan:
Name changed and is now passing on travis

PR #834, opened 2017-01-12
Fix creation of recovery.conf file when recovery configuration is not specified

Envek:
Fixes https://tickets.puppetlabs.com/browse/MODULES-4276

See also https://github.com/puppetlabs/puppetlabs-postgresql/pull/603#issuecomment-272190972

PR #995, opened 2018-06-07
Fix creation of recovery.conf file when recovery configuration is not specified

cdloh:
Fixes https://tickets.puppetlabs.com/browse/MODULES-4276

See also #603 (comment)

Rebase of orginial MR #834

### DIFF
--- a/README.md
+++ b/README.md
@@ -1501,6 +1501,8 @@ Every parameter value is a string set in the template except `recovery_target_in
 
 A detailed description of all listed parameters can be found in the [PostgreSQL documentation](http://www.postgresql.org/docs/current/static/recovery-config.html).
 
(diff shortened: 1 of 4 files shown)


david22swan:
Running on Adhoc: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/modules/view/linux/view/_adhoc/view/postgresql/

david22swan:
Passed on Adhoc

cdloh:
@david22swan the doco already IMO suggests this behavior. Is there somewhere else that I need to update? See below snippet from the repo readme

> Only the specified parameters are recognized in the template. The recovery.conf is only be created if at least one parameter is set and manage_recovery_conf is set to true.

PR #1001, opened 2018-07-03
(MODULES-6542) - Adding SLES 11 & 12 to metadata

pmcmaw:
Currently we are already testing on these platforms and they are compatible, therefore adding as supported OS following product decision. @davinhanlon

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -62,6 +62,13 @@
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


tuxmea:
@pmcmaw : Why SLES 11 SP1 only? I would use "11" so we cover all releases: SP1 to 4.

PR #1021, opened 2018-09-06
Ignore .psqlrc so output is clean and doesn't break Puppet

flaviogurgel:
I had to abandon the former PR because I didn't fork the project before and my local branch, used for the PR, was lost. Sorry for that.
Doing this properly, I'll be able to rebase if needed.


flaviogurgel:
Anything else to do before this gets merged?

PR #1027, opened 2018-09-24
Parameters `databases` to `postgresql::server` for hiera

mrmcshane:
Building on https://github.com/puppetlabs/puppetlabs-postgresql/pull/950

This adds a hash class parameter for simple database/user creation.

An example hiera config would be:
```
name:
  dbname: name_db
  user: name_user
  password: name_pass
  grant: ALL
```


mrmcshane:
No sorry, this can be removed.

david22swan:
@mrmcshane That's too bad, thanks for putting in the work you did thought. Hope to see more from you in the future.

PR #1031, opened 2018-09-28
Fix provider crashing when some options are set in .psqlrc

XANi:
In some cases, like setting `\x auto` or trying to `\echo` anything in .psqlrc will make provider crash:

    Warning: /Stage[main]/Main/Node[some-server]/Postgresql::Server::Role[backup]/Postgresql_psql[ALTER ROLE "backup" NOCREATEROLE]: Skipping because of failed dependencies
    Error: Error executing SQL; psql returned pid 16230 exit 1: 'Expanded display is used automatically.
    server: some-server.example.com
    ERROR:  role "somerole" already exists

passing option to ignore .psqlrc fixes this, and any other changes that might be set up from this file.

We used puppet to change prompt (red on production servers, green on devs) and add hostname to `psql` to eliminate mistakes and it turned out that this module depends on CLI behaviour of psql.

Removing loading the .psqlrc file should make that behaviour more consistent.

### DIFF
--- a/lib/puppet/provider/postgresql_psql/ruby.rb
+++ b/lib/puppet/provider/postgresql_psql/ruby.rb
@@ -11,6 +11,7 @@ Puppet::Type.type(:postgresql_psql).provide(:ruby) do
     end
 
     command = [resource[:psql_path]]
+    command.push('--no-psqlrc')
     command.push('-d', resource[:db]) if resource[:db]
     command.push('-p', resource[:port]) if resource[:port]
     command.push('-t', '-c', '"' + sql.gsub('"', '\"') + '"')
--- a/spec/unit/puppet/provider/postgresql_psql/ruby_spec.rb
+++ b/spec/unit/puppet/provider/postgresql_psql/ruby_spec.rb
@@ -12,7 +12,7 @@ describe Puppet::Type.type(:postgresql_psql).provider(:ruby) do
       let(:attributes) { { db: 'spec_db' } }
 
       it 'executes with the given psql_path on the given DB' do
-        expect(provider).to receive(:run_command).with(['psql', '-d',
+        expect(provider).to receive(:run_command).with(['psql', '-d', '--no-psqlrc',
                                                         attributes[:db], '-t', '-c', '"SELECT \'something\' as \"Custom column\""'], 'postgres',
                                                        'postgres', {})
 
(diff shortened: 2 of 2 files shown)


david22swan:
@XANi Your pr is currently failing against both the spec test and the syntax check.
The spec test's it is failing are:
```
Failed examples:
rspec ./spec/unit/puppet/provider/postgresql_psql/ruby_spec.rb:14 # Puppet::Type::Postgresql_psql::ProviderRuby#run_sql_command with default attributes executes with the given psql_path on the given DB
rspec ./spec/unit/puppet/provider/postgresql_psql/ruby_spec.rb:33 # Puppet::Type::Postgresql_psql::ProviderRuby#run_sql_command with psql_path and db executes with the given psql_path on the given DB
rspec ./spec/unit/puppet/provider/postgresql_psql/ruby_spec.rb:49 # Puppet::Type::Postgresql_psql::ProviderRuby#run_sql_command with search_path string executes with the given search_path
rspec ./spec/unit/puppet/provider/postgresql_psql/ruby_spec.rb:64 # Puppet::Type::Postgresql_psql::ProviderRuby#run_sql_command with search_path array executes with the given search_path
rspec ./spec/unit/puppet/provider/postgresql_psql/ruby_spec.rb:77 # Puppet::Type::Postgresql_psql::ProviderRuby with port string executes with the given port
rspec ./spec/unit/puppet/provider/postgresql_psql/ruby_spec.rb:89 # Puppet::Type::Postgresql_psql::ProviderRuby with connect_settings executes with the given host
```
While the syntax errors are:
```
Offenses:
lib/puppet/provider/postgresql_psql/ruby.rb:14:18: C: Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols. (https://github.com/bbatsov/ruby-style-guide#consistent-string-literals)
    command.push("--no-psqlrc")
                 ^^^^^^^^^^^^^
```

XANi:
@david22swan fixed. But I have question about those tests:

* why in some cases just `psql` is used and in other it is `attributes[:psql_path]` ? Seems like it passes those tests just fine even tho it should fail if psql is not in default path
* what's the point of testing if "just" commandline looks the same. Or testing for order of it ? Shouldn't it just test if connection actually succeeded ?

XANi:
@david22swan never mind, it was fixed in 58e8f051953edd15b52e539f4d12fec2066382d5

PR #1040, opened 2018-12-06
(MODULES-3804) Fix sort order of pg_hba_rule entries

olavmrk:
The `$order`-parameter to `pg_hba_rule` can be either an integer or a
string, but the corresponding `concat`-resource is configured to sort
the entries using alphanumeric sort.

The result is that a `pg_hba_rule` entry with `order => 2` sorts after
an entry with `order => 10`.

This is also a problem with the default `pg_hba_rule` entries, which
uses numeric order values 1, 2, 3, 4, 100 and 101. These are sorted in
the incorrect order (1, 100, 101, 2, 3, 4).

Unfortunately, since string values are allowed in the
`$order`-parameter, we cannot simply change the `concat`-resource to
use numeric sort.

This patch fixes this issue by zero-padding all integer values to
three digits. E.g. 1 becomes `001`, and 10 becomes `010`. This should
work well for integer values from 0 through 999, and works well with
earlier versions of the module which used string values in that range
for the `$order`-parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1400,6 +1400,12 @@ Specifies a way to uniquely identify this resource, but functionally does nothin
 
 Sets an order for placing the rule in `pg_hba.conf`.
 
+This can be either a string or an integer.
+If it is an integer, it will be converted to a string by zero-padding it to three digits.
+E.g. `42` will be zero-padded to the string `'042'`.
+
+The `pg_hba_rule` fragments are sorted using the `alpha` sorting [order](https://forge.puppet.com/puppetlabs/concat/reference#order).
+
 Default value: 150.
 
 #### `postgresql_version`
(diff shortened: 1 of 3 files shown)


david22swan:
@olavmrk This looks like a good change that I would be happy to merge, but first could you update the documentation to mention the zero padding please?

olavmrk:
@david22swan: I have now updated the documentation for the `order` parameter. I now reads:

> ##### `order`
> 
> Sets an order for placing the rule in `pg_hba.conf`.
> 
> This can be either a string or an integer. If it is an integer, it will be converted to a string by zero-padding it to three digits. E.g. `42` will be zero-padded to the string `'042'`.
> 
> The `pg_hba_rule` fragments are sorted using the `alpha` sorting [order](https://forge.puppet.com/puppetlabs/concat/reference#order).
> 
> Default value: 150.

david22swan:
@olavmrk Look's good to me. Thanks for putting in the work. :)

PR #1043, opened 2018-12-20
(MODULES-8352) Don't use empty encoding string on initdb

binford2k:
pgsql 11 no longer allows initdb with an empty encoding string. Only
include the string when $encoding is set.


eimlav:
Thanks @binford2k for supplying a fix. Looking forward to seeing more of the Rangefinder too!

PR #1044, opened 2018-12-20
(maint) remove inconsistent extra variable

binford2k:
This just removes a variable that was confusing because of how
inconsistently it was used, and didn't really appear to be adding much
value anyway.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -63,8 +63,6 @@ class postgresql::server (
   #Deprecated
   $version                    = undef,
 ) inherits postgresql::params {
-  $pg = 'postgresql::server'
-
   if $version != undef {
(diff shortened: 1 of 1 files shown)


david22swan:
@binford2k Do you have a specific reason for wanting this?? I just worry that it is being used external users and as such would break their code.

binford2k:
Since it's a static string and the variable name is longer than the value it contains, I think the probability that it's being used externally is astronomically small. It's also an internal variable, not part of any public API. That said, if you were highly concerned about that, you could grep each repo listed in the rangefinder report for  `${?(::)?postgresql::server::pg}?`

In reality, I just wanted a trivial & low risk change to a class used by a lot of people for rangefinder testing. ;-)

david22swan:
@binford2k If there wasn't a real reason behind the change I think I'll leave it as is just in case.

PR #1047, opened 2019-01-03
(MODULES-8352) - Fix undef encoding breaking PostgreSQL 11

eimlav:
On PostgreSQL 11, the default value for the encoding parameter is undef. When initdb uses this, it throws an error:

```
Notice: /Stage[main]/Postgresql::Server::Initdb/Exec[postgresql_initdb]/returns: initdb: "" is not a valid server encoding name
Error: '/usr/pgsql-11/bin/initdb --encoding '' --pgdata '/var/lib/pgsql/11/data'' returned 1 instead of one of [0]
```

The changes fix this by checking for an encoding value and omitting the argument if it is undef.


eimlav:
Closing as #1043 already exists which rectifies this issue.

PR #1055, opened 2019-02-05
(MODULES-8553) Further cleanup for package tag issues

HelenCampbell:
Update all this modules tagged packages to use 'puppetlabs-postgresql' instead of just 'postgresql' which is ambiguous and causing dependancy issues.


HelenCampbell:
Thanks for the review Charlie, that's it up to date with those two manifests as well :) Let me know if this is ready to go and I'll get it merged and a release of the back of it.

PR #1062, opened 2019-02-22
(FM-7789) - Clean up of PostgreSQL Test's

david22swan:
Test's run on Ubuntu 14.04
Initial Run: 16 minutes 46 seconds (files took 4 minutes 15.5 seconds to load)
First Pass: 12 minutes 23 seconds (files took 4 minutes 5.9 seconds to load)
ServerSpec implemented: Finished in 11 minutes 52 seconds (files took 4 minutes 14.7 seconds to load)


spec/acceptance/server/recovery_spec.rb
```diff
@@ -59,26 +48,7 @@ class { 'postgresql::globals':
...
-      apply_manifest(pp, catch_failures: true)
-      apply_manifest(pp, catch_changes: true)
-    end
-
-    describe file('/tmp/recovery.conf') do
-      it { is_expected.not_to be_file }
-    end
-  end
-
-  describe 'should not manage recovery' do
```
tphoney: is this tested elsewhere ?

spec/acceptance/server/grant_role_spec.rb
```diff
@@ -65,60 +65,6 @@ class { 'postgresql::server': }
     MANIFEST
   end
   let(:pp_two) do
-    <<-MANIFEST.unindent
```
tphoney: is this no longer used ?

spec/acceptance/server/grant_spec.rb
```diff
@@ -434,32 +343,6 @@ class { 'postgresql::server': }
         end
       end
 
-      it 'grant insert on single table test' do
```
tphoney: is this test checked elsewhere ?

spec/acceptance/server/grant_role_spec.rb
david22swan: It's identical to the first manifest. So I decided to just use it and run the check's of the same apply.

david22swan: Used a comparison checker to make sure.

spec/acceptance/server/recovery_spec.rb
david22swan: Was fifty/fifty on removing it. Did not understand what it was actually testing as the only check was if the file existed, and given as it wasn't mentioned in the manifest it never would, something covered in the above test.

david22swan: There also is a unit test that test's setting that value to false.

spec/acceptance/server/grant_spec.rb
david22swan: I merged it into the above test to grant select on a single table test.

david22swan:
screen shot 2019-02-25 at 12 05 19 pm

PR #1073, opened 2019-05-02
Support current version of puppetlabs/apt.

pillarsdotnet:
Modifies [metadata](https://github.com/puppetlabs/puppetlabs-postgresql/tree/master/metadata.json) to fix dependency errors when used with the current version of [puppetlabs-apt](https://github.com/puppetlabs/puppetlabs-apt/blob/master/metadata.json#L3).

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -14,7 +14,7 @@
     },
     {
       "name": "puppetlabs/apt",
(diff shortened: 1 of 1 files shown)


tphoney:
Many thanks @pillarsdotnet  !

PR #1081, opened 2019-05-17
(FM-7500) conversion to use litmus

tphoney:


metadata.json
```diff
@@ -84,6 +84,6 @@
     }
   ],
   "pdk-version": "1.10.0",
-  "template-url": "https://github.com/puppetlabs/pdk-templates#master",
-  "template-ref": "heads/master-0-gf778803"
+  "template-url": "https://github.com/puppetlabs/pdk-templates/#1.10.0",
+  "template-ref": "1.10.0-0-gbba9ac3"
```
eimlav: The pdk update has reverted the template to the template bundled with pdk 1.10.0.

tphoney:
@eimlav pdk convert re-ran.

.travis.yml
```diff
@@ -11,49 +12,150 @@ script:
...
 rvm:
-  - 2.5.3
-stages:
-  - static
-  - spec
-  - acceptance
-  -
-    if: tag =~ ^v\d
-    name: deploy
+  - 2.5.1
```
pmcmaw: The change in rvm version means we are downgrading the version of ruby we are testing. 
It seems from the old config:
Puppet 5 = Ruby 2.4.5
Puppet 6 = Ruby 2.5.3

PR #1083, opened 2019-05-21
(FM-8031) Add RedHat 8 support

eimlav:


eimlav:
Experimental adhoc with RedHat 8 ran successfully:

Screen Shot 2019-05-21 at 11 37 29

PR #1089, opened 2019-07-03
Fix/directory defined twice

arcenik:
Adds parameters to allow **datadir/logdir/xlogdir** to be managed by another class and avoid a **duplicate declaration**.

Example use case: **$datadir** is a mounted filesystem managed by Puppet, so **file { $datadir: ... }** has to be defined outside of **PostgreSQL** scope.

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -1,11 +1,11 @@
 # @summary Class for setting cross-class global overrides.
 #
 # @note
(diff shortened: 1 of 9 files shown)


sheenaajay:
Thanks alot for submitting the PR @arcenik . Could you please add tests to verify the changes.

arcenik:
Only one Travis job failed due to «Resource temporarily unavailable»

PR #1091, opened 2019-07-09
Enable suport for JIT in PostgreSQL >= 11

biertie:
Dear, 

With this merge request JIT can be enabled on PostgreSQL 11.   
As per example from other packages, it is possible to enable JIT and to change the default package name. 

The change has been tested on RHEL only.

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -22,6 +22,7 @@ class postgresql::server::config {
   $service_name               = $postgresql::server::service_name
   $log_line_prefix            = $postgresql::server::log_line_prefix
   $timezone                   = $postgresql::server::timezone
+  $jit_enable                 = $postgresql::server::jit_enable
 
   if ($manage_pg_hba_conf == true) {
     # Prepare the main pg_hba file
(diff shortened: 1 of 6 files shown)


manifests/server/config.pp
```diff
@@ -130,6 +131,15 @@
     }
   }
 
+  if $jit_enable {
```
npwalker: What happens if you set jit_enable = true and then set it to false afterwards?

biertie: Good point. I will rework it so it becomes a bool.

lionce:
Hello @biertie ,

This seems to be a great feature. Did you have time to update your PR accordingly with @npwalker's suggestion? Would you please add unit tests for your additions, to prevent the future issues. Also please update README file explaining this new feature. Please let us know if you need any help on doing this. 
Thank you!

Cheers!

biertie:
@lionce I am planning to work on it further somewhere next month. At the moment I'm to busy with other issues :-)

florindragos:
Hi @biertie 
Is this still work in progress? There are some problems on non-debian systems: `Evaluation Error: Cannot reassign variable '$pgsql_llvmjit_package_name'` :)

sheenaajay:
@biertie PR Looks good. Could you please add unit tests to verify the change and also suggestions from @npwalker.Ran the Adhoc test suites and it's all green.

michaeltlombardi:
Since this hasn't seen any activity in about four months, I'm going to go ahead and close this pull request for the time being. Please re-open this pull request once the next actions are addressed, new information is available, or you have a question related to this pull request. We've become aware of difficulties re-opening pull requests, in the event you cannot please mention `puppetlabs/modules`  with an @ sign in front and we'll re-open this pull request.

Closing the pull request doesn't mean we don't consider this change valuable, just that there are things that need to be addressed before it can be merged. If you have any questions or concerns, please don't hesitate to ping us in slack.puppet.com.

PR #1092, opened 2019-07-10
(MAINT) Add a default function

npwalker:
Prior to this commit, you could not easily use some of the defined
types without also declaring other classes in your manifest like
postgresql::server.  This is because we were grabbing default
values from postgresql::server.

After this commit, we introduce a pupept function that can return
default values based on the calculations in params/globals.  This
allows users to use defined types that use the default function
without declaring other classes.

### DIFF
--- a/functions/default.pp
+++ b/functions/default.pp
@@ -5,6 +5,6 @@ function postgresql::default(
 
   #search for the variable name in params first
   #then fall back to globals if not found
-  getvar("postgresql::params::${parameter_name}",
+  pick( getvar("postgresql::params::${parameter_name}"),
     "postgresql::globals::${parameter_name}")
 }
--- a/functions/default.pp
+++ b/functions/default.pp
@@ -1,3 +1,14 @@
+# @summary This function pull default values
+# from the params or globals classes as
+# appropriate
+#
+# This function pull default values
+# from the params or globals classes as
+# appropriate
+#
+# @example
+#   postgresql::default('variable')
+
 function postgresql::default(
   String $parameter_name
 ){
--- a/functions/default.pp
+++ b/functions/default.pp
@@ -1,14 +1,8 @@
-# @summary This function pull default values
-# from the params or globals classes as
-# appropriate
-#
-# This function pull default values
-# from the params or globals classes as
-# appropriate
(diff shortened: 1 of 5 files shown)


lionce:
Hello @npwalker ,

Thank you for this contribution. As you can see [the tests are failing](https://travis-ci.org/puppetlabs/puppetlabs-postgresql/jobs/556954867). Would it be possible to fix them? Please let us know if you need any help. 

Cheers!

functions/default.pp
```diff
@@ -0,0 +1,10 @@
+function postgresql::default(
+  String $parameter_name
+){
```
tphoney: Can we improve the wording of what this function does,  for yarrdoc/puppet strings. Otherwise 👍  
```
 ## Functions

+### postgresql::default
+
+Type: Puppet Language
+
+The postgresql::default function.
+
+#### `postgresql::default(String $parameter_name)`
+
+The postgresql::default function.
+
+Returns: `Any`
+
+##### `parameter_name`
+
+Data type: `String`
+
+
+
```

tphoney: To generate the docs
``` bundle exec puppet strings generate --format markdown --out REFERENCE.md ```

Comment (npwalker) [reply functions/default.pp]: @tphoney I can't find an example of how to add docs to a puppet function.  I checked a few supported modules and I don't see any puppet functions... also PDK doesn't have a `pdk new function` which I thought might provide me a good template.  

I took the comment section from the top of a new defined type and put it up there but it doesn't seem to get read when generating the docs.  Can you help me out?

Comment (tphoney) [reply functions/default.pp]: https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/lib/puppet/functions/type_of.rb is a good example.  following from https://puppet.com/docs/puppet/6.7/puppet_strings_style.html

**Comment by tphoney** [comment]
@npwalker if you are still having issues with docs we can take over. adding in @carabasdaniel

File: functions/default.pp (hunk 1 above)
Comment (eimlav) [reply functions/default.pp]: @npwalker You are just missing a bash on the last previous to the `function` declaration. Strings style guide recommends  max 140 chars per line so something like the following should do the trick:
```
# @summary 
#   This function pull default values from the params or globals classes as appropriate
#
# @example
#   postgresql::default('variable')
#
function postgresql::default(
```

PR #1093, opened 2019-07-12
(MODULES-9219) - puppetlabs-postgresql : catalog compilation fails when the service command is not installed

blackknight36:
The service command is deprecated and is not installed by default on
newer Redhat releases.  Updated the params.pp file to define the
service_reload parameter based on what service provider is used on the node.

This fixes errors such as the following.

```
Error: /Stage[main]/Postgresql::Server::Reload/Exec[postgresql_reload]: 
Failed to call refresh: Could not find command 'service'
```

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -78,10 +78,33 @@ class postgresql::params inherits postgresql::globals {
         }
         $confdir                = pick($confdir, $datadir)
       }
+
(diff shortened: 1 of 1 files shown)


blackknight36:
Tests appear to be failing due to an issue uncovered in #1099.

PR #1095, opened 2019-07-17
Fix sql statement for the grants condition

TheBigLee:
The sql statement that determines if the `GRANT` should be applied
or not for the case where the `object_type` is "ALL TABLES IN SCHEMA"
is missing a crucial part in the WHERE clause.

The `SELECT` sub-statement should have a condition that links the bove
SELECT statements together so that a correlation can be made and the
comparison actually returns correct values.

Without the correlation, the SQL statement will always return true, if
a single table in the schema has the given `${_privilege}` set.


TheBigLee:
This PR has been superseded by #1096.
I will therefore close this PR.

PR #1056, opened 2019-02-11
MODULES-8616 Check for permissions to be present on all tables, not j…

wimvr:
Check for permissions to be present on all tables, not just one

PR #1096, opened 2019-07-24
Grant all tables in schema fixup

georgehansper:
This pull request addresses the issues in:

* https://github.com/puppetlabs/puppetlabs-postgresql/pull/1056
* https://github.com/puppetlabs/puppetlabs-postgresql/pull/1095

In summary, the problem was that the code was working for adding a single privilege or ALL PRIVILEGES a _single_ table but failed to invoke the necessary changes if there was one or more tables already present that had the required single privilege or all privileges.

I have included the code from https://github.com/puppetlabs/puppetlabs-postgresql/pull/1056 verbatim, and adapted it for the case of ALL PRIVILEGES

The code in pull request 1056 is a bit more readable than the original code, I feel.

I have also updated the associated tests in grant_spec.rb to detect the original problem.

### DIFF
--- a/spec/acceptance/server/grant_spec.rb
+++ b/spec/acceptance/server/grant_spec.rb
@@ -225,6 +225,7 @@ describe 'postgresql::server::grant:', unless: UNSUPPORTED_PLATFORMS.include?(os
           postgresql_psql { "grant all on table test_tbl2 to ${user}":
             command   => "GRANT ALL ON TABLE test_tbl2 TO ${user}",
             db        => $db,
+            psql_user => $owner,
             unless    => "SELECT 1 FROM information_schema.role_table_grants WHERE table_name = 'test_tbl2' AND grantee = '${user}' HAVING count(*)>=7",
             require   => [ Postgresql::Server::Database[$db], Postgresql_psql['create test table 2'], Postgresql::Server::Role[$user] ],
           }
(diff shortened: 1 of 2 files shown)


sheenaajay:
Thanks for submitting the PR @georgehansper.

spec/acceptance/server/grant_spec.rb
```diff
@@ -263,7 +277,7 @@ class { 'postgresql::server': }
           EOS
 
           if Gem::Version.new(postgresql_version) >= Gem::Version.new('9.0')
-            idempotent_apply(pp)
+            apply_manifest(pp, expect_changes: true)
```
sheenaajay: Could you please add the idempotent_apply function back since we need to test the idempotency also which applies the manifest twice.

georgehansper: I have update the tests to use the idempotent_apply() function.

idempotent_apply() is done in 2 stages, because the 1st one sets up the table permissions, and the 2nd one (which uses postgresql::server::grant) 'fixes' the permissions.

When this was a single idempotent_apply() the testbench complained because postgresql::server::grant was making changes every time (as it was supposed to).

Let me know if you want further changes made to the tests.

georgehansper:
Thanks @sheenaajay for merging this PR

PR #1099, opened 2019-08-02
(MODULES-9658) - custom ports are not labeled correctly

blackknight36:
When a node has SELinux enabled the port must be labeled properly to allow the postgresql service to start.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -131,6 +131,7 @@ class postgresql::server (
   $manage_pg_hba_conf         = $postgresql::params::manage_pg_hba_conf,
   $manage_pg_ident_conf       = $postgresql::params::manage_pg_ident_conf,
   $manage_recovery_conf       = $postgresql::params::manage_recovery_conf,
+  Boolean $manage_selinux     = $postgresql::params::manage_selinux,
   $module_workdir             = $postgresql::params::module_workdir,
 
   $manage_datadir             = $postgresql::params::manage_datadir,
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -108,9 +108,28 @@ class postgresql::server::config {
 
   # ensure that SELinux has a proper label for the port defined
   if $postgresql::server::manage_selinux == true and $facts['selinux'] == true {
+    case $facts['osfamily'] {
+      'RedHat', 'Linux': {
+        if $facts['operatingsystem'] == 'Amazon' {
+          $package_name = 'policycoreutils'
+        }
+        else {
+          $package_name = $facts['operatingsystemmajrelease'] ? {
+            '5'     => 'policycoreutils',
+            '6'     => 'policycoreutils-python',
+            '7'     => 'policycoreutils-python',
+            default => 'policycoreutils-python-utils',
+          }
+        }
+      }
+    }
+
+    ensure_packages([$package_name])
+
     exec { "/usr/sbin/semanage port -a -t postgresql_port_t -p tcp ${port}":
-        unless => "/usr/sbin/semanage port -l | grep -qw ${port}",
-        before => Postgresql::Server::Config_entry['port'],
+        unless  => "/usr/sbin/semanage port -l | grep -qw ${port}",
+        before  => Postgresql::Server::Config_entry['port'],
+        require => Package[$package_name],
     }
   }
 
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -106,9 +106,18 @@ class postgresql::server::config {
     }
   }
 
+  # ensure that SELinux has a proper label for the port defined
+  if $postgresql::server::manage_selinux == true and $facts['selinux'] == true {
+    exec { "/usr/sbin/semanage port -a -t postgresql_port_t -p tcp ${port}":
+        unless => "/usr/sbin/semanage port -l | grep -qw ${port}",
+        before => Postgresql::Server::Config_entry['port'],
+    }
+  }
+
   postgresql::server::config_entry { 'port':
     value => $port,
   }
+
   postgresql::server::config_entry { 'data_directory':
     value => $datadir,
   }
--- a/README.md
+++ b/README.md
@@ -333,9 +333,9 @@ While this module supports both 1.x and 2.x versions of the 'puppetlabs-apt' mod
 
 PostGIS is currently considered an unsupported feature, as it doesn't work on all platforms correctly.
 
-### All versions of RHEL/CentOS
+### All versions of RHEL/CentOS with manage_selinux => false
 
-If you have SELinux enabled you must add any custom ports you use to the `postgresql_port_t` context.  You can do this as follows:
+If you have SELinux enabled and you are *not* using the selinux module to manage SELinux (this is the default configuration) you will need to label any custom ports you use with the `postgresql_port_t` context.  The postgresql service will not start until this is done.  To label a port use the semanage command as follows:
 
 ```shell
 semanage port -a -t postgresql_port_t -p tcp $customport
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -143,6 +143,7 @@ class postgresql::globals (
   $manage_pg_hba_conf       = undef,
   $manage_pg_ident_conf     = undef,
   $manage_recovery_conf     = undef,
+  $manage_selinux           = undef,
 
   $manage_package_repo      = undef,
   $module_workdir           = undef,
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -21,6 +21,7 @@ class postgresql::params inherits postgresql::globals {
   $manage_pg_hba_conf         = pick($manage_pg_hba_conf, true)
   $manage_pg_ident_conf       = pick($manage_pg_ident_conf, true)
   $manage_recovery_conf       = pick($manage_recovery_conf, false)
+  $manage_selinux             = pick($manage_selinux, false)
   $package_ensure             = 'present'
   $module_workdir             = pick($module_workdir,'/tmp')
 
--- a/spec/acceptance/alternative_port_spec.rb
+++ b/spec/acceptance/alternative_port_spec.rb
@@ -8,7 +8,7 @@ describe 'postgresql::server', unless: UNSUPPORTED_PLATFORMS.include?(os[:family
   end
   it 'on an alternative port' do
     pp = <<-MANIFEST
-      class { 'postgresql::server': port => '55433' }
+      class { 'postgresql::server': port => '55433', manage_selinux => true }
     MANIFEST
 
     idempotent_apply(pp)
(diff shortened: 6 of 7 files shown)


metadata.json
```diff
@@ -19,6 +19,10 @@
     {
       "name": "puppetlabs/concat",
       "version_requirement": ">= 1.1.0 < 7.0.0"
+    },
+    {
+      "name": "puppet/selinux",
```
bastelfreak: hey people. I'm wondering if the selinux module should stay a soft dependency because it's just used at one place in this module.

manifests/server/config.pp
```diff
@@ -106,9 +106,20 @@
     }
   }
 
+  if $facts['selinux'] {
+    selinux::port { 'postgresql_custom_port':
```
bastelfreak: it would be nice if this could be enabled/disabled with a new parameter like `$postgresql::manage_selinux`. Many people I know already manage selinux settings in their profiles.

blackknight36: @bastelfreak I can add that to the manifest.  Should the default value be set to false?

bastelfreak: My personal opinion would be that the default should be false, otherwise a major release would be required. But that's something that the module owners should decide.

metadata.json
blackknight36: @bastelfreak selinux is a pretty critical part of the OS and if this module needs to modify SELinux settings I would say it is not a soft dependency.

manifests/server.pp
```diff
@@ -127,6 +127,7 @@
   $manage_pg_hba_conf         = $postgresql::params::manage_pg_hba_conf,
   $manage_pg_ident_conf       = $postgresql::params::manage_pg_ident_conf,
   $manage_recovery_conf       = $postgresql::params::manage_recovery_conf,
+  $manage_selinux             = $postgresql::params::manage_selinux,
```
bastelfreak: You can add a datatype to this:
```suggestion
  Boolean $manage_selinux     = $postgresql::params::manage_selinux,
```

metadata.json
bastelfreak: but it's only a critical part if it's a box with selinux where selinux is actually enabled and not handled already in another profile/module. I assume that's a valid statement for only a small number of users.

blackknight36: I understand but selinux is enabled by default in both RHEL and Fedora which covers a lot of use cases.  How would I mark a module as a soft dependency?  There doesn't seem to be an option for that in puppet metadata files.

manifests/server.pp
blackknight36: I've added the data type to this parameter.

blackknight36:
Tests appear to be failing due to a forge error.

```
localhost:2222 failed {"stdout"=>"\e[mNotice: Preparing to install into /etc/puppetlabs/code/environments/production/modules ...\e[0m\n\e[mNotice: Downloading from https://forgeapi.puppet.com ...\e[0m\n", "stderr"=>"\e[1;31mError: execution expired\e[0m\n\e[1;31mError: Try 'puppet help module install' for usage\e[0m\n", "exit_code"=>1, "_error"=>{"kind"=>"puppetlabs.tasks/command-error", "issue_code"=>"COMMAND_ERROR", "msg"=>"The command failed with exit code 1", "details"=>{"exit_code"=>1}}}
```

Is there any way to resolve this?

lionce:
Hello @blackknight36 ,

This PR looks good. Unfortunately acceptance tests are failing (Failed on 5 nodes: redhat-5-x86_64",centos-5-x86_64", oracle-5-x86_64", oracle-6-x86_64", oracle-7-x86_64"). Would it be possible to check them using litmus? Please let us know if you need any help!

Cheers!

`Failures:
    
      1) postgresql::server on an alternative port
         On host `f16svhg5g3ghtgd.delivery.puppetlabs.net'
         Failure/Error: idempotent_apply(pp)
         RuntimeError:
           apply manifest failed
           ` puppet apply /tmp/manifest_20190813_82428_1s1svk.pp`
           ======
           [{"node"=>"f16svhg5g3ghtgd.delivery.puppetlabs.net", "target"=>"f16svhg5g3ghtgd.delivery.puppetlabs.net", "action"=>"command", "object"=>" puppet apply /tmp/manifest_20190813_82428_1s1svk.pp", "status"=>"failure", "result"=>{"stdout"=>"", "stderr"=>"\e[1;31mError: Evaluation Error: Error while evaluating a Resource Statement, Unknown resource type: 'selinux::port' (file: /etc/puppetlabs/code/environments/production/modules/postgresql/manifests/server/config.pp, line: 110, column: 5) on node f16svhg5g3ghtgd.delivery.puppetlabs.net\e[0m\n", "exit_code"=>1, "_error"=>{"kind"=>"puppetlabs.tasks/command-error", "issue_code"=>"COMMAND_ERROR", "msg"=>"The command failed with exit code 1", "details"=>{"exit_code"=>1}}}}]
           
         # ./vendor/bundle/ruby/2.5.0/gems/puppet_litmus-0.8.0/lib/puppet_litmus/serverspec.rb:54:in `apply_manifest'
         # ./vendor/bundle/ruby/2.5.0/gems/puppet_litmus-0.8.0/lib/puppet_litmus/serverspec.rb:12:in `idempotent_apply'
         # ./spec/acceptance/alternative_port_spec.rb:14:in `block (2 levels) in <top (required)>'
    
      2) postgresql::server can connect with psql
         On host `f16svhg5g3ghtgd.delivery.puppetlabs.net'
         Failure/Error: run_shell("cd /tmp; su #{shellescape(user)} -c #{shellescape(psql)}", acceptable_exit_codes: exit_codes, &block)
         RuntimeError:
           shell failed
           `cd /tmp; su postgres -c psql\ -p\ 55433\ --command\=\"\\l\"\ postgres`
           ======
           [{"node"=>"f16svhg5g3ghtgd.delivery.puppetlabs.net", "target"=>"f16svhg5g3ghtgd.delivery.puppetlabs.net", "action"=>"command", "object"=>"cd /tmp; su postgres -c psql\\ -p\\ 55433\\ --command\\=\\\"\\\\l\\\"\\ postgres", "status"=>"failure", "result"=>{"stdout"=>"", "stderr"=>"su: user postgres does not exist\n", "exit_code"=>1, "_error"=>{"kind"=>"puppetlabs.tasks/command-error", "issue_code"=>"COMMAND_ERROR", "msg"=>"The command failed with exit code 1", "details"=>{"exit_code"=>1}}}}]
           
         # ./vendor/bundle/ruby/2.5.0/gems/puppet_litmus-0.8.0/lib/puppet_litmus/serverspec.rb:106:in `run_shell'
         # ./spec/spec_helper_acceptance_local.rb:24:in `psql'
         # ./spec/acceptance/alternative_port_spec.rb:22:in `block (2 levels) in <top (required)>'
    
      3) postgresql::server Port "55433" should be listening
         On host `f16svhg5g3ghtgd.delivery.puppetlabs.net'
         Failure/Error: it { is_expected.to be_listening }
           expected Port "55433" to be listening
           /bin/sh -c ss\ -tunl\ \|\ grep\ -E\ --\ :55433\\\ 
           
         # ./spec/acceptance/alternative_port_spec.rb:18:in `block (3 levels) in <top (required)>'`

blackknight36:
@lionce Where are you seeing these failures?  TravisCI shows that all checks passed, is there another service that I need to check?

blackknight36:
I suspect the failures are due to those distros not having an "os" fact mocked up when the tests run.  The selinux module uses this fact and puppet runs will fail if it is not defined.

lionce:
Hello @blackknight36 ,

We're using litmus to execute acceptance tests. You can run the tests following [this turorial](https://github.com/puppetlabs/puppet_litmus/wiki/Tutorial:-use-Litmus-to-execute-acceptance-tests-with-a-sample-module-(MoTD))
We ran tests on all platforms defined in [provision.yaml file](https://github.com/puppetlabs/puppetlabs-accounts/blob/master/provision.yaml) under release_checks section. 

Cheers!

blackknight36:
@lionce Thanks for the links.  I'll have to see if I can get Litmus working on my system.

> Evaluation Error: Error while evaluating a Resource Statement, Unknown resource type: 'selinux::port'

This means that the selinux module is missing.  I updated the .fixtures.yml file to ensure that the module is installed but it looks like certain distros aren't honoring this file properly.

blackknight36:
@lionce I just tried running the litmus test for Redhat 5 and received the following error:

```
Failed provisioning redhat-5-x86_64 using vmpooler
{"node"=>"localhost", "target"=>"localhost", "action"=>"task", "object"=>"provision::vmpooler", "status"=>"failure", "result"=>{"_output"=>"Cannot file fog file at /home/00/d861703/.fog\n{\"_error\":{\"kind\":\"facter_task/failure\",\"msg\":\"Failed to open TCP connection to vcloud.delivery.puppetlabs.net:80 (getaddrinfo: Name or service not known)\"}}\n", "_error"=>{"kind"=>"puppetlabs.tasks/task-error", "issue_code"=>"TASK_ERROR", "msg"=>"The task failed with exit code 1", "details"=>{"exit_code"=>1}}}}
```
How would I fix this?  Is there a configuration file that controls what cloud service is used?

metadata.json
michaeltlombardi: So, my understanding is that there is a case in which this module will error if the selinux module is not present, and that is where:

- the target node has selinux enabled
- selinux is not being managed by another profile or module
- the node is being managed with this module

Puppet does not distinguish between hard/soft deps, a module is either required or not in the metadata. One option here would be to document the edge case if you don't want to actually include the module in the metadata dependencies.

blackknight36: The module will error out if selinux is enabled and the postgresql alternate port is not labeled properly before the service is started.  The default SELinux policies only allow postgresql to bind to port 5432.

I originally added selinux as a dependency in the metadata file but removed it as I was told we don't want to make it a hard requirement.

IMO the selinux is almost as essential as stdlib.  SELinux is enabled by default on Redhat, Fedora, and I believe OracleLinux since that is based on RHEL but I could be mistaken.

blackknight36: I've done some additional checking and it looks like the puppet-selinux module does not support RedHat 5 so I will need to revisit how to manage SELinux ports.  I will updated this PR once I have a solution.

blackknight36: I've updated the manifest to use an exec resource instead of the selinux::port type.  This eliminates the dependency on the selinux module and will work with all platforms supported by puppet.

blackknight36:
@lionce Tests should pass now.  I have removed the dependency on the selinux module.

florindragos:
@blackknight36 There's still acceptance tests failing on Oracle 6 and Oracle 7. These platforms might not come with `semanage` installed: 
`Dependency Exec[/usr/sbin/semanage port -a -t postgresql_port_t -p tcp 55433] has failures: true\e[0m\n\e[mNotice: Applied catalog in 0.25 seconds\e[0m\n", "stderr"=>"\e[1;31mError: /Stage[main]/Postgresql::Server::Config/Exec[/usr/sbin/semanage port -a -t postgresql_port_t -p tcp 55433]: Could not evaluate: Could not find command '/usr/sbin/semanage'`

florindragos:
Cool! `Successful on 19 nodes` 👍

PR #1108, opened 2019-10-07
Remove duplicate REFERENCE.md file with strange unicode character at end of filename

nudgegoonies:
The file is duplicated but they have differences:
REFERENCE.md
REFERENCE.md\342\200\202
I deleted the additional added one with the strange unicode character at the end of filename. But maybe the original committer must check what version is the one to use @tphoney


tphoney:
Thanks @nudgegoonies  !!
great catch, i thought i was missing something, and i did.

tphoney:
@nudgegoonies the docs will be re-generated anyway when we next release the module.

DavidS:
Upgraded to `bugfix`, as the duplicate file can cause issues on PE LTS 2018.1 and was released as part of 6.2.0:

```
2020-01-09T02:25:31.353Z ERROR [clojure-agent-send-off-pool-16] [p.e.s.f.file-sync-storage-core] Fatal error during file sync commit, requesting shutdown.
java.nio.file.InvalidPathException: Malformed input or input contains unmappable characters: /etc/puppetlabs/code-staging/environments/production/modules/postgresql/REFERENCE.md???
```

PR #1120, opened 2019-11-28
pdksync - FM-8499 remove ubuntu 14 support

lionce:
FM-8499 remove ubuntu 14 support
pdk version: `1.14.0`


DavidS:
Added a commit removing old Ubuntu 14.04 logic. looking at the affected files, there's also some old Debian cruft in there that could be removed/cleaned up.

sheenaajay:
Closing the PR since we are adding the support back.Thank you.

PR #1131, opened 2020-01-13
fix missing systemd override config for EL8 (CentOS and RHEL)

david-barbion:


sheenaajay:
Thanks for submitting the PR @david-barbion .

PR #1135, opened 2020-01-29
Treat $version as an integer for comparison, defaults to string

boydtom:
When setting password_encryption an operator AND calls a comparison of the $version variable to an integer.  $version is a string - breaking the comparison.  This should fix that - although it may not be the best way to do so and someone with greater skill has a better method.

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -141,7 +141,7 @@ class postgresql::server::config {
     value => $port,
   }
 
-  if ($password_encryption) and ($version >= 10){
+  if ($password_encryption) and (versioncmp($version, '10') >= 10){
     postgresql::server::config_entry { 'password_encryption':
       value => $password_encryption,
     }


manifests/server/config.pp
```diff
@@ -141,7 +141,7 @@
     value => $port,
   }
 
-  if ($password_encryption) and ($version >= 10){
+  if ($password_encryption) and ($version.scanf('%i')[0] >= 10){
```
cmccrisken-puppet: You could make use of the [versioncmp](https://puppet.com/docs/puppet/latest/function.html#versioncmp) function, like this:
```
if ($password_encryption) and (versioncmp($version, "10") >= 0)
```
I've run the acceptance and unit tests over this and it seems to be OK.

cmccrisken-puppet: Also spotted another implementation of this in a pending PR: https://github.com/puppetlabs/puppetlabs-concat/pull/627/files#diff-07b462a4b4082856b785d62dee4f7468R42

boydtom: @cmccrisken-puppet I prefer this fix method actually - it's less opaque than mine. I will test it out.  Thanks!

boydtom: I've tested the alternate method and like the results.  I've rebased the commit history to a single commit.  I'm good with the fix if you all are.

cmccrisken-puppet:
LGTM @boydtom - thanks again for the fix! 👍

PR #1137, opened 2020-02-04
Correct versioncmp logic in config.pp

boydtom:
This is a bit embarrassing but my last patch, although not breaking anything from prior state, also will not actually work as intended.  Version comparison works but versioncmp produces results of -1, 0 or 1 and must be compared against that result.  

This fix I tested more thoroughly.  :)  Sorry for the hassle.

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -141,7 +141,7 @@ class postgresql::server::config {
     value => $port,
   }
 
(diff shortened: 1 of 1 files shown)


cmccrisken-puppet:
@boydtom I should have caught this in the review, so there's at least equal, if not more blame on myself! Thanks for getting a fix together so quickly.

PR #1148, opened 2020-02-26
Add Ubuntu 19.04/19.10 compatibility

pedrocr:


michaeltlombardi:
Hi @pedrocr! Thank you for this submission!

However, for supported modules, we can only add [those platforms which are officially supported by Puppet](https://puppet.com/docs/pe/2019.2/supported_operating_systems.html) to the compatibility list in the metadata - the module _may_ work for other platforms but we cannot make any guarantee.

Because 19.04 and 19.10 are not LTS releases - and therefore not in our supported OS list - I'm going to close out this PR.

pedrocr:
This is a really poor policy. I run puppet both for servers (that are on LTS) and on desktop to try things earlier. This is literally the only thing that breaks when I upgrade and it's a really poor user experience to have to go spelunking into an unkown codebase to patch something so basic. There are better ways to signal that you're on an unsupported distro than failing with an error deep inside the code.

michaeltlombardi:
I can raise with the team to discuss re-opening this and look to merge with the understanding that there's a difference between "supported on" and "possibly compatible with" - as a team maintaining over fifty modules, we've got some constraints around what we can actually commit to supporting. In general, if the platform is not listed in the metadata, I would not rely on the module to work for that platform on production systems (though, as you noted, you're looking to test functionality).

pedrocr:
That sounds reasonable to me, thanks.

PR #1150, opened 2020-03-02
Fix incorrectly quoted GRANT cmd on functions

olifre:
The existing code failed miserably to grant permissions on functions. 

Using the following code:
```puppet
postgresql::server::grant { "grant_func_privs":
  role                    => 'rewind',
  db                      => 'postgres',
  privilege               => 'EXECUTE',
  object_type             => 'FUNCTION',
  object_name             => 'pg_catalog.pg_ls_dir',
  object_arguments        => ['text', 'boolean', 'boolean'],
  require                 => Postgresql::Server::Role[$rewind_user],
}
```
The created queries were:
```sql
GRANT EXECUTE ON FUNCTION "pg_catalog.pg_ls_dir"("text","boolean","boolean") TO "rewind";
```
and for the "unless" case:
```sql
SELECT COUNT(*) FROM (SELECT 1 WHERE has_function_privilege('rewind', 'pg_catalog.pg_ls_dir("text","boolean","boolean")', 'EXECUTE') = true);"
```
The patches in this series fix that to be:
```sql
GRANT EXECUTE ON FUNCTION pg_catalog.pg_ls_dir(text,boolean,boolean) TO "rewind";
SELECT COUNT(*) FROM (SELECT 1 WHERE has_function_privilege('rewind', 'pg_catalog.pg_ls_dir(text,boolean,boolean)', 'EXECUTE') = true);"
```
since both the types and the complete function expression must not be quoted.

### DIFF
--- a/manifests/server/grant.pp
+++ b/manifests/server/grant.pp
@@ -392,7 +392,7 @@ define postgresql::server::grant (
         true    => 'function_exists',
         default => undef,
       }
-      $_quoted_args = join($object_arguments.map |$arg| { "\"${arg}\"" }, ',')
+      $_quoted_args = join($object_arguments, ',')
       $arguments = "(${_quoted_args})"
     }
 
--- a/manifests/server/grant.pp
+++ b/manifests/server/grant.pp
@@ -383,7 +383,7 @@ define postgresql::server::grant (
         default => undef,
       }
       $arguments = ''
-     $_enquote_object = false
+      $_enquote_object = false
     }
     'FUNCTION': {
       $unless_privilege = $_privilege ? {
--- a/manifests/server/grant.pp
+++ b/manifests/server/grant.pp
@@ -50,10 +50,12 @@ define postgresql::server::grant (
     default: {
       # default is 'present'
       $sql_command = 'GRANT %s ON %s "%s%s" TO "%s"'
+      $sql_command_unquoted = 'GRANT %s ON %s %s%s TO "%s"'
       $unless_is = true
     }
     'absent': {
       $sql_command = 'REVOKE %s ON %s "%s%s" FROM "%s"'
+      $sql_command_unquoted = 'REVOKE %s ON %s %s%s FROM "%s"'
       $unless_is = false
     }
   }
@@ -115,6 +117,7 @@ define postgresql::server::grant (
         'absent' =>  'role_exists',
       }
       $arguments = ''
+      $_enquote_object = true
     }
     'SCHEMA': {
       $unless_privilege = $_privilege ? {
@@ -131,6 +134,7 @@ define postgresql::server::grant (
       $on_db = $db
       $onlyif_function = undef
       $arguments = ''
+      $_enquote_object = true
     }
     'SEQUENCE': {
       $unless_privilege = $_privilege ? {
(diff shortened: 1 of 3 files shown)


manifests/server/grant.pp
```diff
@@ -375,6 +383,7 @@
         default => undef,
       }
       $arguments = ''
+     $_enquote_object = false
```
cmccrisken-puppet: Fixing indentation typo that caused syntax checks to fail:
```suggestion
      $_enquote_object = false
```

cmccrisken-puppet: Fixes syntax check failures:
```suggestion
      $_enquote_object = false
```

olifre:
Thanks for the friendly feedback!
I was sadly unable to run PDK on my platform with this module. It seems something is very broken my platform (Gentoo), since even after running:
```
pdk bundle install --path vendor/bundle
```
(which succeeds fine and installs the dependencies), trying to `validate` or `unit test` will make PDK attempt to reinstall all dependencies with `sudo`, completely ignoring they are already there in `vendor/bundle`. 

Unless you have an idea what is wrong, I could only develop the tests "blindly" (using travis-feedback). I can certainly do that, but it may take a bit.

olifre:
@sanfrancrisko That should have done the trick to get the Static and Spec parts green. 

I can also look at the acceptance tests, they should run through fine now, but they only ever worked since they never tested granting permissions to a function taking parameters, so it would be better to add something like this to accepteance tests (I can try tomorrow, hopefully won't take too many iterations).

olifre:
I did now also add an acceptance test for granting `EXECUTE` to functions with arguments. 
Must have been beginner's luck, it seems to have played out well on my first attempt (and that was my first acceptance test ever) ;-).

sanfrancrisko:
Thanks for updating the current tests and adding the new coverage too @olifre. Appreciate you doing this as your first acceptance test! 😄 

Change LGTM!

Regarding the issues you were seeing with the PDK - I asked some of my more PDK-savvy colleagues what they thought might be the issue. A few suggestions:
- Drop the `--path` parameter from the `bundle install` command and let the PDK handle this
- If you're using Bundler `>= 2` you can set the path (seeing as using `--path` is now deprecated) globally: `pdk bundle config set path <PATH>` (FWIW, I have it set it to `.bundle`)

Before running `pdk bundle install` and `pdk test unit`, I would recommend removing the `Gemfile.lock` and `vendor` directory from your repo. Let me know how it works out.

olifre:
@sanfrancrisko Thanks for the friendly feedback and the PDK help!
Sadly, it fails for me — I have PDK 1.8.0.0 (latest packaged version for Gentoo) and Bunder 1.17.2 (newer versions are available up to 2.1.4, but I should probably not mix them with an old PDK). 

After cleaning up (removing {{Gemfile.lock}} and {{vendor}}) and then running:
```
pdk bundle install
```
it tries to run `sudo` and wants to overwrite my system's ruby installation :-(. 
For the config (even with bundler 2) I would expect it won't work since PDK sets `BUNDLE_IGNORE_CONFIG: 1` (as I see in the debug output) so any config would be ignored — no? Potentially, that's also why the `path` setting in the older bundler version fails.

PR #1154, opened 2020-03-20
Update contributing link url

igrcic:
It seems that the contributing link hasn't been updated to newest puppet modules contributing link, this simple PR should fix that, if i got the url right


daianamezdrea:
Hi @igrcic , thank you for your contribution ! Yes, the url is good .

PR #1155, opened 2020-03-23
server::extension: Excempt system database "template1" from requires.

olifre:
The system database `template1` is also always present and does not need management by this module.

### DIFF
--- a/manifests/server/extension.pp
+++ b/manifests/server/extension.pp
@@ -50,8 +50,8 @@ define postgresql::server::extension (
     }
   }
 
-  if( $database != 'postgres' ) {
-    # The database postgres cannot managed by this module, so it is exempt from this dependency
+  if( ($database != 'postgres') and ($database != 'template1') ) {
+    # The system databases postgres and template1 cannot managed by this module, so they is exempt from this dependency
     Postgresql_psql {
       require => Postgresql::Server::Database[$database_resource_name],
     }


carabasdaniel:
Hi @olifre,

According to the official documentation [here](https://www.postgresql.org/docs/9.4/manage-ag-templatedbs.html) the template1 is still used by default installation of Postgresql on the CREATE DATABASE command, therefore having the possibility to manage this would allow you to define the templates you need in your next database creations. As a result of this fact we sadly have to decline this PR for now. 

Thank you for your contributions and we are sorry for any inconvenience caused.

olifre:
Hi @carabasdaniel ,

I see! Just to clarify, this change was not about removing the _possibility_ to manage `template1`, but just about removing the _enforcement_ of having to manage the `template1` database with this module to be able to define a `postgresql::server::extension` resource using the `template1` database. 

I do not completely follow why the database itself, which is guaranteed to exist (as documentation still states), must also be managed by the module to be allowed to enable an extension on it (while this is not the case for other builtin databases such as `postgres`). 

In any case, if somebody stumbles upon this at some point, the workaround is simple — just add this to your catalogue:
```
        postgresql::server::database { 'template1':
                istemplate      => true,
        }
```
and then `postgresql::server::extension` will also work with the `template1` database. 

Cheers and thanks,
Oliver

PR #1156, opened 2020-03-23
add extra parameter "extra_systemd_config" 

veninga:
This add an extra parameter to make it possible to add lines to /etc/systemd/system/portgresql-<version>.service.
This is needed for instance to provide for extra open files for postgres.
LimitNOFILE=10000

This is implemented as a multi line string. This way you can add multiple lines to the systemd file.


daianamezdrea:
Hi @veninga , thank you for your contribution !

PR #1159, opened 2020-03-30
Allow usage of grant and role when not managing postgresql::server

tuxmea:
Some installations use other ways to install postgresql.
But people still want to make use of grant and role defined type.

This PR moves all variables with default values set to `postgresql::server::<var>` into the defined types parameter list.
A spec test was added to ensure that postgresql::server class is not part of the catalog.


carabasdaniel:
Hi @tuxmea,

Thank you for your contribution.

PR #1160, opened 2020-04-01
(MODULES-10610) Use correct lower bound for concat version

ghoneycutt:


sheenaajay:
Thanks @ghoneycutt for submitting the PR.

alexjfisher:
Should this really have the `backwards-incompatible` label?  If the old information was wrong, then IMO, fixing it isn't a breaking change.

ghoneycutt:
@sheenaajay @alexjfisher This should not have the backwards-incompatible label applied. The old information was incorrect. This PR fixes that incorrect metadata, it does not change functionality and certainly not in a way that would be backwards-incompatible.

ghoneycutt:
For perspective, the force attribute to concat in `manifests/server/recovery.pp` was introduced in 5.6.0, so the metadata has been broken since that release about two years ago.

sheenaajay:
Thanks @ghoneycutt . Updated the label.

PR #1163, opened 2020-04-15
Finish conversion of `postgresql_acls_to_resources_hash` function

alexjfisher:
In https://github.com/puppetlabs/puppetlabs-postgresql/pull/1129 the 3 functions from this module were ported to the modern function API and namespaced. None of the boilerplate was fixed up, tests weren't updated, documentation wasn't updated etc.

This commit finishes the work for the `postgresql_acls_to_resources_hash` function.

The existing function documentation made it pretty clear that this function is for internal use only.  As such, I've marked it `@api private` and haven't created a non-namespaced shim (like I've being doing with the other functions in this module).


alexjfisher:
I see no evidence that the modules listed by the bot above are using the non-namespaced version of the function I've removed. (I've only checked a few.  I'm not going to go through them all, as that is something the bot could do itself??)

daianamezdrea:
Hi @alexjfisher, thank you for your contribution !

PR #1165, opened 2020-04-20
Fix custom port in extension

Vampouille:
When postgres server does not use a standard port, the resource extension still try to connect with standard postgres port instead of the port configured in `$postgresql::server::port`.
This PR add a test with a custom port to demonstrate issue. A commit to fix the issue will be append.


daianamezdrea:
Hello @Vampouille, thank you for your contribution! Cheers!

PR #1167, opened 2020-04-29
service_ensure => true is now an allowed value (aliased to running) 

binford2k:


sanfrancrisko:
Thanks for the tweak @binford2k! 👍

PR #1179, opened 2020-06-18
Invert psql/package dependency logic

raphink:
When instantiating several times the same PostgreSQL extension with the same
package name (e.g. `pg_trgm` and `rh-postgresql10-postgresql-contrib.x86_64`),
the current code leads to a duplicate declaration due to the clash on the
`before` parameter on the package passed through `ensure_packages`.

This PR inverts the logic to ensure that no `require` or `before` parameter is
passed to `ensure_packages`.


daianamezdrea:
Hi @raphink, this looks good! Thank you for your contribution, I'll merge the PR!

PR #1181, opened 2020-07-27
(MAINT) Indicate supported PostgreSQL versions

npwalker:
Prior to this commit, there was an old limitation noted that we only support up to PostgreSQL 9.5.  

After this commit, we remove that limitation because we support whatever the default for each OS is.


david22swan:
@npwalker 
Making a quick change to point your PR at the newly default `main` branch. 
Apologies if this is inconvenient in any way.

PR #1187, opened 2020-08-10
Allow removal of config_entries via main class

ekohl:
This allows removal of config entries by setting their value to undef.  In doing so, users can remove values via hiera. One example is checkpoint_segments which was removed in PostgreSQL 9.5, but commonly used as a tuning parameter in older versions.

I'd like some verification on whether undef values really couldn't be passed in for some other use case, but I can't imagine any.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -184,7 +184,8 @@ class postgresql::server (
 
   $config_entries.each |$entry, $value| {
     postgresql::server::config_entry { $entry:
-      value => $value,
+      ensure => bool2str($value =~ Undef, 'absent', 'present'),
+      value  => $value,
     }
   }
 
--- a/spec/unit/classes/server_spec.rb
+++ b/spec/unit/classes/server_spec.rb
@@ -209,13 +209,15 @@ describe 'postgresql::server', type: :class do
         config_entries: {
           fsync: 'off',
           checkpoint_segments: '20',
+          remove_me: :undef,
         },
       }
     end
 
     it { is_expected.to compile.with_all_deps }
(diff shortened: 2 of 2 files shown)


david22swan:
@ekohl 
Making a quick change to point your PR at the newly default `main` branch. 
Apologies if this is inconvenient in any way.

sanfrancrisko:
I may circulate your query around the IAC Team @ekohl : 
> I'd like some verification on whether undef values really couldn't be passed in for some other use case, but I can't imagine any.

But, in principle, it seems OK to me.

It does appear as though the test you added is failing. From what I can see, in [this loop](https://github.com/puppetlabs/puppetlabs-postgresql/pull/1187/files#diff-6fa83dd25c6a65d4afde2eb252c8b945R185-R189), we're receiving:
```
$entry = fsync
$value = 'off'
```
...four times. I'm not seeing any of the additional params being evaluated.

ekohl:
It looks like `nil` isn't mapped to `undef` in the tests. Trying with `:undef` to see if that's mapped to `undef`.

ekohl:
Spec tests pass now. The acceptance tests fail, but those look unrelated to my changes.

daianamezdrea:
Hi @ekohl, indeed, the failures from acceptance tests are not from this PR, thank you for submitting the PR! I'll have a look to see what is happening here. Cheers!

sanfrancrisko (approved):
Thanks for the enhancement @ekohl and addressing the test failures.

I've rebased with `puppetlabs:main` to pull in a workaround for the acceptance test failures. I'll merge the PR when the tests complete (and go green).

PR #1188, opened 2020-08-27
Fix contrib package name under debian 10

neomilium:
Under Debian Buster, when we use only debian repo (ie. without using postgresql repo), the PostgreSQL contrib package `postgresql-contrib-11` does not exist, only `postgresql-contrib` is available.

This PR fixes the wrong package name.

Note: Without this fix patch, module usage does not fail during puppet run, instead it notices the following message every run:
```log
Notice: /Stage[main]/Postgresql::Server::Contrib/Package[postgresql-contrib]/ensure: created (corrective)
```

IMHO, this behavior is due to some magic done by `apt`, here the log when I run manually the install with the wrong package name:
```shell-session
$ sudo apt install postgresql-contrib-11
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Note, selecting 'postgresql-11' instead of 'postgresql-contrib-11'
postgresql-11 is already the newest version (11.7-0+deb10u1).
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -187,7 +187,7 @@ class postgresql::params inherits postgresql::globals {
           /^6/    => pick($java_package_name, 'libpg-java'),
           default => pick($java_package_name, 'libpostgresql-jdbc-java'),
         },
-      default  => pick($java_package_name, 'libpostgresql-jdbc-java'),
+        default  => pick($java_package_name, 'libpostgresql-jdbc-java'),
       }
       $perl_package_name      = pick($perl_package_name, 'libdbd-pg-perl')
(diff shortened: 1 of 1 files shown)


sanfrancrisko:
Thanks for this fix @neomilium and for the detailed explanation of the issue. This change LGTM!

Could I ask if you could rebase your branch with `puppetlabs:main` - I tried to do this to pull in a workaround for the current failing CI tests, but I do not have permission to push to your fork. 

The tests will automatically kick off again and then I'll merge when we've got the green light 😃

neomilium:
@sanfrancrisko Branch rebased, have fun!

PR #1190, opened 2020-09-03
Add new common repo which contains add-ons

jorhett:
Detailed in https://yum.postgresql.org/news/new-repo-rpms-released/, many add-ons and extensions were moved to a common repo. Without this common repo certain addon installations will fail.

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -57,6 +57,7 @@
 # @param repo_proxy Sets the proxy option for the official PostgreSQL yum-repositories only.
 #
 # @param repo_baseurl Sets the baseurl for the PostgreSQL repository. Useful if you host your own mirror of the repository.
+# @param yum_repo_commonurl Sets the url for the PostgreSQL common Yum repository. Useful if you host your own mirror of the YUM repository.
 #
 # @param needs_initdb Explicitly calls the initdb operation after the server package is installed and before the PostgreSQL service is started.
 #
@@ -142,6 +143,7 @@ class postgresql::globals (
   $postgis_version                                 = undef,
   $repo_proxy                                      = undef,
   $repo_baseurl                                    = undef,
+  $yum_repo_commonurl                              = undef,
 
   $needs_initdb                                    = undef,
 
@@ -265,9 +267,10 @@ class postgresql::globals (
   # Setup of the repo only makes sense globally, so we are doing this here.
   if($manage_package_repo) {
     class { 'postgresql::repo':
-      version => $globals_version,
-      proxy   => $repo_proxy,
-      baseurl => $repo_baseurl,
+      version   => $globals_version,
+      proxy     => $repo_proxy,
+      baseurl   => $repo_baseurl,
+      commonurl => $yum_repo_commonurl,
     }
   }
 
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -3,6 +3,7 @@ class postgresql::repo (
   $version = undef,
   $proxy = undef,
   $baseurl = undef,
+  $commonurl = undef,
 ) {
   case $facts['os']['family'] {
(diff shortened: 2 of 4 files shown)


sanfrancrisko (commented):
Thanks for the enhancement @jorhett - LGTM. Unfortunately, at the moment there is an issue with our tests on the CentOS 8 containers which we're trying to resolve at the moment. Once we're operational again, I'll re-kick the tests and merge.

igalic:
I'm finding the placement or the naming a bit odd, given that this only applies to RHEL systems

jorhett:
> I'm finding the placement or the naming a bit odd, given that this only applies to RHEL systems

Yes, `_commonurl` seems odd but changing `_baseurl` would be a breaking change so I went with the minimal change. If you want a complete fix we should set baseurl to the repo top-level location, and then added common and pgdg as suffixes to that path. I'm happy to submit it, but it will be a breaking change :(

Even better alternative is for us to install the repos RPM package and allow them to adjust their repos as they see fit 👍 

I'm happy to go whatever way you want, but doing nothing is not an option since existing uses of this module do not work now that binaries have been moved out of the single repo https://www.postgresql.org/about/news/2027/

sanfrancrisko:
@jorhett , @igalic One possible suggestion might be to rename the `$repo_commonurl` parameter to `$yumrepo_commonurl` (or `$yum_repo_commonurl`) given this exclusively pertains to the PG YUM repos? We could also update the parameter description so it's a bit more obvious in the `REFERENCE.md`:
> Sets the url for the PostgreSQL common repository. Useful if you host your own mirror of the repository. Applies to RHEL / Fedora systems only.

...although hopefully the parameter name makes it obvious enough.

Also, @jorhett - whenever you get a chance could you rebase this branch with `puppetlabs:main` - we excluded CentOS 8 test execution temporarily whilst we look in to the issue. We have alternative methods to test CentOS 8 internally, so we're not losing coverage. This seems quite a specific case to that particular Docker container.

jorhett:
I am amazed that an officially supported Puppetlabs repo is still broken exactly 1 year after this change was made by the postgresql upstream maintainers.

I brought this issue to you, and even provided a patch. And this fix was held up because you wanted the variable renamed?

I'm not trying to be rude, but how was this a good use of your time and mine? You can rename a variable with a few keystrokes. Bouncing it back to me where it sat for 7 months because I've been too busy on other stuff... while your module is broken. This wasn't lack of tests. This wasn't a problem with the implementation. This was a variable name. And this module has been broken for a full year now... 

Anyway, it's really not clear what part you wanted renamed so I made a guess and implemented the change I think you want. Take a look now. And if you think another variable should be renamed, how about you just do that?

sanfrancrisko:
@jorhett Apologies for this languishing for so long. This fell through the cracks and shouldn't have been held up for as long as it was. Given there were other parties interested in the PR, and parameter naming can sometimes be surprisingly contentious, the intention was to leave this for a week or two max to ensure no other comments about naming landed in, then just rename and merge if nobody came back and it was still open.

For my part, apologies I didn't get returning to this sooner than I did - similar to yourself, many other things were on my plate. Regardless, the IAC Team now triage open PRs on Mondays, so we really shouldn't have gotten in to this position. I have a feeling how this blind spot emerged in our process and I'm going to chat with the team this week about it so we don't hit this again.

Thanks again for the contribution and apologies for the delay in getting it landed.

sanfrancrisko (approved):

sanfrancrisko:
@jorhett This is now in `v7.1.0` of the module which is live on the Forge now

jorhett:
Thank you. Numerous teams are happy to stop running local forks 👍 

> I have a feeling how this blind spot emerged in our process 

Of course that never happens to me 🤣 so I ain't gonna throw any more stones from my glass house 😆

PR #1195, opened 2020-09-24
add hostgssenc type to pg_hba rules

osijan:
Postgres 12 supports hostgss encryption. Add hostgssenc to $type to allow this hba rule type to pass the assert_type function.


sanfrancrisko (approved):

sanfrancrisko:
Thanks for the enhancement @osijan

PR #1199, opened 2020-10-19
Manage postgresql_conf_path file permissions

ekohl:
This adds a parameter (defaulting to true) to manage the postgresql_conf_path file resource. The actual content isn't managed because that's what postgresql::server::config_entry is for. However, managing the file ensures the correct permissions.

It also provides a workaround for [PUP-10548](https://tickets.puppetlabs.com/browse/PUP-10548) when using Red Hat SCL for packages when the correct SELinux file context isn't present.  Without managing the file, an admin will need to either make sure the package is present before running Puppet, manage the file via another module or manually set the file context after Puppet ran. With this workaround, Puppet will converge on the second run and actually start PostgreSQL.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -1,34 +1,35 @@
 # @api private
 class postgresql::params inherits postgresql::globals {
-  $version                    = $postgresql::globals::globals_version
-  $postgis_version            = $postgresql::globals::globals_postgis_version
-  $listen_addresses           = undef
-  $port                       = 5432
-  $log_line_prefix            = undef
-  $ip_mask_deny_postgres_user = '0.0.0.0/0'
-  $ip_mask_allow_all_users    = '127.0.0.1/32'
-  $ipv4acls                   = []
-  $ipv6acls                   = []
-  $encoding                   = $postgresql::globals::encoding
-  $locale                     = $postgresql::globals::locale
-  $data_checksums             = $postgresql::globals::data_checksums
-  $timezone                   = $postgresql::globals::timezone
-  $service_ensure             = 'running'
-  $service_enable             = true
-  $service_manage             = true
-  $service_restart_on_change  = true
-  $service_provider           = $postgresql::globals::service_provider
-  $manage_pg_hba_conf         = pick($manage_pg_hba_conf, true)
-  $manage_pg_ident_conf       = pick($manage_pg_ident_conf, true)
-  $manage_recovery_conf       = pick($manage_recovery_conf, false)
-  $manage_selinux             = pick($manage_selinux, false)
-  $package_ensure             = 'present'
-  $module_workdir             = pick($module_workdir,'/tmp')
-  $password_encryption        = undef
-  $extra_systemd_config       = ''
-  $manage_datadir             = true
-  $manage_logdir              = true
-  $manage_xlogdir             = true
+  $version                      = $postgresql::globals::globals_version
+  $postgis_version              = $postgresql::globals::globals_postgis_version
+  $listen_addresses             = undef
+  $port                         = 5432
+  $log_line_prefix              = undef
+  $ip_mask_deny_postgres_user   = '0.0.0.0/0'
+  $ip_mask_allow_all_users      = '127.0.0.1/32'
+  $ipv4acls                     = []
+  $ipv6acls                     = []
+  $encoding                     = $postgresql::globals::encoding
+  $locale                       = $postgresql::globals::locale
+  $data_checksums               = $postgresql::globals::data_checksums
+  $timezone                     = $postgresql::globals::timezone
+  $service_ensure               = 'running'
+  $service_enable               = true
+  $service_manage               = true
+  $service_restart_on_change    = true
+  $service_provider             = $postgresql::globals::service_provider
+  $manage_pg_hba_conf           = pick($manage_pg_hba_conf, true)
+  $manage_pg_ident_conf         = pick($manage_pg_ident_conf, true)
+  $manage_recovery_conf         = pick($manage_recovery_conf, false)
+  $manage_postgresql_conf_perms = pick($manage_postgresql_conf_perms, true)
+  $manage_selinux               = pick($manage_selinux, false)
+  $package_ensure               = 'present'
+  $module_workdir               = pick($module_workdir,'/tmp')
+  $password_encryption          = undef
+  $extra_systemd_config         = ''
+  $manage_datadir               = true
+  $manage_logdir                = true
+  $manage_xlogdir               = true
 
   # Amazon Linux's OS Family is 'Linux', operating system 'Amazon'.
   case $facts['os']['family'] {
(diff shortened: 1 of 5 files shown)


ekohl:
There are a ton of whitespace changes here, but the actual change itself is much smaller.

sheenaajay:
Thanks alot @ekohl for your contribution. Acceptance tests are failing with the following error `     Error: Puppet::Util::FileType::FileTypeFlat could not write /tmp/postgresql.conf: Permission denied @ rb_sysopen - /tmp/postgresql.conf`
Please let us know if you need more information. Thank you.

ekohl:
Looks like the default file mode on Debian is 644 so I'll need to introduce a parameter for the file permission.

ekohl:
I updated it and added the parameter. I only know its value for Red Hat (taken from CentOS 7 & 8) and Debian (taken from 8, 9 & 10, assuming Ubuntu is the same).

ekohl:
All builds failed with
```
.rake aborted!

NoMethodError: undefined method `trace' for #<Logging::Logger:0x000000000444b570>

Did you mean?  trap

/home/travis/build/puppetlabs/puppetlabs-postgresql/vendor/bundle/ruby/2.5.0/gems/bolt-2.24.1/lib/bolt/pal/logging.rb:15:in `handle'
```

This looks like a bolt bug rather than something I did.

sheenaajay:
@ekohl Thank you for the quick response. Looks like its using older version of bolt 2.24.1
https://travis-ci.org/github/puppetlabs/puppetlabs-postgresql/jobs/741465093#L498
Let me clear the cache and kick for a rerun. Thank you

ekohl:
Haven't been able to quite figure out why this failed. Rebased it to get the output cleanups in to make the output a bit easier to read.

ekohl:
So the problem on Debian based systems is with `/tmp`:
```console
# ls -ld /tmp /tmp/postgresql.conf 
drwxrwxrwt. 2 root     root     4096 Nov 10 14:09 /tmp
-rw-------. 1 postgres postgres    0 Nov 10 14:08 /tmp/postgresql.conf
# echo test >> /tmp/postgresql.conf 
bash: /tmp/postgresql.conf: Permission denied
# id
uid=0(root) gid=0(root) groups=0(root)
```

This is probably because it's mounted as sticky. Changing the path to `/etc` should help.

ekohl:
This is really hard to debug because litmus output is really not telling me a lot. I'd guess something can't read the config file, but I can't quite figure it out.

sheenaajay:
@ekohl Thanks alot for the quick changes. Was looking at the errors. Its failing with the folllowing error ` 1) postgresql::server::grant_role: grants a role to a user/superuser
     On host `localhost:2224'
     Failure/Error: LitmusHelper.instance.run_shell("cd /tmp; su #{shellescape(user)} -c #{shellescape(psql)}", acceptable_exit_codes: exit_codes, &block)
     RuntimeError:
       shell failed
       `cd /tmp; su psql_grant_role_tester -c psql\ --command\=\"SELECT\ 1\ WHERE\ pg_has_role\(\'psql_grant_role_tester\',\ \'test_group\',\ \'MEMBER\'\)\ \=\ true\"\ grant_role_test`
       ======
       [{"target"=>"localhost:2224", "action"=>"command", "object"=>"cd /tmp; su psql_grant_role_tester -c psql\\ --command\\=\\\"SELECT\\ 1\\ WHERE\\ pg_has_role\\(\\'psql_grant_role_tester\\',\\ \\'test_group\\',\\ \\'MEMBER\\'\\)\\ \\=\\ true\\\"\\ grant_role_test", "status"=>"failure", "value"=>{"stdout"=>"", "stderr"=>"Error: Invalid data directory\n", "exit_code"=>1, "_error"=>{"kind"=>"puppetlabs.tasks/command-error", "issue_code"=>"COMMAND_ERROR", "msg"=>"The command failed with exit code 1", "details"=>{"exit_code"=>1}}}}]
       
     # ./vendor/bundle/ruby/2.5.0/gems/puppet_litmus-0.18.4/lib/puppet_litmus/puppet_helpers.rb:173:in `block in run_shell'
     # ./vendor/bundle/ruby/2.5.0/gems/honeycomb-beeline-2.3.0/lib/honeycomb/client.rb:70:in `start_span'
     # ./vendor/bundle/ruby/2.5.0/gems/puppet_litmus-0.18.4/lib/puppet_litmus/puppet_helpers.rb:157:in `run_shell'
     # ./spec/spec_helper_acceptance_local.rb:47:in `psql'
     # ./spec/acceptance/server/grant_role_spec.rb:164:in `block (2 levels) in <top (required)>'
`
/spec/acceptance/server/grant_role_spec.rb:164

https://github.com/ekohl/puppetlabs-postgresql/blob/manage-postgresql-config/spec/acceptance/server/grant_role_spec.rb#L164

` ## Check that the role was granted to the user
      psql('--command="SELECT 1 WHERE pg_has_role(\'psql_grant_role_tester\', \'test_group\', \'MEMBER\') = true" grant_role_test', 'psql_grant_role_tester') do |r|`

https://github.com/ekohl/puppetlabs-postgresql/blob/manage-postgresql-config/spec/spec_helper_acceptance_local.rb#L47

`def psql(psql_cmd, user = 'postgres', exit_codes = [0, 1], &block)
  psql = "psql #{psql_cmd}"
  LitmusHelper.instance.run_shell("cd /tmp; su #{shellescape(user)} -c #{shellescape(psql)}", acceptable_exit_codes: exit_codes, &block)
end`

By default it executes psql command from /tmp directory. We can modify the psql function to accepts one parameter to  make the config path configurable.

ekohl:
There is a known issue that you often end up executing it from `/root` which can't be read. I can try to change it to `/` which should also not produce an error but I don't expect that to matter.

manifests/server/config.pp
```diff
@@ -100,6 +102,15 @@
     }
   }
 
+  if $manage_postgresql_conf_perms {
+    file { $postgresql_conf_path:
+      ensure => file,
+      owner  => $user,
+      group  => $group,
+      mode   => '0600',
```
DavidS, line 110: shouldn't this be using `postgresql_conf_mode`?

ekohl: I think you may have found the thing I was struggling to find. Sometimes you're blind to your own code.

ekohl:
Thanks to @DavidS it's green now.

sheenaajay:
wow superb. Thank you @DavidS and @ekohl .

manifests/params.pp
```diff
@@ -78,6 +79,7 @@
           default  => pick($datadir, "/var/lib/pgsql/${version}/data"),
         }
         $confdir                = pick($confdir, $datadir)
+        $postgresql_conf_mode   = pick($postgresql_conf_mode, '0600')
```
sheenaajay, line 82: Could you please clarify on  different permission modes on Amazon and Debian. Thank you

ekohl: I don't know what Amazon should be since I have no access to it, but on a fresh CentOS 7:
```
-rw-------. 1 postgres postgres 19845 Nov 16 16:10 /var/lib/pgsql/data/postgresql.conf
```

Then on a fresh Debian 10:
```
-rw-r--r-- 1 postgres postgres 24160 Nov 16 16:37 /etc/postgresql/11/main/postgresql.conf
```

I'm not sure how many versions I checked when I originally wrote the change, but I recall that Debian and Ubuntu versions were the same.

sheenaajay:
Thank you @ekohl  for your contribution.

PR #1201, opened 2020-10-27
Change - Use systemd drop-in directory for unit overrides

blackknight36:
The .include directive has been deprecated and support for it may be removed in a future systemd release.  This fixes warnings issued by systemd which are shown as follows.

```
systemd[1]: /etc/systemd/system/postgresql-9.6.service:1: .include directives are deprecated, and support for them will be removed in a future version of systemd. Please use drop-in files instead.
```

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -216,7 +216,12 @@ class postgresql::server::config {
       # - $datadir
       # - @extra_systemd_config
 
-    if (versioncmp($facts['puppetversion'], '5.0.0') <= 0) {
+    if (versioncmp($facts['puppetversion'], '6.1.0') < 0) {
+      exec { 'restart-systemd':
+        command     => 'systemctl daemon-reload',
+        refreshonly => true,
+        path        => '/bin:/usr/bin:/usr/local/bin',
+      }
       $systemd_notify = [Exec['restart-systemd'], Class['postgresql::server::service']]
     }
     else {
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -205,48 +205,50 @@ class postgresql::server::config {
     }
   }
 
-  if $facts['os']['family'] == 'RedHat' {
-    if $facts['os']['release']['full'] =~ /^7|^8/ or $facts['os']['name'] == 'Fedora' {
+  # RHEL 7 and 8 both support drop-in files for systemd units.  The old include directive is deprecated and may be removed in future systemd releases.
+  # Gentoo also supports drop-in files.
+  if $facts['os']['family'] in ['RedHat', 'Gentoo'] and $facts['service_provider'] == 'systemd' {
       # Template uses:
-      # - $::operatingsystem
+      # - $facts['os']['name']
+      # - $facts['os']['release']['major']
       # - $service_name
       # - $port
       # - $datadir
-      file { 'systemd-override':
-        ensure  => file,
-        path    => "/etc/systemd/system/${service_name}.service",
-        owner   => root,
-        group   => root,
-        content => template('postgresql/systemd-override.erb'),
-        notify  => [Exec['restart-systemd'], Class['postgresql::server::service']],
-        before  => Class['postgresql::server::reload'],
-      }
-      exec { 'restart-systemd':
-        command     => 'systemctl daemon-reload',
-        refreshonly => true,
-        path        => '/bin:/usr/bin:/usr/local/bin',
-      }
+      # - @extra_systemd_config
+
+    if (versioncmp($facts['puppetversion'], '5.0.0') <= 0) {
+      $systemd_notify = [Exec['restart-systemd'], Class['postgresql::server::service']]
     }
-  }
-  elsif $facts['os']['family'] == 'Gentoo' {
-    # Template uses:
-    # - $::operatingsystem
-    # - $service_name
-    # - $port
-    # - $datadir
-    file { 'systemd-override':
-      ensure  => file,
-      path    => "/etc/systemd/system/${service_name}.service",
-      owner   => root,
-      group   => root,
-      content => template('postgresql/systemd-override.erb'),
-      notify  => [Exec['restart-systemd'], Class['postgresql::server::service']],
-      before  => Class['postgresql::server::reload'],
+    else {
+      $systemd_notify = Class['postgresql::server::service']
     }
-    exec { 'restart-systemd':
-      command     => 'systemctl daemon-reload',
-      refreshonly => true,
-      path        => '/bin:/usr/bin:/usr/local/bin',
+
+    file {
+      default:
+          ensure => file,
+          owner  => root,
+          group  => root,
+          notify => $systemd_notify,
+          before => Class['postgresql::server::reload'],
+
+      ;
+
+      'systemd-conf-dir':
+          ensure => directory,
+          path   => "/etc/systemd/system/${service_name}.service.d",
+      ;
+
+      'systemd-override':
+          path    => "/etc/systemd/system/${service_name}.service.d/${service_name}.conf",
+          content => template('postgresql/systemd-override.erb'),
+          require => File['systemd-conf-dir'],
+      ;
+
+      # Remove old unit file to avoid conflicts
+      'old-systemd-override':
+          ensure => absent,
+          path   => "/etc/systemd/system/${service_name}.service",
+      ;
     }
   }
 }
--- a/spec/unit/classes/server/config_spec.rb
+++ b/spec/unit/classes/server/config_spec.rb
@@ -316,7 +316,6 @@ describe 'postgresql::server::config', type: :class do
           },
           selinux: { 'enabled' => true },
         },
-        concat_basedir: tmpfilename('server'),
         kernel: 'Linux',
         id: 'root',
         path: '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
--- a/spec/unit/classes/server/config_spec.rb
+++ b/spec/unit/classes/server/config_spec.rb
@@ -354,7 +354,6 @@ describe 'postgresql::server::config', type: :class do
           },
           selinux: { 'enabled' => true },
         },
-        concat_basedir: tmpfilename('server'),
         kernel: 'Linux',
         id: 'root',
         path: '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
--- a/spec/unit/classes/server/config_spec.rb
+++ b/spec/unit/classes/server/config_spec.rb
@@ -18,12 +18,14 @@ describe 'postgresql::server::config', type: :class do
           'hardware'     => 'x86_64',
           'name'         => 'CentOS',
           'release'      => {
-            'full'  => '7.6.1810',
+            'full'  => '7.9.2009',
             'major' => '7',
-            'minor' => '6',
+            'minor' => '9',
           },
           selinux: { 'enabled' => true },
         },
+        operatingsystemrelease: '7.9.2009',
+        service_provider: 'systemd',
       }
     end
 
@@ -36,15 +38,55 @@ describe 'postgresql::server::config', type: :class do
         .that_requires('Package[policycoreutils-python-utils]')
     end
 
-    it 'has the correct systemd-override file' do
-      is_expected.to contain_file('systemd-override').with(
-        ensure: 'file', path: '/etc/systemd/system/postgresql.service',
-        owner: 'root', group: 'root'
-      )
+    it 'removes the old systemd-override file' do
+      is_expected.to contain_file('old-systemd-override')
+        .with(ensure: 'absent', path: '/etc/systemd/system/postgresql.service')
     end
-    it 'has the correct systemd-override file #content' do
+
+    it 'has the correct systemd-override drop file' do
+      is_expected.to contain_file('systemd-override').with(
+        ensure: 'file', path: '/etc/systemd/system/postgresql.service.d/postgresql.conf',
+        owner: 'root', group: 'root'
+      ) .that_requires('File[systemd-conf-dir]')
+    end
+
+    it 'has the correct systemd-override file #regex' do
       is_expected.to contain_file('systemd-override') \
-        .with_content(%r{.include \/usr\/lib\/systemd\/system\/postgresql.service})
+        .with_content(%r{(?!^.include)})
+    end
+
+    context 'RHEL 7 host with Puppet 5' do
+      let(:facts) do
+        {
+          kernel: 'Linux',
+          id: 'root',
+          path: '/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin',
+          selinux: true,
+          os: {
+            'architecture' => 'x86_64',
+            'family'       => 'RedHat',
+            'hardware'     => 'x86_64',
+            'name'         => 'CentOS',
+            'release'      => {
+              'full'  => '7.9.2009',
+              'major' => '7',
+              'minor' => '9',
+            },
+            selinux: { 'enabled' => true },
+          },
+          operatingsystemrelease: '7.9.2009',
+          puppetversion: '5.0.1',
+          service_provider: 'systemd',
+        }
+      end
+
+      it 'has systemctl restart command' do
+        is_expected.to contain_exec('restart-systemd').with(
+          command: 'systemctl daemon-reload',
+          refreshonly: true,
+          path: '/bin:/usr/bin:/usr/local/bin',
+        )
+      end
     end
 
     describe 'with manage_package_repo => true and a version' do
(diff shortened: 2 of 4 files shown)


daianamezdrea:
Hi @blackknight36, thank you for submitting this PR, can you have a look at Travis failures which are coming from this?
```
manifests/server/config.pp - WARNING: indentation of => is not properly aligned (expected in column 16, but found it in column 17) on line 211
manifests/server/config.pp - WARNING: indentation of => is not properly aligned (expected in column 16, but found it in column 17) on line 212
manifests/server/config.pp - WARNING: indentation of => is not properly aligned (expected in column 16, but found it in column 17) on line 213
manifests/server/config.pp - WARNING: indentation of => is not properly aligned (expected in column 16, but found it in column 17) on line 214
manifests/server/config.pp - WARNING: indentation of => is not properly aligned (expected in column 17, but found it in column 16) on line 219 
```

blackknight36:
> Hi @blackknight36, thank you for submitting this PR, can you have a look at Travis failures which are coming from this?
> 
> ```
> manifests/server/config.pp - WARNING: indentation of => is not properly aligned (expected in column 16, but found it in column 17) on line 211
> manifests/server/config.pp - WARNING: indentation of => is not properly aligned (expected in column 16, but found it in column 17) on line 212
> manifests/server/config.pp - WARNING: indentation of => is not properly aligned (expected in column 16, but found it in column 17) on line 213
> manifests/server/config.pp - WARNING: indentation of => is not properly aligned (expected in column 16, but found it in column 17) on line 214
> manifests/server/config.pp - WARNING: indentation of => is not properly aligned (expected in column 17, but found it in column 16) on line 219 
> ```

Yes, I should be able to fix it tomorrow.

manifests/server/config.pp
```diff
@@ -201,15 +201,26 @@
...
       file { 'systemd-override':
         ensure  => file,
-        path    => "/etc/systemd/system/${service_name}.service",
+        path    => "/etc/systemd/system/${service_name}.service.d/${service_name}.conf",
         owner   => root,
         group   => root,
         content => template('postgresql/systemd-override.erb'),
         notify  => [Exec['restart-systemd'], Class['postgresql::server::service']],
         before  => Class['postgresql::server::reload'],
+        require => File['systemd-conf-dir'],
```
ekohl, line 221: Isn't this autorequired since Puppet 2.7?

templates/systemd-override.erb
```diff
@@ -1,7 +1,8 @@
 <%- if scope.lookupvar('::osfamily') == 'Gentoo' -%>
 .include /usr/lib64/systemd/system/<%= @service_name %>.service
-<%- elsif scope.lookupvar('::operatingsystem') == 'Fedora' -%>
+<%- elsif scope.lookupvar('::operatingsystem') == 'Fedora' and scope.lookupvar('::operatingsystem') <= '31' -%>
 .include /lib/systemd/system/<%= @service_name %>.service
+<%- elsif scope.lookupvar('::operatingsystem') == 'Fedora' and scope.lookupvar('::operatingsystem') >= '32' -%>
```
ekohl, line 5: I think drop ins should work on all supported variants of the Red Hat OS family.
```suggestion
<%- elsif @facts['os']['family'] == 'Red Hat' -%>
```

In fact, I'm not sure if there's any systemd version that doesn't support this.

manifests/server/config.pp
```diff
@@ -201,15 +201,26 @@
...
+        path   => "/etc/systemd/system/${service_name}.service.d",
+        owner   => root,
+        group   => root,
+        notify  => [ Exec['restart-systemd'], Class['postgresql::server::service'] ],
+        before  => Class['postgresql::server::reload'],
+      }
+
       file { 'systemd-override':
         ensure  => file,
-        path    => "/etc/systemd/system/${service_name}.service",
```
ekohl, line 206: For migrations it's a good idea to ensure the old location is absent. Otherwise an incorrect service could be loaded.

ekohl (changes requested):

manifests/server/config.pp
blackknight36: I'm not certain about that.  I've had issues in the past with puppet trying to create files in a directory that doesn't exist which is why I added the requirement here.  If you want to leave it off that's fine though.

templates/systemd-override.erb
blackknight36: I didn't want to break servers running older Fedora releases and anything older than F32 is EOL now which is why I limited this change to more modern releases.  I can run some tests on our Fedora 20 nodes to see if drop-in directories are supported but I didn't see anything in the documentation mentioning that feature.  

Another way to handle this would be to create a "systemd_version" fact which may work better than limiting options to just the RedHat OS family.

manifests/server/config.pp
blackknight36: You are correct.  I can add a resource to ensure that the old location is removed.

templates/systemd-override.erb
ekohl: It's supported on EL 7 which is based on Fedora 18ish IIRC. I wouldn't be surprised if it worked on even the ancient Fedora 17 that's listed in globals.pp.

manifests/server/config.pp
ekohl: In spec tests you can verify it:
```ruby
it { is_expected.to contain_file('/etc/systemd/postgresql-server.d/override.conf').that_requires('/etc/systemd/postgresql-server.d') }
```

blackknight36: I just did a quick test and it appears that there is no autorequire for the parent directory.  The unit test fails after adding a test similar to above.

templates/systemd-override.erb
blackknight36: The oldest Fedora server I have available right now is running Fedora 20.  The man page for systemd.unit makes no mention of support for drop-in directories.  The man page on CentOS 7 *does* mention them which makes sense since EL7 runs systemd version 219 vs. 208 on Fedora 20.

Knowing this I should be able to make these changes work with Fedora 20 and up.

manifests/server/config.pp
blackknight36: I've added a resource to ensure the old file location is deleted.

maage:
I see this issue at CentOS 8 too:

Wed 2020-11-11 11:50:46 UTC hostname systemd[1]: /etc/systemd/system/postgresql-13.service:1: .include directives are deprecated, and support for them will be removed in a future version of systemd. Please use drop-in files instead.

```
$ rpm -q systemd
systemd-239-31.el8_2.2.x86_64
```

```
$ facter os
{
  architecture => "x86_64",
  family => "RedHat",
  hardware => "x86_64",
  name => "CentOS",
  release => {
    full => "8.2.2004",
    major => "8",
    minor => "2"
  },
  selinux => {
    config_mode => "enforcing",
    config_policy => "targeted",
    current_mode => "enforcing",
    enabled => true,
    enforced => true,
    policy_version => "31"
  }
}
```

blackknight36:
> I see this issue at CentOS 8 too:
> 
> Wed 2020-11-11 11:50:46 UTC hostname systemd[1]: /etc/systemd/system/postgresql-13.service:1: .include directives are deprecated, and support for them will be removed in a future version of systemd. Please use drop-in files instead.
> 
> ```
> $ rpm -q systemd
> systemd-239-31.el8_2.2.x86_64

I can update this PR to include CentOS 8 as well.

blackknight36:
Another thing I noticed is that the postgresql::server::config class does not appear to address RHEL 5 or 6 although these releases are supported according to the metadata file.  Should there be checks set up for these releases as well?

maage:
RHEL/CentOS 5/6 does not use systemd, they use old upstart, service - command and /etc/init.d -directory setup.

Regarding to upstream supports:

https://access.redhat.com/support/policy/updates/errata#Life_Cycle_Dates

RH extended support is going to end at the end of 2020-11. And RHEL6 maintenance support is going to end also 2020-11, but extended support is still possible until 2024-06.

metadata.json also does not mention Fedora (currently 31, 32, 33).

I guess any supported OS version should be supported with testing and no testing is necessary for unsupported OS combos.

sheenaajay:
Thanks alot @blackknight36 for the changes. Could you please fix the travis failures.
Thank you @ekohl and @maage for the valuable comments.

blackknight36:
> Thanks alot @blackknight36 for the changes. Could you please fix the travis failures.
> Thank you @ekohl and @maage for the valuable comments.

The Travis failures have been fixed.  Please let me know if you need any other changes.

sheenaajay (approved):

templates/systemd-override.erb
sheenaajay: @blackknight36 we have to include conditions for Redhat 8 since the deprecated warnings are coming there too.Will run the tests on all OSes and see if its break anything. Thanks alot for your contribution.

blackknight36: Thanks, I forgot to add that.  I'll fix the template and push again.

blackknight36: Ideally the include statement should be removed on *all* OSes, not just Redhat.  I'm sure that Gentoo also includes a newer version of systemd which would be affected in the future as well.  I don't have a Gentoo server to test on however.

ekohl: I think worries about drop ins are really not needed. This is literally the only reason I know that includes are a thing. On my Gentoo boxes I use drop ins regularly. I think that on all distros you can rely on them.

templates/systemd-override.erb
```diff
@@ -1,8 +1,6 @@
-<%- if scope.lookupvar('::osfamily') == 'Gentoo' -%>
+<%- if @os["name"] == 'Gentoo' -%>
 .include /usr/lib64/systemd/system/<%= @service_name %>.service
-<%- elsif scope.lookupvar('::operatingsystem') == 'Fedora' -%>
-.include /lib/systemd/system/<%= @service_name %>.service
-<% else -%>
+<%- elsif (@os["family"] == 'RedHat' and @os["release"]["major"] <= '7') or (@os["name"] == 'Fedora' and @os["release"]["major"] <= '31') -%>
```
ekohl, line 3: EL7 also supports drop ins and I'm pretty sure all Fedoras too.

templates/systemd-override.erb
```diff
@@ -1,8 +1,6 @@
-<%- if scope.lookupvar('::osfamily') == 'Gentoo' -%>
+<%- if @os["name"] == 'Gentoo' -%>
```
ekohl, line 1: Gentoo supports drop ins just fine.

manifests/server/config.pp
```diff
@@ -195,7 +195,48 @@
...
+            ensure => absent,
+            path   => "/etc/systemd/system/${service_name}.service",
+        ;
+      }
+
+      exec { 'restart-systemd':
+        command     => 'systemctl daemon-reload',
+        refreshonly => true,
+        path        => '/bin:/usr/bin:/usr/local/bin',
+      }
```
ekohl, line 236: This is only needed on Puppet < 6.1.0. Newer versions do this when needed.

ekohl (commented):
Not a module maintainer, but I'd strongly urge you to drop the old include behavior.

templates/systemd-override.erb
blackknight36: Looks like you're correct however I'm not sure when EL7 added support for drop-in dirs. EL 7.9 includes a newer version of systemd which does support drop-in dirs according to the systemd.unit man page.  I can the PR to work with this.

manifests/server/config.pp
blackknight36: Newer versions of puppet or systemd?

templates/systemd-override.erb
blackknight36: Thanks, I will update the template.

templates/systemd-override.erb
ekohl: I've been using drop ins on EL7 for ages, also before EL 7.9. Never knew otherwise.

manifests/server/config.pp
ekohl: Puppet 6.1.0. See https://tickets.puppetlabs.com/browse/PUP-3483 for details.

templates/systemd-override.erb
blackknight36: I've updated the manifest to create drop-in files on RedHat 7 and 8 nodes.  Gentoo will also now do the same.

manifests/server/config.pp
```diff
@@ -222,21 +226,15 @@
...
 
-        # Remove old unit file to avoid service failures
+        # Remove old unit file to avoid conflicts
         'old-systemd-override':
             ensure => absent,
             path   => "/etc/systemd/system/${service_name}.service",
         ;
       }
-
-      exec { 'restart-systemd':
```
ekohl: To be clear, this should be kept in since the module still supports Puppet 5, but you can guard it on a version fact.

manifests/server/config.pp
blackknight36: The exec command has been removed.

manifests/server/config.pp
```diff
@@ -222,21 +226,15 @@
...
-        command     => 'systemctl daemon-reload',
-        refreshonly => true,
-        path        => '/bin:/usr/bin:/usr/local/bin',
-      }
     }
 
-    if $facts['os']['release']['major'] == '7' or ($facts['os']['name'] == 'Fedora' and $facts['os']['release']['major'] <= '31') {
+    if $facts['os']['name'] == 'Fedora' and $facts['os']['release']['major'] <= '31' {
```
ekohl, line 237: Old Fedora also works well with drop ins. I think it's safe to drop this part altogether.

spec/unit/classes/server/config_spec.rb
```diff
@@ -61,13 +68,14 @@ class { 'postgresql::server': }
...
-          ensure: 'file', path: '/etc/systemd/system/postgresql-9.4.service',
+          ensure: 'file', path: '/etc/systemd/system/postgresql-9.4.service.d/postgresql-9.4.conf',
           owner: 'root', group: 'root'
         )
       end
+
       it 'has the correct systemd-override file #regex' do
         is_expected.to contain_file('systemd-override') \
-          .with_content(%r{.include \/usr\/lib\/systemd\/system\/postgresql-9.4.service})
+          .with_content(%r{(?!^.include)})
```
ekohl, line 78:
```suggestion
          .without_content(%r{\.include})
```

manifests/server/config.pp
```diff
@@ -194,19 +194,23 @@
     }
   }
 
-  if $facts['os']['family'] == 'RedHat' {
-    if $facts['os']['release']['major'] == '8' or ($facts['os']['name'] == 'Fedora' and $facts['os']['release']['major'] >= '32') {
+  # RHEL 7 and 8 both support drop-in files for systemd units.  The old include directive is deprecated and may be removed in future systemd releases.
+  # Gentoo also supports drop-in files.
+  if ($facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] =~ /7|8/) or ($facts['os']['name'] == 'Fedora' and $facts['os']['release']['major'] >= '32') or $facts['os']['family'] == 'Gentoo' {
```
ekohl, line 199: I'd also use drop ins on Fedora and simplify it to:
```suggestion
  if $facts['os']['family'] in ['RedHat', 'Gentoo'] and $facts['service_provider'] == 'systemd' {
```

This uses the [service_provider](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/facter/service_provider.rb) fact from stdlib to detect the service provider.

manifests/server/config.pp
blackknight36: I'll change this to use the $puppetversion fact.

manifests/server/config.pp
blackknight36: That makes sense and TBH do we even *care* about old Fedora releases?  Anything older than F32 is now EOL and anybody running Fedora in production would be advised to upgrade.

ekohl: Exactly. If EL7 supports it, that means ancient Fedora supports it. Fedora isn't even listed in metadata.json so technically there is no guarantee at all that it works.

blackknight36: I've updated the conditional OS checks using the changes suggested and also removed unit tests for ancient Fedora releases.

manifests/server/config.pp
```diff
@@ -205,48 +205,57 @@
...
-        before  => Class['postgresql::server::reload'],
-      }
-      exec { 'restart-systemd':
-        command     => 'systemctl daemon-reload',
-        refreshonly => true,
-        path        => '/bin:/usr/bin:/usr/local/bin',
-      }
+      # - @extra_systemd_config
+
+    if (versioncmp($facts['puppetversion'], '5.0.0') <= 0) {
```
ekohl, line 220: I think you still need to declare the systemd restart. It's also on all versions older than 6.1.0
```suggestion
    if (versioncmp($facts['puppetversion'], '6.1.0') < 0) {
      exec { 'restart-systemd':
        command     => 'systemctl daemon-reload',
        refreshonly => true,
        path        => '/bin:/usr/bin:/usr/local/bin',
      }
```

spec/unit/classes/server/config_spec.rb
```diff
@@ -36,15 +39,112 @@
...
+        is_expected.to contain_file('systemd-override') \
+          .without_content(%r{\.include})
+      end
+    end
+  end
+
+  describe 'on Redhat 8' do
+    let(:facts) do
+      {
+        concat_basedir: tmpfilename('server'),
```
ekohl, line 101: [Concat basedir is dead](https://github.com/puppetlabs/puppetlabs-postgresql/commit/3fba56bd8ab36e4862f9ea7b2e9cc69206271531)
```suggestion
```

spec/unit/classes/server/config_spec.rb
```diff
@@ -163,82 +276,99 @@ class { 'postgresql::server': }
...
+          name: 'CentOS',
+          release: {
+            'full'  => '7.9.2009',
+            'major' => '7',
+            'minor' => '9',
+          },
           selinux: { 'enabled' => true },
         },
-        osfamily: 'Gentoo',
+        concat_basedir: tmpfilename('server'),
```
ekohl, line 306: [concat_basedir is dead](https://github.com/puppetlabs/puppetlabs-postgresql/commit/3fba56bd8ab36e4862f9ea7b2e9cc69206271531)
```suggestion
```

spec/unit/classes/server/config_spec.rb
```diff
@@ -163,82 +276,99 @@ class { 'postgresql::server': }
...
-          release: { 'full' => '7.0' },
+          family: 'Gentoo',
+          name: 'Gentoo',
+          release: {
+            'full'  => 'unused',
+            'major' => 'unused',
+          },
           selinux: { 'enabled' => true },
         },
+        concat_basedir: tmpfilename('server'),
```
ekohl, line 344: Same
```suggestion
```

ekohl (changes requested):

sheenaajay:
Thank you @blackknight36 @ekohl for your contribution.

sheenaajay (approved):

vchepkov:
This change breaks existing postgres instance (pre 6.9.0). One need to issue systemctl daemon-reload first, after old unit file was deleted and only after that restart the newly configured service, otherwise service fails

blackknight36:
> This change breaks existing postgres instance (pre 6.9.0). One need to issue systemctl daemon-reload first, after old unit file was deleted and only after that restart the newly configured service, otherwise service fails

Thanks for the report.  I will see if there is a way to avoid this.

vchepkov:
I would if `issues` were enabled on the repository.
I see that a fix is in progress https://github.com/puppetlabs/puppetlabs-postgresql/pull/1230

PR #1209, opened 2020-11-19
(IAC-1189) - Fix for SLES 15 SP 1 and later

david22swan:
Current code only allows for SLES 15 SP0 while released versions currently go up to SP2. 
Have extended to potential SP9 to prevent the need to update again in the future as the final SP version is as of yet unknown.


david22swan:
Fix for GCP specifically

carabasdaniel (approved):

PR #1222, opened 2020-12-22
Support CentOS 8 Stream

ekohl:
CentOS 8 Stream identifies itself without a minor version. This modifies the default to handle this. For completeness, the OS facts:

```console
# facter os
{
  architecture => "x86_64",
  family => "RedHat",
  hardware => "x86_64",
  name => "CentOS",
  release => {
    full => "8",
    major => "8"
  },
  selinux => {
    enabled => false
  }
}
```

This is the first thing I found. I have not yet done a real test myself so for now this is a draft.

PR #1225, opened 2021-01-10
Clean up globals logic to support CentOS 8 stream

ekohl:
This is a code simplification for distributions that don't change the 
PostgreSQL version in a major release. As a bonus fact, it adds support for
CentOS 8 Stream which identifies itself without a minor version.

For completeness, the OS facts from CentOS 8 Stream:

```console
# facter os
{
  architecture => "x86_64",
  family => "RedHat",
  hardware => "x86_64",
  name => "CentOS",
  release => {
    full => "8",
    major => "8"
  },
  selinux => {
    enabled => false
  }
}
```

This is an alternative to https://github.com/puppetlabs/puppetlabs-postgresql/pull/1222.

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -204,10 +204,7 @@ class postgresql::globals (
       },
       default => undef,
     },
-    'Archlinux' => $facts['os']['name'] ? {
-      /Archlinux/ => '9.2',
-      default => '9.2',
-    },
+    'Archlinux' => '9.2',
     'Gentoo' => '9.5',
     'FreeBSD' => '93',
     'OpenBSD' => $facts['os']['release']['full'] ? {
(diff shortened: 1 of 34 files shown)


sheenaajay:
Thanks @ekohl for your contribution as always. 
Could you please take a look at unit tests which are failing with the second commit.

ekohl:
Looks like I need to do a bit more.

ekohl:
I went through all files that used the full version. Also went through tests. It's confusing that this module places the files in `spec/unit/defines` rather than `spec/defines` as is the convention.

ekohl:
:green_heart: now

sheenaajay (approved):

sheenaajay:
Thank you @ekohl for the fixes.

ekohl:
@sheenaajay anything holding back a merge? I hope to see this in a released version soon.

sheenaajay:
@ekohl Sorry for the delay. Next one on the list :-) . We have a release coming soon for a bug fix and can get this also included. Thank you.

PR #1226, opened 2021-01-13
Fix postgresql::sql task when password is not set

smortex:
When password is not set, `{ 'PGPASSWORD' => password } unless
password.nil?` evaluates to `nil`, setting `env_hash` to `nil`.

`nil` is then passed as the first argument of `Open3.capture3` which
result in an exception:

```
/opt/puppetlabs/puppet/lib/ruby/2.7.0/open3.rb:213:in `spawn': no implicit conversion of nil into String (TypeError)
	from /opt/puppetlabs/puppet/lib/ruby/2.7.0/open3.rb:213:in `popen_run'
	from /opt/puppetlabs/puppet/lib/ruby/2.7.0/open3.rb:101:in `popen3'
	from /opt/puppetlabs/puppet/lib/ruby/2.7.0/open3.rb:281:in `capture3'
	from /opt/puppetlabs/mcollective/tasks-spool/899e9e0635bb53f1a09b4716548ff69f/files/sql.rb:13:in `get'
	from /opt/puppetlabs/mcollective/tasks-spool/899e9e0635bb53f1a09b4716548ff69f/files/sql.rb:27:in `<main>'
```

Make sure `env_hash` is always a hash to fix this issue.

### DIFF
--- a/tasks/sql.rb
+++ b/tasks/sql.rb
@@ -6,7 +6,8 @@ require 'open3'
 require 'puppet'
 
 def get(sql, database, user, port, password, host)
-  env_hash = { 'PGPASSWORD' => password } unless password.nil?
+  env_hash = {}
+  env_hash['PGPASSWORD'] = password unless password.nil?
   cmd_string = "psql -c \"#{sql}\""
   cmd_string += " --dbname=#{database}" unless database.nil?
   cmd_string += " --username=#{user}" unless user.nil?


sheenaajay:
Thank you @smortex for your contribution.

sheenaajay:
Thanks @smortex . We have timeout issues on centos 8 for travis. We have opened a ticket for the issue. Will be disabling them till the issue is sorted. So that it wont hold any PRS. Thanks for your contribution.

sheenaajay (approved):

PR #1230, opened 2021-01-27
(bug) reload the service after the deletion of old unit file

sheenaajay:
Bug Details
Problem reported on Puppet version 6.20, OS Centos 8

Proposed solution
Issue systemctl daemon-reload first, after the deletion of old unit file followed by the restart newly configured service.

Create separate file resource for deleting old unit file and specifying the ordering to make sure it deletes the files first, then daemon- reload followed by service start

Puppet agent 5 Results
`Notice: /Stage[main]/Postgresql::Server::Config/File[old-systemd-override]/ensure: removed
Info: /Stage[main]/Postgresql::Server::Config/File[old-systemd-override]: Scheduling refresh of Exec[restart-systemd]
Debug: /Stage[main]/Postgresql::Server::Config/File[old-systemd-override]: The container Class[Postgresql::Server::Config] will propagate my refresh event
Info: /Stage[main]/Postgresql::Server::Config/File[old-systemd-override]: Scheduling refresh of Class[Postgresql::Server::Service]
Debug: /Stage[main]/Postgresql::Server::Config/Exec[restart-systemd]: 'systemctl daemon-reload' won't be executed because of failed check 'refreshonly'
Debug: Exec[restart-systemd](provider=posix): Executing 'systemctl daemon-reload'
Debug: Executing: 'systemctl daemon-reload'
Notice: /Stage[main]/Postgresql::Server::Config/Exec[restart-systemd]: Triggered 'refresh' from 1 event
Debug: /Stage[main]/Postgresql::Server::Config/Exec[restart-systemd]: The container Class[Postgresql::Server::Config] will propagate my refresh event
Debug: Prefetching parsed resources for postgresql_conf
Debug: Class[Postgresql::Server::Config]: The container Stage[main] will propagate my refresh event
Debug: Class[Postgresql::Server::Config]: The container Class[Postgresql::Server] will propagate my refresh event
Info: Class[Postgresql::Server::Service]: Scheduling refresh of Anchor[postgresql::server::service::begin]
Info: Class[Postgresql::Server::Service]: Scheduling refresh of Service[postgresqld]
Info: Class[Postgresql::Server::Service]: Scheduling refresh of Anchor[postgresql::server::service::end]
Notice: /Stage[main]/Postgresql::Server::Service/Anchor[postgresql::server::service::begin]: Triggered 'refresh' from 1 event
[...]

PR #1232, opened 2021-02-05
(IAC-1408) Release Prep v6.10.0

sheenaajay:


ekohl (approved):

ekohl:
FYI: we did see a pipeline fail on a PostgreSQL service restart which was running Puppet 6. I'm debugging that now, but if that's the same bug as #1230 then perhaps it needs additional work.

ekohl:
In the review of #1230 I was very much under the impression that it only affected Puppet 5. However, in our system upgrades we also see a problem with Puppet 6 which doesn't perform the `daemon-reload`. It is my theory that `/bin/systemctl show --property=NeedDaemonReload -- postgresql` returns something other than `yes` so that https://github.com/puppetlabs/puppet/blob/97ca00bcb935a2efe7a6cac1b8e071418b9a3dc4/lib/puppet/provider/service/systemd.rb#L150-L153 is never triggered. I think that's probably because the unit file is missing entirely according to systemd.

I'll try to write up a PR and I'd appreciate it if that could make it into the release.

sheenaajay:
> In the review of #1230 I was very much under the impression that it only affected Puppet 5. However, in our system upgrades we also see a problem with Puppet 6 which doesn't perform the `daemon-reload`. It is my theory that `/bin/systemctl show --property=NeedDaemonReload -- postgresql` returns something other than `yes` so that https://github.com/puppetlabs/puppet/blob/97ca00bcb935a2efe7a6cac1b8e071418b9a3dc4/lib/puppet/provider/service/systemd.rb#L150-L153 is never triggered. I think that's probably because the unit file is missing entirely according to systemd.
> 
> I'll try to write up a PR and I'd appreciate it if that could make it into the release.

Thanks alot @ekohl  will hold on the release .

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
 
 All notable changes to this project will be documented in this file. The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](http://semver.org).
 
+## [v6.10.0](https://github.com/puppetlabs/puppetlabs-postgresql/tree/v6.10.0) (2021-02-05)
+
+[Full Changelog](https://github.com/puppetlabs/puppetlabs-postgresql/compare/v6.9.0...v6.10.0)
+
+### Added
+
+- Set default PostgresQL version for FreeBSD [\#1227](https://github.com/puppetlabs/puppetlabs-postgresql/pull/1227) ([olevole](https://github.com/olevole))
```
ekohl, line 11: Nit: should it be fixed to
```suggestion
- Set default PostgreSQL version for FreeBSD [\#1227](https://github.com/puppetlabs/puppetlabs-postgresql/pull/1227) ([olevole](https://github.com/olevole))
```

ekohl (commented):
I submitted https://github.com/puppetlabs/puppetlabs-postgresql/pull/1233 and would like to see it included.

sheenaajay:
> I submitted #1233 and would like to see it included.

Yeah sure. Thank you

ekohl (approved):

PR #1233, opened 2021-02-05
Also perform systemd daemon-reload on Puppet 6.1+

ekohl:
The way daemon-reload is implemented in Puppet 6.1 is by calling

```
systemctl show --property=NeedDaemonReload -- $SERVICE
```

If that returns yes then it performs systemctl daemon-reload. However, this
doesn't work reliable because of the particular case.

```console
# systemctl show --property=NeedDaemonReload -- postgresql
Failed to get properties: Access denied
```

This is probably because the original unit file is removed.

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -217,8 +217,9 @@ class postgresql::server::config {
   # Gentoo also supports drop-in files.
   if $facts['os']['family'] in ['RedHat', 'Gentoo'] and $facts['service_provider'] == 'systemd' {
     # While Puppet 6.1 and newer can do a daemon-reload if needed, systemd
-    # doesn't appear to report that correctly. This is probably because its
-    # unit file is actually removed.
+    # doesn't appear to report that correctly in all cases.
+    # One such case seems to be when an overriding unit file is removed from /etc
+    # and the original one from /lib *should* be used again.
     #
     # This can be removed when Puppet < 6.1 support is dropped *and* the file
     # old-systemd-override is removed.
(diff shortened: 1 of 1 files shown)


manifests/server/config.pp
```diff
@@ -216,42 +216,40 @@
...
-        path        => '/bin:/usr/bin:/usr/local/bin',
-        before      => Class['postgresql::server::service'],
-      }
-      $systemd_notify = [Exec['restart-systemd'], Class['postgresql::server::service']]
-    }
-    else {
-      $systemd_notify = Class['postgresql::server::service']
+    # While Puppet 6.1 and newer can do a daemon-reload if needed, systemd
+    # doesn't appear to report that correctly. This is probably because its
+    # unit file is actually removed.
```
evgeni, line 221:
```suggestion
    # While Puppet 6.1 and newer can do a daemon-reload if needed, systemd
    # doesn't appear to report that correctly in all cases.
    # One such case seems to be when an overriding unit file is removed from /etc
    # and the original one from /lib *should* be used again.
```

sheenaajay (approved):

evgeni (approved):
wording nitpick on the explanation in the comment, ACK otherwise

this is in line with what we discussed on IRC regarding how and why we've been seeing this error on Puppet6 and how to avoid it.

manifests/server/config.pp
ekohl: Thanks!

sheenaajay:
Thank you @ekohl for your contribution.

ekohl:
FYI: this is https://tickets.puppetlabs.com/browse/PUP-9473.

PR #1239, opened 2021-02-16
Add DNF module management

ekohl:
On EL8 and Fedora DNF can manage modules. This is a method of providing multiple versions on the same OS. Only one DNF module can be active at the same time.

### DIFF
--- a/manifests/dnfmodule.pp
+++ b/manifests/dnfmodule.pp
@@ -1,0 +1,20 @@
+# @summary Manage the DNF module
+#
+# On EL8 and Fedora DNF can manage modules. This is a method of providing
+# multiple versions on the same OS. Only one DNF module can be active at the
+# same time.
+#
+# @api private
+class postgresql::dnfmodule (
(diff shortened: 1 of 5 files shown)


ekohl:
I am running this on my own server and there it appears to work well.

Zak-Kent:
fwiw, when investigating [Server-2500](https://tickets.puppetlabs.com/browse/SERVER-2500) and a workaround which was added for this issue in the PuppetDB and Puppetserver acceptance tests I found this pr. I was able to apply the changes here to my test setup and they resolved this issue we were seeing on RHEL-8. I don't have a good sense of what else might be impacted by these changes but once some flavor of this is merged we should be able to update the [puppetlabs-puppetdb](https://github.com/puppetlabs/puppetlabs-puppetdb) module to pick up this fix.

sheenaajay (approved):

sheenaajay:
Thanks @ekohl for your contribution as always.

PR #1240, opened 2021-02-19
Fix command shell escaping

DavidS:
This changes a couple of places where the postgresql module was passing unsafe values through a shell to avoid the shell altogether.

Big thanks to @smortex for reporting the issue.

### DIFF
--- a/lib/puppet/provider/postgresql_psql/ruby.rb
+++ b/lib/puppet/provider/postgresql_psql/ruby.rb
@@ -16,7 +16,7 @@ Puppet::Type.type(:postgresql_psql).provide(:ruby) do
     command = [resource[:psql_path]]
     command.push('-d', resource[:db]) if resource[:db]
     command.push('-p', resource[:port]) if resource[:port]
-    command.push('-t', '-X', '-c', '"' + sql.gsub('"', '\"') + '"')
+    command.push('-t', '-X', '-c', sql)
 
     environment = fetch_environment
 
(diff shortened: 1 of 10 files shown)


DavidS:
The rhel-7, puppet7-nightly failure is IAC-1420, so we can ignore that and get this merged and released on Monday.

sheenaajay (approved):

ekohl (approved):
I love it when you have more functionality with less code. I also hate it when code tries to join commands when you can actually pass an array of arguments. Big :+1: for this direction.

ekohl:
Not sure if this introduced it, but in another module I started to see:
```
  Warning: /Postgresql_psql[ALTER ROLE jiraadm ENCRYPTED PASSWORD ****]: Unable to mark 'unless' as sensitive: unless is a parameter and not a property, and cannot be automatically redacted.
```

vchepkov:
```Warning: /Postgresql_psql[ALTER ROLE puppetdb ENCRYPTED PASSWORD ****]: Unable to mark 'unless' as sensitive: unless is a parameter and not a property, and cannot be automatically redacted.```

vchepkov:
@DavidS , I think `unless` can't be Sensitive, same as `only_if`, btw
https://github.com/puppetlabs/puppetlabs-postgresql/blob/main/manifests/server/role.pp#L136

smortex:
@ekohl @vchepkov I just opened #1249 to track this issue.

ekohl:
https://github.com/puppetlabs/puppetlabs-postgresql/pull/1253 is another regression I found with this PR.

PR #1249, opened 2021-02-26
Fix warning introduced in #1240 (08a1752)

smortex:
`unless` does not accept a Sensitive parameter, and passing one result in a warning being output:

```
(/Postgresql_psql[ALTER ROLE puppetdb ENCRYPTED PASSWORD ****]) Unable to mark 'unless' as sensitive: unless is a parameter and not a property, and cannot be automatically redacted.
```

This is not really an issue here since this query is not displayed in the logs, even in debug mode.

### DIFF
--- a/manifests/server/role.pp
+++ b/manifests/server/role.pp
@@ -133,7 +133,7 @@ define postgresql::server::role (
       }
       postgresql_psql { "ALTER ROLE ${username} ENCRYPTED PASSWORD ****":
         command     => Sensitive("ALTER ROLE \"${username}\" ${password_sql}"),
-        unless      => Sensitive("SELECT 1 FROM pg_shadow WHERE usename = '${username}' AND passwd = '${pwd_hash_sql}'"),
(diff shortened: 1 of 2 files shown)


smortex:
Cc authors of #1240 : @DavidS @sheenaajay

ekohl (approved):

sheenaajay:
@smortex Thanks alot for the quick fix.Much appreciated for your time.
Just wondering if we need a sanitisation mechanism  to avoid shell injection for the values passed for username/password.

ekohl:
Since that never was in place, perhaps this should be merged now to avoid the regression (warning). I can't think of a solution that does work.

sheenaajay:
Thanks @ekohl and @smortex .Just had a quick discussion with team and we are looking into it. Will get the fix in. Thank you.

sheenaajay:
Following ticket is opened for sensitive data support needed for unless command
https://tickets.puppetlabs.com/browse/PUP-10950. Thank you.

sheenaajay:
Following changes added to fix the warning message. Thank you.
https://github.com/puppetlabs/puppetlabs-postgresql/pull/1258

smortex:
Replaced by #1258

PR #1279, opened 2021-06-12
Use Puppet-Datatype Sensitive for Passwords

cocker-cc:
- add Parameter "sensitive" to Function postgresql_password to decide if its Returnvalue should be of Datatype Sensitive
- let defined Type postgresql::server::role accept Datatype Sensitive for $password_hash
- let defined Type postgresql::server::db accept Datatype Sensitive for $password
- let Class postgresql::server accept Datatype Sensitive for $postgres_password
- let defined Type postgresql::validate_db_connection accept Sensitive for $database_password

### DIFF
--- a/lib/puppet/functions/postgresql/postgresql_password.rb
+++ b/lib/puppet/functions/postgresql/postgresql_password.rb
@@ -6,15 +6,25 @@ Puppet::Functions.create_function(:'postgresql::postgresql_password') do
   #   The clear text `username`
   # @param password
   #   The clear text `password`
+  # @param sensitive
+  #   If the Postgresql-Passwordhash should be of Datatype Sensitive[String]
   #
-  # @return [String]
+  # @return
   #   The postgresql password hash from the clear text username / password.
   dispatch :default_impl do
-    param 'Variant[String[1],Integer]', :username
-    param 'Variant[String[1],Integer]', :password
+    required_param 'Variant[String[1], Integer]', :username
+    required_param 'Variant[String[1], Sensitive[String[1]], Integer]', :password
+    optional_param 'Boolean', :sensitive
+    return_type 'Variant[String, Sensitive[String]]'
   end
 
-  def default_impl(username, password)
-    'md5' + Digest::MD5.hexdigest(password.to_s + username.to_s)
+  def default_impl(username, password, sensitive = false)
+    password = password.unwrap if password.respond_to?(:unwrap)
+    result_string = 'md5' + Digest::MD5.hexdigest(password.to_s + username.to_s)
+    if sensitive
+      Puppet::Pops::Types::PSensitiveType::Sensitive.new(result_string)
+    else
+      result_string
+    end
   end
 end
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -83,7 +83,7 @@
 # @param extra_systemd_config Adds extra config to systemd config file, can for instance be used to add extra openfiles. This can be a multi line string
 #
 class postgresql::server (
-  $postgres_password                               = undef,
+  Optional[Variant[String[1], Sensitive[String[1]], Integer]] $postgres_password = undef,
 
   $package_name                                    = $postgresql::params::server_package_name,
   $package_ensure                                  = $postgresql::params::package_ensure,
--- a/manifests/server/passwd.pp
+++ b/manifests/server/passwd.pp
@@ -1,6 +1,11 @@
 # @api private
 class postgresql::server::passwd {
-  $postgres_password = $postgresql::server::postgres_password
+  $postgres_password = if $postgresql::server::postgres_password =~ Sensitive {
+    $postgresql::server::postgres_password.unwrap
+  } else {
+    $postgresql::server::postgres_password
+  }
+
   $user              = $postgresql::server::user
   $group             = $postgresql::server::group
   $psql_path         = $postgresql::server::psql_path
--- a/.sync.yml
+++ b/.sync.yml
@@ -1,6 +1,9 @@
 ---
 ".gitlab-ci.yml":
   delete: true
+.puppet-lint.rc:
+  extra_disabled_lint_checks:
+    - 140chars-check
 appveyor.yml:
   delete: true
 
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -863,7 +863,7 @@ The following parameters are available in the `postgresql::server` class:
 
 ##### <a name="postgres_password"></a>`postgres_password`
 
-Data type: `Any`
+Data type: `Variant[String, Sensitive[String]]`
 
 Sets the password for the postgres user to your specified value. By default, this setting uses the superuser account in the Postgres database, with a user called postgres and no password.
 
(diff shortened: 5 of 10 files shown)


.puppet-lint.rc
```diff
@@ -1 +1,2 @@
 --relative
+--no-140chars-check
```
ekohl, line 2: This is included in the PDK template (https://github.com/puppetlabs/pdk-templates/blob/main/moduleroot/.puppet-lint.rc.erb) without a matching `.sync.yml` entry. That means it'll be overridden on the next update.

lib/puppet/functions/postgresql/postgresql_password.rb
```diff
@@ -6,15 +6,25 @@
   #   The clear text `username`
   # @param password
   #   The clear text `password`
+  # @param sensitive
+  #   If the Postgresql-Passwordhash should be of Datatype Sensitive[String]
   #
   # @return [String]
```
ekohl, line 12: If you set `return_type` (which I like), you can drop the type here.
```suggestion
  # @return
```

manifests/server/passwd.pp
```diff
@@ -1,6 +1,11 @@
 # @api private
 class postgresql::server::passwd {
-  $postgres_password = $postgresql::server::postgres_password
+  if $postgresql::server::postgres_password =~ Sensitive[String] {
```
ekohl, line 3: No need to be so specific here
```suggestion
  if $postgresql::server::postgres_password =~ Sensitive {
```

manifests/server.pp
```diff
@@ -83,7 +83,7 @@
 # @param extra_systemd_config Adds extra config to systemd config file, can for instance be used to add extra openfiles. This can be a multi line string
 #
 class postgresql::server (
-  $postgres_password                               = undef,
+  Optional[Variant[String, Sensitive[String]]] $postgres_password = undef,
```
ekohl, line 86: Just thinking about compatibility here: could a user specify an integer password?

.puppet-lint.rc
cocker-cc: Just a quick Question:  Should I `git commit --amend && git push -f` the Changes, or should I append another Commit?

manifests/server.pp
cocker-cc: You mean Compatibility to `postgresql::postgresql_password`?
```
    required_param 'Variant[String[1], Sensitive[String[1]], Integer]', :password
```
I don't see `Integer` as valid there neither.

.puppet-lint.rc
cocker-cc: I added the Check to `.sync.yml`, too.

lib/puppet/functions/postgresql/postgresql_password.rb
cocker-cc: Suggestion implemented.

manifests/server/passwd.pp
cocker-cc: Suggestion implemented.

ekohl (commented):
Given what we've seen in theforeman/foreman, I think a function `stdlib::unwrap_sensitive` may actually make sense, given how common it is that you need to unwrap something. What do you think?

cocker-cc:
> Given what we've seen in theforeman/foreman, I think a function `stdlib::unwrap_sensitive` may actually make sense, given how common it is that you need to unwrap something. What do you think?

I have [this PR](https://github.com/puppetlabs/puppet/pull/8651) pending.  If and when this is accepted, your suggested Function is not needed anymore.  Perhaps You could vote for this PR.

david22swan:
@ekohl Now that the associated pr has been merged what are your thoughts on this?
Everything LGTM but don't want to step on your toes since you've got an open comment on it.

lib/puppet/functions/postgresql/postgresql_password.rb
```diff
@@ -6,15 +6,25 @@
...
+  #   If the Postgresql-Passwordhash should be of Datatype Sensitive[String]
   #
-  # @return [String]
+  # @return
   #   The postgresql password hash from the clear text username / password.
   dispatch :default_impl do
-    param 'Variant[String[1],Integer]', :username
-    param 'Variant[String[1],Integer]', :password
+    required_param 'Variant[String[1], Integer]', :username
+    required_param 'Variant[String[1], Sensitive[String[1]], Integer]', :password
```
ekohl, line 16: You are removing the `Integer` as an input value, which was valid before.
```suggestion
    required_param 'Variant[String[1], Integer, Sensitive[String[1]], Integer]', :password
```

manifests/server.pp
ekohl: Due to other work I didn't have time to look into this for a while. How is it not valid?

cocker-cc: In my Understanding a Password is a String, not an Integer.  But anyway, I adapted your Suggestion:  

```
Optional[Variant[String[1], Sensitive[String[1]], Integer]] $postgres_password = undef,
```

ekohl: Just for your understanding, you can have a numeric password and if you don't quote it in YAML (with Hiera for example) it becomes `Integer`. I agree it's not that likely, but it can happen and we should be consistent.

manifests/server/passwd.pp
```diff
@@ -1,6 +1,11 @@
 # @api private
 class postgresql::server::passwd {
-  $postgres_password = $postgresql::server::postgres_password
+  $postgres_password = if $postgresql::server::postgres_password =~ Sensitive {
```
david22swan, line 3: Just one comment, and I apologize for it being at the very last second, but the unwrap shown here is not needed from what I understand.
According to the official documentation on the type, the sensitive data type only effects the logs and reports, when it comes to the actual changes being applied it has no effect. Unwrap is only used in situations where you want the value to be output into a specific report or log.
```
Sensitive types in the Puppet language are strings marked as sensitive.
The value is displayed in plain text in the catalog and manifest, but is redacted from logs and reports.
Because the value is maintained as plain text, use it only as an aid to ensure that sensitive values are not inadvertently disclosed.
```
This is taken from the documentation for the type shown here: https://puppet.com/docs/puppet/7/lang_data_sensitive.html

Again, apologies for not mentioning this earlier but I only just became aware of this after utilising the Sensitive type in a pr of my own.

david22swan: Gonna resolve this as the other use's on unwrap are valid and seem necessary for the code to work as they involve manipulation of the value.
Sorry for the bother.

david22swan:
@cocker-cc Everything LGTM so gonna go ahead and merge.
Thank you for all the work you have put in and I look forward to seeing more from you in the future :)

david22swan (approved):

david22swan:
Just quickly closed and reopened to rekick the test's. No need to worry

manifests/server/passwd.pp
cocker-cc: > but the unwrap shown here is not needed from what I understand.

My main Goal is, to make Component-Modules like `postgresql` to accept `Sensitive`, because an Author of Profile-Modules should not bother with `unwrap`.  Component-Modules "just have to work".  My `if`-Clause is cheap Means to achive this.
But for the Internals of the Module `postgresql` you are right, in the Way, that `unwrap` should happen in the least possible Situation.  This means: not the Class `postgresql::server::passwd` should take Care of it, but instead the following Function `postgresql::postgresql_escape` and the `exec`.

PR #1282, opened 2021-07-02
Do not add version component to repo definition

weastur:
According to the [official documentation](https://www.postgresql.org/download/linux/ubuntu/) there is no need to add version component to repository line.

There are no predefined list of allowed components names, but some tools, like `command-not-found`, manage they own list and do not expect to meet something like `13` in component name. Source code [here](https://salsa.debian.org/jak/command-not-found/-/blob/master/CommandNotFound/db/creator.py#L18) This causes exception like this while `apt update`

```
Traceback (most recent call last):
  File "/usr/lib/cnf-update-db", line 26, in <module>
    col.create(db)
  File "/usr/share/command-not-found/CommandNotFound/db/creator.py", line 94, in create
    self._fill_commands(con)
  File "/usr/share/command-not-found/CommandNotFound/db/creator.py", line 132, in _fill_commands
    self._parse_single_contents_file(con, f, fp.stdout)
  File "/usr/share/command-not-found/CommandNotFound/db/creator.py", line 274, in _parse_single_contents_file
    priority = component_priorities[component]
KeyError: '13'
```

Strictly speaking this is not the direct problem of this module, but my suggestion is to align behaviour of puppet module with official PostgreSQL documentation.

### DIFF
--- a/manifests/repo/apt_postgresql_org.pp
+++ b/manifests/repo/apt_postgresql_org.pp
@@ -17,7 +17,7 @@ class postgresql::repo::apt_postgresql_org inherits postgresql::repo {
   -> apt::source { 'apt.postgresql.org':
     location => $_baseurl,
     release  => "${facts['os']['distro']['codename']}-pgdg",
-    repos    => "main ${postgresql::repo::version}",
+    repos    => 'main',
     key      => {
       id     => 'B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8',
       source => 'https://www.postgresql.org/media/keys/ACCC4CF8.asc',


manifests/repo/apt_postgresql_org.pp
```diff
@@ -17,7 +17,7 @@
   -> apt::source { 'apt.postgresql.org':
     location => $_baseurl,
     release  => "${facts['os']['distro']['codename']}-pgdg",
-    repos    => "main ${postgresql::repo::version}",
+    repos    => "main",
```
ekohl, line 20:
```suggestion
    repos    => 'main',
```

ekohl (commented):
https://wiki.postgresql.org/wiki/Apt tells me this isn't just for Ubuntu but just `apt.postgresql.org` in general so :+1: from me.

I've allowed it to run the tests so we'll see what they say when they complete or if you need to  You may need to modify some tests. At least the style guide says to use single quotes for strings without variables (and I think puppet-lint also enforces this).

manifests/repo/apt_postgresql_org.pp
weastur: fixed

smortex (approved):

ekohl (approved):

sheenaajay (approved):

david22swan (approved):

david22swan:
@weastur Apologies for the wait but everything is looking good to me so I'm gonna go ahead and merge.
The CentOS 8 failures are a known issue unrelated to this pr.

Thanks for putting in the work and I look forward to seeing more from you in the future

PR #1285, opened 2021-08-03
(maint) - Set max_issues for changelog generator to 500

david22swan:
Tests failures are from a known issue and unrelated to changes.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -48,6 +48,7 @@ if Bundler.rubygems.find_name('github_changelog_generator').any?
     raise "Set CHANGELOG_GITHUB_TOKEN environment variable eg 'export CHANGELOG_GITHUB_TOKEN=valid_token_here'" if Rake.application.top_level_tasks.include? "changelog" and ENV['CHANGELOG_GITHUB_TOKEN'].nil?
     config.user = "#{changelog_user}"
     config.project = "#{changelog_project}"
+    config.max_issues = 500
     config.future_release = "#{changelog_future_release}"
     config.exclude_labels = ['maintenance']
     config.header = "# Change log\n\nAll notable changes to this project will be documented in this file. The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](http://semver.org)."
--- a/.sync.yml
+++ b/.sync.yml
@@ -10,6 +10,7 @@ Gemfile:
     - gem: github_changelog_generator
 Rakefile:
   changelog_user: puppetlabs
+  changelog_max_issues: 500
 spec/spec_helper.rb:
   mock_with: ":rspec"
   coverage_report: true


Rakefile
```diff
@@ -53,6 +53,7 @@ if Bundler.rubygems.find_name('github_changelog_generator').any?
     config.header = "# Change log\n\nAll notable changes to this project will be documented in this file. The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](http://semver.org)."
     config.add_pr_wo_labels = true
     config.issues = false
+    config.max_issues = 500
```
alexjfisher, line 56: After 500 issues, what happens?  Is it the oldest ones that get dropped?

alexjfisher: actually, how does this even apply for this project, where github issues are disabled?

david22swan: No, nothing get's dropped, it just ignores anything past that. If they are already in the changelog they will stay there, but if something where to happen, i.e. their label where to change, they wouldn't be updated like they would be now.

alexjfisher: Are you sure?  I can't reproduce that.  Testing this behaviour is a bit tricky though as even if you set max_issues to a very small number, a minimum of 100 are fetched. https://github.com/github-changelog-generator/github-changelog-generator/blob/d801469b07932c3647fafa7fed2796fa3b9ebff4/lib/github_changelog_generator/octo_fetcher.rb#L17

david22swan: Yeh, we previously used this fix for apache, though it was accidentally stripped by a pdk update, so I'm confident it will cause no problems

david22swan: Have tested the fix and it worked: https://github.com/puppetlabs/puppetlabs-postgresql/actions/runs/1093892098

sheenaajay (approved):

Rakefile
alexjfisher: ok.  I'm guessing that anything that wasn't fetched happens to be older than 5.3.0 and is covered by what's in [HISTORY.md](https://github.com/puppetlabs/puppetlabs-postgresql/blob/main/HISTORY.md#530)

david22swan: Ah, no.
This work's separately from the HISTORY.md.

With the HISTORY.md it set's a hard end point, cover everything up to this and if something doesn't match what it retrieves, it changes it.
This change however rather than setting a goal, makes it so that once it reaches the set amount of issues, it it stops looking any further and simply accepts what comes after that point is correct and doesn't need to be looked at.

I'm sorry if I'm not explaining it that well.

sheenaajay:
Thank you @david22swan @alexjfisher

Rakefile
alexjfisher: I always thought the CHANGELOG.md was regenerated completely when github-changelog-generator was run.
So... I tested by setting max_issues to 100 (the smallest number that makes a difference since the issues are fetched in batches of 100).  Existing entries get removed (unless they're part of HISTORY.md that just gets appended verbatim).  See https://github.com/alexjfisher/puppetlabs-postgresql/commit/123ab705155a1288a46dec9321a58ffefa0ad49f

But I guess if this ever does become a problem then there's the solution. Just prepend some of the existing CHANGELOG.md to HISTORY.md as a workaround.

PR #1267, opened 2021-04-15
MODULES-11049 - Implement default privileges changes

mtancoigne:
Following the [feature request](https://tickets.puppetlabs.com/browse/MODULES-11049), this PR allow to grant or revoke default privileges on objects in a schema.

I still have to update the documentation.

PR #1289, opened 2021-08-11
[IAC-1735] - update spec tests

adrianiurca:
The tests related to classes or defines which are using `ensure_packages` function from `stdlib` needs to be updated accordingly to this PR https://github.com/puppetlabs/puppetlabs-stdlib/pull/1196

### DIFF
--- a/spec/spec_helper_local.rb
+++ b/spec/spec_helper_local.rb
@@ -44,10 +44,6 @@ def param(type, title, param)
   param_value(catalogue, type, title, param)
 end
 
-def postgresql_version
-  '9.6'
-end
-
 shared_examples 'postgresql_password function' do
   it { is_expected.not_to eq(nil) }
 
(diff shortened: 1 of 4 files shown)


spec/spec_helper_local.rb
```diff
@@ -44,6 +44,10 @@ def param(type, title, param)
   param_value(catalogue, type, title, param)
 end
 
+def postgresql_version
+  '9.6'
+end
+
```
smortex, line 50: This function does not seem to be used in the rest of the changes of this PR.  Was it added by mistake?

smortex (commented):
LGTM, but I am confused about this new function :thinking:

ekohl (approved):
I removed the additional method. Will squash on merge.

ekohl:
Ah, I now see this was added because it's actually used in `spec/unit/defines/server/default_privileges_spec.rb`. https://github.com/puppetlabs/puppetlabs-postgresql/pull/1267/files#r689125725 was wrong to begin with.

ekohl:
I opened https://github.com/puppetlabs/puppetlabs-postgresql/pull/1291 which takes most of this commit, but fixes the broken unit test separately with let blocks.

PR #1296, opened 2021-09-08
Fix changing default encoding

smortex:
Starting with 79fa35591e099a7b6981b873b930b943e5dc1827 it is not
possible to set a custom encoding on systems where initdb is not needed
(e.g. Debian).  The autorequire introduce a circular dependency:

```
dependency cycles found: (Anchor[postgresql::server::service::begin] =>
                          Postgresql_psql[Set template1 encoding to UTF-8] =>
                          Class[Postgresql::Server::Initdb] =>
                          Postgresql_conf[listen_addresses] =>
                          Class[Postgresql::Server::Service] =>
                          Anchor[postgresql::server::service::begin])
```

Move the relevant code from postgresql::server::initdb to
postgresql::server::late_initdb and include it when applicable so that
the postgresql_psql command gets run after the service has started up.

### DIFF
--- a/manifests/server/late_initdb.pp
+++ b/manifests/server/late_initdb.pp
@@ -1,0 +1,39 @@
+# @summary Manage the default encoding when database initialization is managed by the package
+#
+# @api private
+class postgresql::server::late_initdb {
+  assert_private()
+
+  $encoding       = $postgresql::server::encoding
+  $user           = $postgresql::server::user
+  $group          = $postgresql::server::group
+  $psql_path      = $postgresql::server::psql_path
+  $port           = $postgresql::server::port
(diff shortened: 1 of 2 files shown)


manifests/server/late_initdb.pp
```diff
@@ -0,0 +1,37 @@
+# @api private
```
bastelfreak, line 1: please also add a `@summary`  here

smortex: Done (other private classes do not have one but it should not hurt to tell what the class is about) :smile:

bastelfreak (approved):

david22swan (approved):
LGTM
Rekicking tests
Will merge once they are green

PR #1298, opened 2021-09-15
Support setting default_privileges on all schemas

fish-face:
The Postgres default is for the absent specification of a schema name
when altering default privileges to apply to all schemas.
Support that behaviour, but keep the current default behaviour for
an unset schema parameter.

This means that the behaviour doesn't match up with the postgres specification - a better match would be if leaving the `schema` parameter unset also left off the `IN SCHEMA` clause, but this would break backwards compatibility.

### DIFF
--- a/manifests/server/default_privileges.pp
+++ b/manifests/server/default_privileges.pp
@@ -6,7 +6,7 @@
 # @param db Specifies the database to which you are granting access.
 # @param object_type Specify target object type: 'FUNCTIONS', 'ROUTINES', 'SEQUENCES', 'TABLES', 'TYPES'.
 # @param privilege Specifies comma-separated list of privileges to grant. Valid options: depends on object type.
-# @param schema Target schema. Defaults to 'public'.
+# @param schema Target schema. Defaults to 'public'. Can be set to '' to apply to all schemas.
 # @param psql_db Defines the database to execute the grant against. This should not ordinarily be changed from the default.
 # @param psql_user Specifies the OS user for running psql. Default value: The default user for the module, usually 'postgres'.
 # @param psql_path Specifies the OS user for running psql. Default value: The default user for the module, usually 'postgres'.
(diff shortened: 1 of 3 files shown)


manifests/server/default_privileges.pp
```diff
@@ -128,12 +135,12 @@
   }
 
   $_unless = $ensure ? {
-    'absent' => "SELECT 1 WHERE NOT EXISTS (SELECT * FROM pg_default_acl AS da JOIN pg_namespace AS n ON da.defaclnamespace = n.oid WHERE '%s=%s' = ANY (defaclacl) AND nspname = '%s' and defaclobjtype = '%s')",
-    default  => "SELECT 1 WHERE EXISTS (SELECT * FROM pg_default_acl AS da JOIN pg_namespace AS n ON da.defaclnamespace = n.oid WHERE '%s=%s' = ANY (defaclacl) AND nspname = '%s' and defaclobjtype = '%s')"
+    'absent' => "SELECT 1 WHERE NOT EXISTS (SELECT * FROM pg_default_acl AS da LEFT JOIN pg_namespace AS n ON da.defaclnamespace = n.oid WHERE '%s=%s' = ANY (defaclacl)%s and defaclobjtype = '%s')",
+    default  => "SELECT 1 WHERE EXISTS (SELECT * FROM pg_default_acl AS da LEFT JOIN pg_namespace AS n ON da.defaclnamespace = n.oid WHERE '%s=%s' = ANY (defaclacl)%s and defaclobjtype = '%s')"
```
fish-face, line 139: I think actually there is an issue here: if you leave `schema` unset and the default privilege exists for a specific schema, the subquery here will return a row, blocking a change.

fish-face: This should be fixed with the most recent force-push.

fish-face:
The inline documentation reflected a previous version of the code; this has now been fixed. In particular this should now be backwards compatible. (I think an incompatible change would result in a better interface, but I don't know what requirements you'd have in this case. It could be a very significant difference in a small number of cases!)

fish-face:
I think the only failure here was a timeout - is there anything you need from me?

david22swan:
@fish-face Sorry for leaving this so long, been a little full on here lately. 
Yeh I'm happy to merge this, not sure when it will get released though.

Anyway thanks again for the hard work

david22swan (approved):

PR #1300, opened 2021-09-22
Default privileges support schemas

fish-face:
WIP because it is dependent on [PR 1298](https://github.com/puppetlabs/puppetlabs-postgresql/pull/1298) since schema privileges require the `schema` parameter to be applying to all schemas (schemas can't be nested, so it doesn't make sense to set privileges for new schemas created "within the public schema" or something).

This would also benefit from acceptance tests once I know how to make them work.

### DIFF
--- a/manifests/server/default_privileges.pp
+++ b/manifests/server/default_privileges.pp
@@ -23,7 +23,8 @@ define postgresql::server::default_privileges (
     /(?i:^ROUTINES$)/,
     /(?i:^SEQUENCES$)/,
     /(?i:^TABLES$)/,
-    /(?i:^TYPES$)/
+    /(?i:^TYPES$)/,
+    /(?i:^SCHEMAS$)/
   ] $object_type,
   String $schema                   = 'public',
   String $psql_db                  = $postgresql::server::default_database,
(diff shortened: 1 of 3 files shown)


fish-face:
Any advice on getting the tests which rely on specific versions working?

fish-face:
I think the issue was simply that the check in the module was using `versioncmp(..., '10.0')` rather than `versioncmp(..., '10')` - can you confirm that the latter is correct?

manifests/server/default_privileges.pp
```diff
@@ -140,6 +141,21 @@
       }
       $_check_type = 'T'
     }
+    'SCHEMAS': {
+      if (versioncmp($version, '10') == -1) {
```
ekohl, line 145: Perhaps this works better?
```suggestion
      unless (versioncmp($version, '10') >= 0) {
```

fish-face: I think this would be quite surprising since it would silently ignore attempts to alter schema privileges on unsupported platforms, rather than warning about the situation? Probably the result is that the desired database security is not there and it's either inaccessible or insecure...

ekohl: I think what I suggested actually makes the check work, but tests should properly indicate this.

fish-face: Ah I misread very badly! I don't know what the preferred semantics is ("fail if X" vs "fail unless not X") so I can invert that if desired. BTW just in case you saw the older comments and not the most recent one, it's perhaps worth repeating that the tests that didn't time out look successful now (I changed `10.0` to `10`, having not realised this behaviour of `versioncmp`)

ekohl: Yes, there's some odd behavior and I think this syntax will work with both `9.x` and `10`, `11` etc.

fish-face: so the version supplied is `'10'`, `versioncmp('10', '10.0') == -1` and `versioncmp('10', '10') == 0`, so the check was triggering incorrectly and is no longer. Also `versioncmp('10.0', '10') == 1`, so even if a longer version string were passed it should still be OK.

I am not sure if you still want me to switch out `if` for `unless` and invert the check though?

ekohl: If this implementation works then I'm fine with it and no need to change it.

fish-face: OK, let's keep it that way - as far as I can tell the only issue now is the timed out CI runs which I believe are ever-present!

fish-face:
Marked as not WIP.

fish-face:
Hi, aware you are pretty busy - if you are able to have another look at this soon that would be great.

ekohl:
I'm away until next week.

ekohl (approved):
Not my area of PostgreSQL but the Puppet code looks ok so I'm merging this. Thanks!

PR #1302, opened 2021-09-27
(IAC-1598) - Remove Support for Debian 8

david22swan:


ekohl (approved):
Looks like that other than `params.pp` and `globals.pp` there's actually no code to support Debian 8 so I'm fine with removing this just from metadata.json.

PR #1303, opened 2021-09-27
Add Red Hat Enterprise Linux 9 support

ekohl:
The comment indicates this is something that needs to be done on EL6 and older. The current check would also trigger on EL9. This uses a version comparison on the major version. We can leave out the Fedora check since we don't need to care about Fedora < 7.

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -183,6 +183,7 @@ class postgresql::globals (
       },
       'Amazon' => '9.2',
       default => $facts['os']['release']['major'] ? {
+        '9'     => '13',
         '8'     => '10',
         '7'     => '9.2',
         '6'     => '8.4',
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -82,28 +82,12 @@ class postgresql::params inherits postgresql::globals {
         $postgresql_conf_mode   = pick($postgresql_conf_mode, '0600')
       }
 
-      case $facts['os']['name'] {
-        'Amazon': {
-          $service_reload = "service ${service_name} reload"
-          $service_status = "service ${service_name} status"
-        }
-
-        # RHEL 5 uses SysV init, RHEL 6 uses upstart, RHEL 7+ uses systemd.
-        'RedHat', 'CentOS', 'Scientific', 'OracleLinux': {
-          if versioncmp($facts['os']['release']['major'], '7') >= 0 {
-            $service_reload = "systemctl reload ${service_name}"
-            $service_status = "systemctl status ${service_name}"
-          } else {
-            $service_reload = "service ${service_name} reload"
-            $service_status = "service ${service_name} status"
-          }
-        }
-
-        # Default will catch Fedora which uses systemd
-        default: {
-          $service_reload = "systemctl reload ${service_name}"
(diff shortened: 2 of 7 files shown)


bastelfreak:
is that all to get EL 9 support? Because that's currently not listed in the metadata.json.

ekohl:
It probably needs more work for real EL9 support. I was going over the places where the OS major version was used and this jumped out. It also needs to set a default PG version in globals.pp.

bastelfreak (dismissed):

ekohl:
I fixed a versioncmp check and added the default version in globals.pp.

bastelfreak (approved):

ekohl:
I must admit I didn't have an EL9 box at hand to actually test this out. That's why I've labeled it as initial support and didn't add it to metadata.json.

smortex (approved):
Looks good!

Can't we count on acceptance tests as a base for supporting this OS?

ekohl:
Not easily. I started in Beaker but the problem is that you don't have RPMs for EL9 and the EL8 ones don't install. See https://github.com/voxpupuli/beaker-hostgenerator/pull/225#discussion_r705461162 as well.

ekohl:
Looks like CI here doesn't automatically run tests on EL9 if you add it to metadata.

kajinamit:
I think you should also update manifests/params.pp. Please find the below change I made in another PR.

https://github.com/puppetlabs/puppetlabs-postgresql/pull/1321/files#diff-e78db79abb0b543310767fb9d2814cc4e2408f613de4985186251c70bdd4edbb

ekohl:
I've chosen to rely on the service_provider param/fact instead. I'm still not sure how PDK determines how to run tests acceptance tests.

ekohl:
I'm still blocked on a question to the Puppet folks for actual testing.

chelnak:
I've removed the label for now. This one will have remain on hold until we support the required OS.

ekohl:
Rebased to resolve conflicts and added RHEL 9 to metadata.json.

ekohl:
Like this? Should I also update the (last) commit message or is the current one sufficient?

david22swan:
@ekohl Yeh like that, but could you also remove the CentOS Stream 9 from the title and metadata.
Since the puppet agent doesn't support it yet we can't test it yet and so can't claim official support, had missed that you added it to the metadata as well sorry.
We're using the below issue to track when support will be added however if you want to follow it:
https://github.com/puppetlabs/content-and-tooling-team/issues/10

david22swan:
Not sure what is causing the issue's with the AlmaLinux/Rocky/CentOS 8 tests here but it seems to be consistent.

ekohl:
Yes, but they have existed for a long time I think. Perhaps they commonly failed in puppetlabs-apache, but I recall them timing out very often in some module(s).

david22swan:
Have rekicked the tests with extra debugging

david22swan:
@ekohl 
Have tried looking into a cause for this and found none, its replicating on the main branch so gonna go ahead and merge your pr.
Thanks for putting in the work and sorry for the wait.

david22swan (approved):
LGTM

PR #1306, opened 2021-09-29
MODULES-11201: add service_name for Ubuntu 18.04 and later

moritz-makandra:
This will set `$service_status` for Ubuntu hosts

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -184,6 +184,8 @@ class postgresql::params inherits postgresql::globals {
       if $facts['os']['name'] == 'Debian' and versioncmp($facts['os']['release']['major'], '8') >= 0 {
         # Jessie uses systemd
         $service_status = pick($service_status, "/usr/sbin/service ${service_name}@*-main status")
+      } elsif $facts['os']['name'] == 'Ubuntu' and versioncmp($facts['os']['release']['major'], '18.04') >= 0 {
+        $service_status = pick($service_status, "/usr/sbin/service ${service_name}@*-main status")
       } elsif $facts['os']['name'] == 'Ubuntu' and versioncmp($facts['os']['release']['major'], '15.04') >= 0 {
         # Ubuntu releases since vivid use systemd
         $service_status = pick($service_status, "/usr/sbin/service ${service_name} status")
--- a/spec/unit/classes/server/service_spec.rb
+++ b/spec/unit/classes/server/service_spec.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'postgresql::server::service', type: :class do
+  let :pre_condition do
+    'include postgresql::server'
+  end
+
+  let :facts do
+    {
+      os: {
+        family: 'Debian',
+        name: 'Ubuntu',
+        release: { 'full' => '18.04', 'major' => '18.04' },
+      },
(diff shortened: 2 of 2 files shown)


spec/unit/classes/server/service_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  let :pre_condition do
+    'include postgresql::server'
+  end
+
+  let :node_params do
+    {
+      'lsbmajdistrelease' => '18.04',
+    }
+  end
```
ekohl, line 16: Where it this used?

spec/unit/classes/server/service_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+        family: 'Debian',
+        name: 'Ubuntu',
+        release: { 'full' => '18.04', 'major' => '18.04' },
+      },
+    }
+  end
+
+  it {
+    is_expected.to contain_class('postgresql::server::service')
+    is_expected.to contain_service('postgresqld').with_name('postgresql')
```
ekohl, line 30: This looks like it would also pass prior to this change which means it isn't a good test.

manifests/params.pp
```diff
@@ -209,6 +209,8 @@
       } elsif $facts['os']['name'] == 'Ubuntu' and versioncmp($facts['os']['release']['major'], '15.04') >= 0 {
         # Ubuntu releases since vivid use systemd
         $service_status = pick($service_status, "/usr/sbin/service ${service_name} status")
+      } elsif $facts['os']['name'] == 'Ubuntu' and versioncmp($facts['os']['release']['major'], '18.04') >= 0 {
+        $service_status = pick($service_status, "/usr/sbin/service ${service_name}@*-main status")
```
ekohl, line 213: This will never be executed. It must be before the `15.04` block because `18.04` is already greater than `15.04`.

ekohl (changes requested):

spec/unit/classes/server/service_spec.rb
moritz-makandra: This is used in `params.pp` Line 168 to determine `$service_name`. The test will fail when the variable is empty.

spec/unit/classes/server/service_spec.rb
moritz-makandra: I added service status to the test

manifests/params.pp
moritz-makandra: Well spotted

spec/unit/classes/server/service_spec.rb
ekohl: Still, shouldn't that be a fact? I've submitted https://github.com/puppetlabs/puppetlabs-postgresql/pull/1307 for that to be dropped though so I'd prefer to go that route and then rebase this.

spec/unit/classes/server/service_spec.rb
ekohl: Could you chain it so there's a single `contain_service('postgresqld')`? I also think it's a best practice to use a single `is_expected.to` within `it {}`.

moritz-makandra:
> My PR has been merged. Could you rebase this? After that I think this should be good to merge.

I did the rebase.

david22swan (approved):

PR #1309, opened 2021-11-01
debian package postgresql-contrib virtual package

carlosduelo:
Debian family provides postgresql-contriub-XY as a virtual package of postgresql-XY.

Not sure if allowing virtual packages here is the best solution :thinking: 

Referencies:
https://packages.debian.org/bullseye/postgresql-contrib-13
https://packages.ubuntu.com/bionic/postgresql-contrib

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -135,8 +135,6 @@ class postgresql::params inherits postgresql::globals {
       $java_package_name      = pick($java_package_name, 'postgresql-jdbc')
       # Archlinux doesn't have develop packages
       $devel_package_name     = pick($devel_package_name, 'postgresql-devel')
-      # Archlinux does have postgresql-contrib but it isn't maintained
-      $contrib_package_name   = pick($contrib_package_name,'undef')
       # Archlinux postgresql package provides plperl
       $plperl_package_name    = pick($plperl_package_name, 'undef')
       $plpython_package_name  = pick($plpython_package_name, 'undef')
@@ -163,7 +161,6 @@ class postgresql::params inherits postgresql::globals {
 
       $client_package_name    = pick($client_package_name, "postgresql-client-${version}")
       $server_package_name    = pick($server_package_name, "postgresql-${version}")
-      $contrib_package_name = pick($contrib_package_name, "postgresql-contrib-${version}")
       if $postgis_version and versioncmp($postgis_version, '2') < 0 {
         $postgis_package_name = pick($postgis_package_name, "postgresql-${version}-postgis")
       } elsif $postgis_version and versioncmp($postgis_version, '3') >= 0 {
@@ -203,7 +200,6 @@ class postgresql::params inherits postgresql::globals {
 
       $client_package_name  = pick($client_package_name, 'UNSET')
       $server_package_name  = pick($server_package_name, 'postgresql')
-      $contrib_package_name = pick_default($contrib_package_name, undef)
       $devel_package_name   = pick_default($devel_package_name, undef)
       $java_package_name    = pick($java_package_name, 'jdbc-postgresql')
       $perl_package_name    = pick($perl_package_name, 'DBD-Pg')
--- a/manifests/server/contrib.pp
+++ b/manifests/server/contrib.pp
@@ -5,22 +5,20 @@
 # @param package_ensure
 #   Ensure the contrib package is installed.
 class postgresql::server::contrib (
-  String $package_name      = $postgresql::params::contrib_package_name,
-  String[1] $package_ensure = 'present'
+  Optional[String[1]] $package_name = $postgresql::params::contrib_package_name,
+  String[1] $package_ensure         = 'present'
 ) inherits postgresql::params {
-  if $facts['os']['family'] == 'Gentoo' {
-    fail('osfamily Gentoo does not have a separate "contrib" package, postgresql::server::contrib is not supported.')
-  }
+  if $package_name {
+    package { 'postgresql-contrib':
+      ensure => $package_ensure,
+      name   => $package_name,
+      tag    => 'puppetlabs-postgresql',
+    }
 
-  package { 'postgresql-contrib':
-    ensure => $package_ensure,
-    name   => $package_name,
-    tag    => 'puppetlabs-postgresql',
+    anchor { 'postgresql::server::contrib::start': }
+    -> Class['postgresql::server::install']
+    -> Package['postgresql-contrib']
+    -> Class['postgresql::server::service']
+    anchor { 'postgresql::server::contrib::end': }
   }
-
-  anchor { 'postgresql::server::contrib::start': }
-  -> Class['postgresql::server::install']
-  -> Package['postgresql-contrib']
-  -> Class['postgresql::server::service']
-  anchor { 'postgresql::server::contrib::end': }
 }
--- a/spec/unit/classes/server/contrib_spec.rb
+++ b/spec/unit/classes/server/contrib_spec.rb
@@ -43,5 +43,4 @@ describe 'postgresql::server::contrib', type: :class do
       is_expected.to contain_package('postgresql-contrib').with(tag: 'puppetlabs-postgresql')
     end
   end
-
 end
--- a/spec/unit/classes/server/contrib_spec.rb
+++ b/spec/unit/classes/server/contrib_spec.rb
@@ -43,4 +43,37 @@ describe 'postgresql::server::contrib', type: :class do
       is_expected.to contain_package('postgresql-contrib').with(tag: 'puppetlabs-postgresql')
     end
   end
+
+  describe 'on Gentoo' do
+    let :facts do
+      {
+        os: {
+          family: 'Gentoo',
+          name: 'Gentoo',
+        },
+      }
+    end
+
+    it 'postgresql-contrib should not be installed' do
+      is_expected.to compile
+      is_expected.not_to contain_package('postgresql-contrib')
+    end
+  end
+
+  describe 'on Debian' do
+    let :facts do
+      {
+        os: {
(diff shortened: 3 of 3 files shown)


manifests/params.pp
```diff
@@ -161,26 +164,28 @@
...
       $service_name = pick($service_name, 'postgresql')
 
-      $client_package_name    = pick($client_package_name, "postgresql-client-${version}")
-      $server_package_name    = pick($server_package_name, "postgresql-${version}")
-      $contrib_package_name = pick($contrib_package_name, "postgresql-contrib-${version}")
+      $client_package_name     = pick($client_package_name, "postgresql-client-${version}")
+      $server_package_name     = pick($server_package_name, "postgresql-${version}")
+      # debian package postgresql-${version} provides virtual package postgresql-contrib-${version}
+      $contrib_package_name    = pick($contrib_package_name, "postgresql-contrib-${version}")
+      $contrib_virtual_package = pick($contrib_virtual_package, 'yes')
```
ekohl, line 171: Is `'yes'` a valid boolean? I'd expect `true`

ekohl (commented):
If it's a virtual, I wonder if a solution that also works for Gentoo (which also doesn't have a real package) would be better. Like making `contrib_package_name` an `Optional[String[1]]` and if it's `undef` then don't install it.

I'm not sure that's the best solution either. What are your thoughts on it?

manifests/params.pp
carlosduelo: I tried your point here 6d9958a

this is what you meant?

manifests/params.pp
```diff
@@ -42,44 +42,46 @@
...
-        $docs_package_name      = pick($docs_package_name, 'postgresql-docs')
-        $plperl_package_name    = pick($plperl_package_name, 'postgresql-plperl')
-        $plpython_package_name  = pick($plpython_package_name, 'postgresql-plpython')
-        $service_name           = pick($service_name, 'postgresql')
-        $bindir                 = pick($bindir, '/usr/bin')
-        $datadir                = $facts['os']['name'] ? {
+        $client_package_name     = pick($client_package_name, 'postgresql')
+        $server_package_name     = pick($server_package_name, 'postgresql-server')
+        $contrib_package_name    = pick($contrib_package_name,'postgresql-contrib')
+        $contrib_virtual_package = pick($contrib_virtual_package, undef)
```
ekohl, line 48: I recall that `pick()` has some problems with `pick($x, undef)`. I'm not sure this is valid.

Also, here it's guaranteed to have a value since `$contrib_package_name` is guaranteed to be a `String` at this point since `pick($x, 'MyString')` guarantees that.
```suggestion
```

manifests/params.pp
```diff
@@ -161,26 +164,28 @@
...
       $service_name = pick($service_name, 'postgresql')
 
-      $client_package_name    = pick($client_package_name, "postgresql-client-${version}")
-      $server_package_name    = pick($server_package_name, "postgresql-${version}")
-      $contrib_package_name = pick($contrib_package_name, "postgresql-contrib-${version}")
+      $client_package_name     = pick($client_package_name, "postgresql-client-${version}")
+      $server_package_name     = pick($server_package_name, "postgresql-${version}")
+      # debian package postgresql-${version} provides virtual package postgresql-contrib-${version}
+      $contrib_package_name    = pick($contrib_package_name, "postgresql-contrib-${version}")
+      $contrib_virtual_package = pick($contrib_virtual_package, true)
```
ekohl, line 171: That means you can simply leave these two out
```suggestion
```

manifests/params.pp
```diff
@@ -130,27 +132,28 @@
...
+      $client_package_name     = pick($client_package_name, 'postgresql')
+      $server_package_name     = pick($server_package_name, 'postgresql-libs')
+      $java_package_name       = pick($java_package_name, 'postgresql-jdbc')
       # Archlinux doesn't have develop packages
-      $devel_package_name     = pick($devel_package_name, 'postgresql-devel')
+      $devel_package_name      = pick($devel_package_name, 'postgresql-devel')
       # Archlinux does have postgresql-contrib but it isn't maintained
-      $contrib_package_name   = pick($contrib_package_name,'undef')
+      $contrib_package_name    = pick($contrib_package_name, undef)
+      $contrib_virtual_package = pick($contrib_virtual_package, undef)
```
ekohl, line 142: These can be left out
```suggestion
```

manifests/params.pp
```diff
@@ -42,44 +42,46 @@
...
-        $devel_package_name     = pick($devel_package_name, "postgresql${package_version}-devel")
-        $java_package_name      = pick($java_package_name, "postgresql${package_version}-jdbc")
-        $docs_package_name      = pick($docs_package_name, "postgresql${package_version}-docs")
-        $plperl_package_name    = pick($plperl_package_name, "postgresql${package_version}-plperl")
-        $plpython_package_name  = pick($plpython_package_name, "postgresql${package_version}-plpython")
-        $service_name           = $facts['os']['name'] ? {
+        $client_package_name     = pick($client_package_name, "postgresql${package_version}")
+        $server_package_name     = pick($server_package_name, "postgresql${package_version}-server")
+        $contrib_package_name    = pick($contrib_package_name,"postgresql${package_version}-contrib")
+        $contrib_virtual_package = pick($contrib_virtual_package, undef)
```
ekohl, line 65:
```suggestion
```

manifests/server/contrib.pp
```diff
@@ -5,17 +5,17 @@
...
+  if $package_name != undef {
+    package { 'postgresql-contrib':
+      ensure        => $package_ensure,
+      name          => $package_name,
+      allow_virtual => $allow_virtual,
+      tag           => 'puppetlabs-postgresql',
+    }
   }
 
   anchor { 'postgresql::server::contrib::start': }
```
ekohl, line 21: These anchors will fail since `Package['postgresql-contrib']` will not exist.

manifests/server/contrib.pp
```diff
@@ -5,17 +5,17 @@
...
 ) inherits postgresql::params {
-  if $facts['os']['family'] == 'Gentoo' {
-    fail('osfamily Gentoo does not have a separate "contrib" package, postgresql::server::contrib is not supported.')
-  }
-
-  package { 'postgresql-contrib':
-    ensure => $package_ensure,
-    name   => $package_name,
-    tag    => 'puppetlabs-postgresql',
+  if $package_name != undef {
```
ekohl, line 12: This is sufficient:
```suggestion
  if $package_name {
```

manifests/server/contrib.pp
```diff
@@ -5,17 +5,17 @@
 # @param package_ensure
 #   Ensure the contrib package is installed.
 class postgresql::server::contrib (
-  String $package_name      = $postgresql::params::contrib_package_name,
-  String[1] $package_ensure = 'present'
+  Optional[String[1]] $package_name = $postgresql::params::contrib_package_name,
+  String[1] $package_ensure         = 'present'
+  Optional[Boolean] $allow_virtual  = $postgresql::params::contrib_virtual_package,
```
ekohl, line 10: I don't think this is needed anymore. If you do keep it in, please have them match. So `$postgresql::params::contrib_virtual_package` should be named `$postgresql::params::contrib_allow_virtual_package`. Another is this:
```suggestion
  Optional[Boolean] $allow_virtual  = undef,
```

If you do keep it, it also needs a puppet-strings `@param` documentation tag.

ekohl (changes requested):
I think that a `$contrib_virtual_package` is not needed at all. Instead, what I meant was to make `$contrib_package_name` itself default to `undef` if there is no (real) contrib package.

manifests/server/contrib.pp
carlosduelo: Thanks for your help and guidance 

I tried here :) 0e5cb49ae7b3444cd5552b0a1ebcb7990ede7154

manifests/server/contrib.pp
```diff
@@ -8,19 +8,17 @@
   String $package_name      = $postgresql::params::contrib_package_name,
```
ekohl, line 8: This should be `Optional[String[1]]` so `undef` is accepted.
```suggestion
  Optional[String[1]] $package_name = $postgresql::params::contrib_package_name,
```

carlosduelo: Fixed f57f823

spec/unit/classes/server/contrib_spec.rb
```diff
@@ -40,21 +43,4 @@
...
-          name: 'Gentoo',
-        },
-      }
-    end
-
-    it 'fails to compile' do
-      expect {
-        is_expected.to compile
-      }.to raise_error(%r{is not supported})
-    end
```
ekohl, line 58: Can you modify this to test the package is not defined (which is the new implementation)? You could do the same for Debian.
```ruby
it { is_expected.to compile }
it { is_expected.not_to contain_package('postgresql-contrib')
```

ekohl (commented):
This is indeed the implementation I suggested. Minor testing suggestion, but otherwise :+1:

spec/unit/classes/server/contrib_spec.rb
carlosduelo: Done 6d7ca60

ekohl (approved):
Code wise this looks good to me. I started the tests and if they're green this should be merged. Thanks!

carlosduelo:
> Code wise this looks good to me. I started the tests and if they're green this should be merged. Thanks!

@ekohl can you help me? I don't know why some of them where canceled :(

ekohl:
Looks like they timed out but I think it's safe to ignore those so I'm going to go ahead and merge this. Thanks @carlosduelo!

PR #1310, opened 2021-11-04
Allow systemd to mask postgresql service file

kim-sondrup:
When masking the postgresql service with systemd a symlink pointing to `/dev/null` will be created at `/etc/systemd/system/${service_name}.service`

This pull request is to ensure Postgresql Puppet module to not try to undo the masking by removing the symlink.


ekohl (approved):
This looks ok to me. I've started CI.

PR #1315, opened 2021-12-01
Fix table grant with schema

vaol:
There was an issue using `postgresql::server::grant::onlyif_exists = true` and specifying a schema `postgresql::server::grant::object_name = ['myschema', 'mytable']` as the onlyif check was including the schema in the query in pg_* views.

### DIFF
--- a/manifests/server/grant.pp
+++ b/manifests/server/grant.pp
@@ -425,10 +425,8 @@ define postgresql::server::grant (
   # }
   case $_object_name {
     Array:   {
-      $_togrant_object = $_enquote_object ? {
-        false   => join($_object_name, '.'),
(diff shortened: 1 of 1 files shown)


vaol:
Hello, yes this is still important as it fixes an issue. Thanks for reviewing !

ekohl:
The EL 8  test failures can be ignored (they also fail without this PR), but I do wonder if you can come up with a test that shows the new behavior.

vaol:
Hello, thanks for the note. I just added a test on this specific case. Thank you for reviewing.

david22swan (approved):
LGTM

david22swan:
@vaol Thanks for putting in the work :)

PR #1316, opened 2021-12-05
Fix python package name in RHEL/CentOS 8

kajinamit:
In RHEL/CentOS8, python3 is used globally instead of python2. Because
of that python-* package is no longer available and most of python
packages are named python3-*.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -109,7 +109,11 @@ class postgresql::params inherits postgresql::globals {
       $psql_path           = pick($psql_path, "${bindir}/psql")
 
       $perl_package_name   = pick($perl_package_name, 'perl-DBD-Pg')
-      $python_package_name = pick($python_package_name, 'python-psycopg2')
+      if $facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'], '8') >= 0 {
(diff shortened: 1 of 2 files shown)


tobias-urdin:
Probably do release major >= 8 or the same issue will be present in 9

kajinamit:
> Probably do release major >= 8 or the same issue will be present in 9

That is definitely better. I updated the change accordingly.

manifests/params.pp
```diff
@@ -109,7 +109,11 @@
       $psql_path           = pick($psql_path, "${bindir}/psql")
 
       $perl_package_name   = pick($perl_package_name, 'perl-DBD-Pg')
-      $python_package_name = pick($python_package_name, 'python-psycopg2')
+      if $facts['os']['family'] == 'RedHat' and versioncmp($facts['os']['release']['major'], '8') >= 0 {
```
ekohl, line 112: Just thinking out loud: this will also cover Fedora. It will break Fedora versions which are still primarily Python 2 but those are EOL anyway. Probably not a big deal.

ekohl (approved):
This looks good to me. I started the tests.

smortex (approved):

ekohl:
Test failures look like timeouts, ignoring those.

PR #1321, opened 2022-02-05
Initial support of CentOS 9 Stream

kajinamit:
This change is the initial implementation to support CentOS 9 Stream.


ekohl (commented):
I already submitted https://github.com/puppetlabs/puppetlabs-postgresql/pull/1303. I think that's more complete but I haven't had a chance to verify it yet.

kajinamit:
@ekohl Thanks for that information. I'll close this PR because there already exists a PR for the same purpose.

PR #1324, opened 2022-02-08
pdksync - (IAC-1787) Remove Support for CentOS 6

david22swan:
(IAC-1787) Remove Support for CentOS 6
pdk version: `2.2.0`

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -33,7 +33,6 @@
     {
       "operatingsystem": "CentOS",
       "operatingsystemrelease": [
-        "6",
         "7",
         "8"
       ]


smortex (approved):
Seems safe since we only stop saying the module is supporting this OS.

ekohl:
I think this is OK. I guess it is intentional that RHEL 6 stays since that's not EOL. Is that OK?

ekohl:
I'm very well aware of the CentOS 6 issues and I'm surprised it took you this long.

And you're completely right that my underlying question was: how much can we drop. Looking at the code it suggests RHEL 6 and SLES 11 are the last Apache 2.2 systems and cleaning that up will massively simplify the codebase.

PR #1332, opened 2022-03-23
README.md: correct postgresql_conn_validator example

bastelfreak:
it's required to set the psql_path parameter. Otherwise you end up with
something like:
```
Debug: Executing: '--tuples-only --quiet --no-psqlrc --host 127.0.0.1
--port 5432 --username grafana --dbname grafana --command SELECT 1'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,8 +321,9 @@ postgresql_conn_validator { 'validate my postgres connection':
   db_username       => 'mydbuser',
   db_password       => 'mydbpassword',
   db_name           => 'mydbname',
-}->
-exec { 'rake db:migrate':
+  psql_path         => '/usr/bin/psql',
+}
+-> exec { 'rake db:migrate':
   cwd => '/opt/myrubyapp',
 }
 ```


kenyon (approved):

smortex (approved):
A valid default value would be better, but at least show that it does not exist!  LGTM!

bastelfreak:
@smortex I totally agree. I was really surprised to find out that there isn't. Maybe there was one in the past :thinking:

ekohl:
Doesn't look like it ever had a default.
https://github.com/puppetlabs/puppetlabs-postgresql/commit/c37b50c5d500c6fd6f684c50f936aa84195c970d
I wouldn't mind one, or it could be modified to use `$PATH` by default.

kenyon:
Bad @github-actions bot.

david22swan:
@bastelfreak Apologies on the wait but this LGTM so have gone ahead and merged.
Thank's for your contribution :)

PR #1338, opened 2022-04-12
Use python3 package in Debian/bullseye and above

zulcss:
Debian 11 uses python3 by default so install the python3-psycopg2
for Debian bullseye and above.

Signed-off-by: Charles Short <charles.short@windriver.com>


david22swan:
@zulcss Agree with @bastelfreak above.
Also, the spec test regarding this will need updated to check for the new python package on debian 11

david22swan (changes requested):
@zulcss Agree with @bastelfreak above.
Also, the spec test regarding this will need updated to check for the new python package on debian 11

PR #1341, opened 2022-04-29
Add Ubuntu 22.04 support

evgeni:


evgeni:
Oh, huh, there is no Puppet for 22.04 yet? Well, that's a blocker.

bastelfreak:
We also have no ubuntu 22.04 factsets in https://github.com/voxpupuli/facterdb/pulls yet, so spec tests won't be executed for that OS as well.

PR #1342, opened 2022-05-18
pdksync - (MAINT) Stale-bot config/msg update

LukasAud:


ekohl (commented):
Can't you store these in a reusable workflow? See https://docs.github.com/en/actions/learn-github-actions/reusing-workflows as well. In https://github.com/voxpupuli/gha-puppet VP uses this with great success.

PR #1346, opened 2022-06-21
Fix error when creating roles on AWS RDS

hackdna:
More details: https://tickets.puppetlabs.com/browse/MODULES-5068


manifests/server/role.pp
```diff
@@ -147,7 +147,7 @@
       }
       postgresql_psql { "ALTER ROLE ${username} ENCRYPTED PASSWORD ****":
         command   => Sensitive("ALTER ROLE \"${username}\" ENCRYPTED PASSWORD '${pwd_hash_sql}'"),
-        unless    => Sensitive("SELECT 1 FROM pg_shadow WHERE usename = '${username}' AND passwd = '${pwd_hash_sql}'"),
+        unless    => Sensitive("SELECT 1 FROM pg_user WHERE usename = '${username}' AND passwd = '${pwd_hash_sql}'"),
```
smortex, line 150: Given `pg_catalog.pg_user` is defined as:

```
 SELECT pg_shadow.usename,
    pg_shadow.usesysid,
    pg_shadow.usecreatedb,
    pg_shadow.usesuper,
    pg_shadow.userepl,
    pg_shadow.usebypassrls,
    '********'::text AS passwd,
    pg_shadow.valuntil,
    pg_shadow.useconfig
   FROM pg_shadow;
```

passwd will never match.

hackdna: Thank you! Can we just not check the `passwd` value then:
`SELECT 1 FROM pg_user WHERE usename = '${username}'`

pmcmaw:
Removing needs attention label, apologies we haven't got around to triaging and reviewing this PR.

pmcmaw:
Would you mind rebasing @smortex, im going to add it to a priority list to get picked up on our next community day.

hackdna:
No worries. Looking forward to getting this 5 year old issue finally addressed.

smortex:
> Would you mind rebasing @smortex, im going to add it to a priority list to get picked up on our next community day.

Poke @hackdna as they own the PR.

manifests/server/role.pp
smortex: How can you check the password is conform to expectation without checking the password? In an idempotent way :open_mouth: ?

smortex: There are probably AWS specificities that allow to do this in such a context… But i am not a user of AWS and can't give pointers unfortunately.

smortex: Sure, but `$update_password` defaults to true (and even if it was false by default) and updating the password at each puppet run does not make sense.

Regarding `pg_shadow` deprecation, you are :100:. `pg_shadow` being defined as:
```
 SELECT pg_authid.rolname AS usename,
    pg_authid.oid AS usesysid,
    pg_authid.rolcreatedb AS usecreatedb,
    pg_authid.rolsuper AS usesuper,
    pg_authid.rolreplication AS userepl,
    pg_authid.rolbypassrls AS usebypassrls,
    pg_authid.rolpassword AS passwd,
    pg_authid.rolvaliduntil AS valuntil,
    s.setconfig AS useconfig
   FROM pg_authid
     LEFT JOIN pg_db_role_setting s ON pg_authid.oid = s.setrole AND s.setdatabase = 0::oid
  WHERE pg_authid.rolcanlogin;
```

the idempotent way to update the query would be:

```suggestion
        unless    => Sensitive("SELECT 1 FROM pg_authid WHERE rolname = '${username}' AND rolpassword = '${pwd_hash_sql}'"),
```

No access to an "AWS RDS" systems, so can't tell if it would solve the root issue.  If it does this is cool, otherwise you need to find a suitable query.

hackdna: Thank you for the suggestion. Unfortunately, there is no access to `pg_authid` either. Yes, updating the password on each Puppet run is not clean but it is idempotent. Another way to address this is to set `$update_password` to `false` by default or to expose as a parameter to `postgresql::server::db`. Thoughts?

GSPatton:
hey @hackdna, we're looking into this. Could you rebase and resolve any conflicts? Thanks!

hackdna:
Great, thanks for looking into this. I think I've resolved the merge conflict.

GSPatton:
Hey @hackdna . pg_roles is also a deprecated view, so replacing instances of one deprecated view with another is not a satisfactory fix unfortunately. I advise you research other potential fixes and create another PR when you've found something and we'll happily take a look.

Thank you for your contributions!

hackdna:
Hey @GSPatton, I've used `pg_roles` because it's already used extensively in `role.pp`. Also, another "legacy" view `pg_shadow` was used in a recently merged PR: https://github.com/puppetlabs/puppetlabs-postgresql/pull/1347/files. How is this case different?

PR #1349, opened 2022-07-04
Fix service status detection on Debian-based OSes

arjenz:
`/usr/sbin/service postgresql@*-main status` on at least Ubuntu 18.04, 20.04 and 22.04, always exits 0 (even when service is down), so puppet always believes the service is up (which both causes all kinds of failures and prevents puppet from starting the stopped service).

Not sure why all the convoluted logic in this bit is needed, but this fixes it for me on 18 and 20 (and also 22 but that's not yet supported by this module).

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -166,18 +166,13 @@ class postgresql::params inherits postgresql::globals {
       $bindir                 = pick($bindir, "/usr/lib/postgresql/${version}/bin")
       $datadir                = pick($datadir, "/var/lib/postgresql/${version}/main")
       $confdir                = pick($confdir, "/etc/postgresql/${version}/main")
-      if $facts['os']['name'] == 'Debian' and versioncmp($facts['os']['release']['major'], '8') >= 0 {
-        # Jessie uses systemd
-        $service_status = pick($service_status, "/usr/sbin/service ${service_name}@*-main status")
-      } elsif $facts['os']['name'] == 'Ubuntu' and versioncmp($facts['os']['release']['major'], '18.04') >= 0 {
-        $service_status = pick($service_status, "/usr/sbin/service ${service_name}@*-main status")
-      } elsif $facts['os']['name'] == 'Ubuntu' and versioncmp($facts['os']['release']['major'], '15.04') >= 0 {
-        # Ubuntu releases since vivid use systemd
-        $service_status = pick($service_status, "/usr/sbin/service ${service_name} status")
+      if pick($service_provider, $facts['service_provider']) == 'systemd' {
+        $service_reload = "systemctl reload ${service_name}"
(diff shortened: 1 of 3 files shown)


arjenz:
I would think it would also work on Debian. After reading all this it wasn't quite clear to me as to why all this logic is there (instead of, as you mentioned, rely on the default, which is working fine) so I didn't dare wiping it, just fixed its behavior for 18 and up.

I'd be all for removing all this :)

ekohl:
> Now I also see that RH-based distros don't respect the `service_status` global parameter, which is a bug in itself.

I opened https://github.com/puppetlabs/puppetlabs-postgresql/pull/1351 for this.

> I'd be all for removing all this :)

I'd be in favor of this. @david22swan any thoughts?

arjenz:
@ekohl @david22swan I rewrote it to match the bits @ekohl added for RHEL, tested (successfully) it on Debian 10/11, Ubuntu {18,20,22}.04. I think people who rely on other service tooling than `service` or `systemctl` can just set it as a parameter.

ekohl (changes requested):
It looks like the unit tests need to be modified to reflect these changes, otherwise :+1:

manifests/params.pp
```diff
@@ -166,18 +166,13 @@
...
-      } elsif $facts['os']['name'] == 'Ubuntu' and versioncmp($facts['os']['release']['major'], '15.04') >= 0 {
-        # Ubuntu releases since vivid use systemd
-        $service_status = pick($service_status, "/usr/sbin/service ${service_name} status")
+      if pick($service_provider, $facts['service_provider']) == 'systemd' {
+        $service_reload = "systemctl reload ${service_name}"
+        $service_status = pick($service_status, "systemctl status ${service_name}")
       } else {
-        $service_status = pick($service_status, "/etc/init.d/${service_name} status | /bin/egrep -q 'Running clusters: .+|online'")
+        $service_reload = "service ${service_name} reload"
+        $service_status = pick($service_status, "service ${service_name} status")
```
ekohl, line 174: Oh, I think these paths need to be absolute.

arjenz: Oh yeah, I cheated a bit, as it moved from `/bin/` to `/usr/bin/` or the other way around, but I think both work anyway (as does omitting the path, but I agree, having full path is better :) ).

ekohl: I'd be OK with that (I've seen that Puppet does apply `$PATH` to it, unlike `exec`), but then the unit tests should match.

ekohl (approved):
Waiting for CI to complete

ekohl (approved):
Thanks! I think SLES is unrelated but I can't tell because https://github.com/puppetlabs/puppetlabs-postgresql/actions/runs/2653307902 excluded SLES.

PR #1350, opened 2022-07-04
Add version for Ubuntu 22.04

arjenz:
Ubuntu 22.04 ships with version 14.

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -208,6 +208,7 @@ class postgresql::globals (
         /^(18.04)$/ => '10',
         /^(20.04)$/ => '12',
         /^(21.04|21.10)$/ => '13',
+        /^(22.04)$/ => '14',
         default => undef,
       },
       default => undef,


ekohl:
I'm not sure if CI already has support for running Ubuntu 22.04, which would be good to have. IIRC the packages do exist now. @david22swan could you answer this?

arjenz:
I don't think there's full support yet, repo service mentions `jammy`, but I think it was pretty empty still (I installed `puppet-agent_6.27.1-1focal_amd64.deb` manually to test, "it just works"). 

I explicitly also chose to just fill the variable (and not touch any metadata referring to 22.04), as with the variable filled it seems to "just work", which I guess is the first step towards "Is supported".

ekohl (approved):
I'd be fine with merging this partial support, but adding official support would be nice.

david22swan:
@ekohl @arjenz Ubuntu 22.04 support has just recently been added to the Agent nightlies, but not yet the official release, so it is in the pipeline.
There are a few other tasks that precede it but once they are done we will be moving onto it and getting support out.

In regards to this change, it LGTM so I'm happy to get some compatibility in prior to the official support getting added.
Will put it in as a bugfix though to indicate it's not an outright addition.

PR #1353, opened 2022-07-20
(GH-1352) - Updating postgresql service version on SLES 

### LINKED ISSUE #1352: SLES Acceptance test failures
We have some persistent failures with the SLES acceptance tests.

image

@tvpartytonight has identified that they are failing because the the zypper provider does not have the repo necessary for postgres package installation.

``` bash
Error: Execution of '/usr/bin/zypper --quiet install --auto-agree-with-licenses --no-confirm postgresql10-server' returned 104: No provider of 'postgresql10-server' found.
```

pmcmaw:
Fixes #1352 
Updating postgresql service package version to be inline with https://zypp.postgresql.org/howtozypp/


ekohl:
When I experiment I sometimes drop all unrelated OSes from `metadata.json` to speed up CI/waste fewer resources.

ekohl:
In that case, welcome back :) And if you're in Europe, you can also blame it on the temperature.

chelnak (approved):

manifests/globals.pp
```diff
@@ -227,8 +227,8 @@
         /11\.4/     => '94',
         /12\.0/     => '93',
         /12\.[1-3]/ => '94',
-        /12\.[4-5]/ => '10',
-        /15\.[0-9]/     => '10',
+        /12\.[4-5]/ => '12',
+        /15\.[0-9]/ => '14',
         default     => '96',
```
ekohl, line 232: I don't know SLES. Are point releases also supported? Is this really valid for 15.0 as well or was it only introduced in some later 15.x version?

PR #1356, opened 2022-08-09
Only require password when used

arjenz:
If the if in line 39 evaluates to false, the user has been created
elsewhere, and we don't actually use (and thus need) password in
`postgresql::server::db` class.

### DIFF
--- a/manifests/server/db.pp
+++ b/manifests/server/db.pp
@@ -1,7 +1,7 @@
 # @summary Define for conveniently creating a role, database and assigning the correctpermissions.
 #
-# @param user User to create and assign access to the database upon creation. Mandatory.
-# @param password Required Sets the password for the created user.
+# @param user User to assign access to the database upon creation (will be created if not defined elsewhere). Mandatory.
+# @param password Sets the password for the created user (if a user is created).
 # @param comment Defines a comment to be stored about the database using the PostgreSQL COMMENT command.
 # @param dbname Sets the name of the database to be created.
 # @param encoding Overrides the character set during creation of the database.
@@ -13,7 +13,7 @@
 # @param owner Sets a user as the owner of the database.
 define postgresql::server::db (
   $user,
-  Variant[String, Sensitive[String]] $password,
+  Optional[Variant[String, Sensitive[String]]] $password = undef,
   $comment    = undef,
   $dbname     = $title,
   $encoding   = $postgresql::server::encoding,


ekohl (commented):
I wonder about this use case. If you already define the user elsewhere, can't you simply use `postgresql::server::database` in the later places? What additional benefit is there in this defined type?

I see a tablespace relation, which is also in `postgresql::server::database`.

Then there is a `GRANT`. That isn't in `postgresql::server::database`, except for an owner. AFAIK that already implies all privileges so the only edge case is if you want that grant.

Am I missing anything?

arjenz:
Well the major use case from my end is "this is how we happen to use it, and it stopped working when https://github.com/puppetlabs/puppetlabs-postgresql/commit/d878d13f030cdd872cbb04c43ad0f065fec0b923#diff-09acae987abfb6f2de2796ebbbcee8911d8326cf762451972538f4f78984f35c was introduced".
Before that commit setting it to `undef` was fine, but afterwards not anymore.
Looking just at the code in this class, without taking global context into perspective this solution kind of makes sense. "Why require data when it's not actually used or needed for this class".

arjenz:
Oh, and right now, as a workaround, we're passing around `password: 'unused'` in hiera, since the value gets ignored anyway, so I feel solving this in the code is a bit more elegant :)

manifests/server/db.pp
```diff
@@ -37,6 +37,9 @@
   }
 
   if ! defined(Postgresql::Server::Role[$user]) {
+    if ! $password {
+      fail('Please provide password or create a user: if you rely on postgresql::server::db to create a user, a password is needed')
+    }
```
ekohl, line 42: This is technically not really true. Role can create a user without a password.
```suggestion
```

ekohl (commented):
Mostly because it makes the code (even more) compile order dependent. If the inclusion happens in another order, it may fail because no password is provided. Then again, you could already have code where there are 2 different passwords specified and the real password used ends up being determined by the compile order.

Now looking deeper, I see `postgresql::server::role` can create a user without a password. So I think this is valid if you drop the `fail()` part. That should then also be reflected in the `@param` tags.

arjenz:
> 

Done :)

ekohl:
Did you forget to push?

arjenz:
> Did you forget to push?

Yes, yes I did :)

ekohl (approved):

ekohl (approved):
@david22swan this would have been nice to get in the release. However, I can't merge it because of a failed test.

david22swan (approved):
LGTM

Will get it merged in now

PR #1363, opened 2022-08-23
Fix puppet-strings documentation

ekohl:
In 37e5bea4399ab926e6041190347e24d55a08a51c an empty line was added, splitting the documentation block. Removing the empty line creates a single documentation block again.

Fixes: 37e5bea4399ab926e6041190347e24d55a08a51c

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -77,7 +77,7 @@
 # @param roles Specifies a hash from which to generate postgresql::server::role resources.
 # @param config_entries Specifies a hash from which to generate postgresql::server::config_entry resources.
 # @param pg_hba_rules Specifies a hash from which to generate postgresql::server::pg_hba_rule resources.
-
+#
 # @param backup_enable Whether a backup job should be enabled.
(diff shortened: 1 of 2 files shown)


manifests/server.pp
```diff
@@ -77,7 +77,7 @@
 # @param roles Specifies a hash from which to generate postgresql::server::role resources.
 # @param config_entries Specifies a hash from which to generate postgresql::server::config_entry resources.
 # @param pg_hba_rules Specifies a hash from which to generate postgresql::server::pg_hba_rule resources.
-
+#
```
ekohl, line 80: Interestingly enough, this is why we have https://github.com/voxpupuli/puppet-lint-param-docs. However, it's disabled here and somehow doesn't catch it. That's a bug.

pmcmaw: Amazing @ekohl thanks for this find!

ekohl: It's the most common mistake I've seen with puppet-strings. When things disappear, look for empty lines.

pmcmaw (approved):

PR #1366, opened 2022-09-13
(maint) Codebase Hardening

david22swan:
Changes made to ensure that no malformed commands are passed through to the system.
Certain commands were left undivided as the commands did not get correctly interpreted.


manifests/server/config.pp
```diff
@@ -142,8 +142,11 @@
 
     ensure_packages([$package_name])
 
+    $exec_command = ['/usr/sbin/semanage', 'port', '-a', '-t', 'postgresql_port_t', '-p', 'tcp', shell_escape($port)]
+    $exec_unless = ["/usr/sbin/semanage port -l | grep -qw ${shell_escape($port)}"]
     exec { "/usr/sbin/semanage port -a -t postgresql_port_t -p tcp ${port}":
-      unless  => "/usr/sbin/semanage port -l | grep -qw ${port}",
+      command => $exec_command,
+      unless  => $exec_unless,
```
ekohl, line 149: I'm wondering if this is needed. Isn't port a `Stdlib::Port`? If so, the value is already bound to an integer in a specific range by the type system.

manifests/server/config_entry.pp
```diff
@@ -98,10 +98,13 @@
   # is managed for us in postgresql::server::config.
   if $facts['os']['name'] == 'Debian' or $facts['os']['name'] == 'Ubuntu' {
     if $name == 'data_directory' {
+      $stop_command = ['service', shell_escape($postgresql::server::service_name), 'stop']
+      $stop_onlyif = ['service', shell_escape($postgresql::server::service_name), 'status']
+      $stop_unless = ["grep \"data_directory = '${shell_escape($value)}'\" ${shell_escape($postgresql::server::postgresql_conf_path)}"]
```
ekohl, line 103: Why isn't this a proper array?
```suggestion
      $stop_unless = ['grep', "data_directory = '${shell_escape($value)}'", shell_escape($postgresql::server::postgresql_conf_path)]
```

manifests/server/config_entry.pp
```diff
@@ -130,10 +136,13 @@
     } elsif $name == 'data_directory' {
       # We need to force postgresql to stop before updating the data directory
       # otherwise init script breaks
+      $stop_command = ['service', shell_escape($postgresql::server::service_name), 'stop']
+      $stop_onlyif = ['service', shell_escape($postgresql::server::service_name), 'status']
+      $stop_unless = ["grep 'PGDATA=${shell_escape($value)}' /etc/sysconfig/pgsql/postgresql"]
```
ekohl, line 141: Shouldn't this also be an array with multiple values?
```suggestion
      $stop_unless = ['grep', "PGDATA=${shell_escape($value)}", '/etc/sysconfig/pgsql/postgresql']
```

ekohl (commented):
It surprises me that you need to escape values when passing as an array. In most programming languages the array is passed straight to the invocation, without any shell involved. Doesn't Puppet do this natively? If not, why are command arrays even implemented in the first place?

manifests/server/config_entry.pp
david22swan: Separating into an array does not seem to work correctly when used with grep

manifests/server/config_entry.pp
david22swan: As above

manifests/server/config.pp
david22swan: thats a good point

manifests/server/config_entry.pp
david22swan: > command

As to not needing to shell escape, yeh I was a little over enthusiastic

david22swan: Also hit a similar issue with exec command in passwd.pp

ekohl: What kind of issue? Could it actually be double escaping?

In normal programming (Ruby, Python, C, etc) if you pass an array it gets passed straight into the program so no escaping is needed.

manifests/validate_db_connection.pp
```diff
@@ -66,7 +66,7 @@
     default => "PGPASSWORD=${database_password_unsensitive}",
   }
   $cmd = join([$cmd_init, $cmd_host, $cmd_user, $cmd_port, $cmd_dbname], ' ')
-  $validate_cmd = "${validcon_script_path} ${sleep} ${tries} '${cmd}'"
+  $validate_cmd = [$validcon_script_path, $sleep, $tries, join(["'", $cmd, "'"], '')]
```
ekohl, line 69: Now you're basically doing 2 joins. This feels very complicated. Perhaps it's better to keep `$cmd` as an array and then deal with it here?

david22swan: Tried it in one but the problem is that they both need a different join value, i.e. ' ' and ''....
Suppose I could add the ' to the first and last variables, cmd_init and cmd_dbname but that just seems wrong to me

david22swan: Or do they... have been assuming but should double check that

david22swan: Ok yeh, looks like having a space between the ' and the values is fine, atleast according to the tests.
Not sure why I thought it wouldn't be

manifests/server/config_entry.pp
david22swan: Following up, you suggested fix throws this error:
```
Error: /Stage[main]/Postgresql::Server::Config/Postgresql::Server::Config_entry[data_directory]/Exec[postgresql_stop_data_directory]: Could not evaluate: Could not find command 'data_directory'
```

david22swan: Adding the escaped characters thows the same error

ekohl: I think that's because `unless` doesn't implement it in the same way: https://puppet.com/docs/puppet/7/type.html#exec-attribute-unless
You need to pass it as an array within an array. So
```suggestion
      $stop_unless = [['grep', "data_directory = '${value}'", $postgresql::server::postgresql_conf_path]],
```

manifests/server/config.pp
```diff
@@ -142,8 +142,11 @@
 
     ensure_packages([$package_name])
 
+    $exec_command = ['/usr/sbin/semanage', 'port', '-a', '-t', 'postgresql_port_t', '-p', 'tcp', $port]
+    $exec_unless = ["/usr/sbin/semanage port -l | grep -qw ${port}"]
```
ekohl, line 146: Note this will still call a shell. In this case it's safe, but good to remember.

manifests/validate_db_connection.pp
```diff
@@ -86,9 +86,10 @@
   }
 
   $exec_name = "validate postgres connection for ${database_username}@${database_host}:${database_port}/${database_name}"
+  $exec_command = ['echo', "'Unable to connect to defined database using:", join([$cmd, "' && false"], '')]
```
ekohl, line 89: The `&& false` means a shell is used. I'm not sure how it would work, but I think that's just going to fail if you pass it as an array.

Instead, I think the `validate_cmd` should fail in a way that is clear. We control the script so why can't that have a proper echo and exit code?

manifests/server/config_entry.pp
david22swan: Ok yeh, double wrapping looks to have worked

manifests/server/config.pp
david22swan: 👍

manifests/validate_db_connection.pp
david22swan: Have put it back to a single string.
As to the second part, that may be outside the scope for this.

david22swan:
Not sure where these 6 failures are coming from?

manifests/validate_db_connection.pp
ekohl: I wonder if these days you may even be able to implement it as a deferred function.

manifests/server/config.pp
ekohl: It does make me wonder: should this actually just be a string and kept as-is? That does make it more explicit what it is and the diff smaller.

david22swan: thats a fair point, swapping it to a string

manifests/server/config_entry.pp
```diff
@@ -98,10 +98,13 @@
   # is managed for us in postgresql::server::config.
   if $facts['os']['name'] == 'Debian' or $facts['os']['name'] == 'Ubuntu' {
     if $name == 'data_directory' {
+      $stop_command = ['service', $postgresql::server::service_name, 'stop']
+      $stop_onlyif = ['service', $postgresql::server::service_name, 'status']
+      $stop_unless = [['grep', "data_directory = '${shell_escape($value)}'", $postgresql::server::postgresql_conf_path]]
```
ekohl, line 103: I still think `shell_escape` should never be used when you pass it as an array. It's doing something twice: shell_escape is when you pass something through a shell, but passing it as an array doesn't spawn a shell at all.

spec/defines/validate_db_connection_spec.rb
```diff
@@ -30,7 +30,10 @@
     it { is_expected.to contain_postgresql__validate_db_connection('test') }
 
     it 'has proper path for validate command' do
-      is_expected.to contain_exec('validate postgres connection for test@test:5432/test').with(unless: %r{^/usr/local/bin/validate_postgresql_connection.sh\s+\d+})
+      is_expected.to contain_exec('validate postgres connection for test@test:5432/test').with(unless: ['/usr/local/bin/validate_postgresql_connection.sh',
+                                                                                                        4,
+                                                                                                        30,
+                                                                                                        "' /usr/bin/psql --tuples-only --quiet  -h test  -U test  -p 5432  --dbname test  '"])
```
ekohl, line 36: The quote here looks odd. I'd expect:
```suggestion
                                                                                                        "/usr/bin/psql --tuples-only --quiet  -h test  -U test  -p 5432  --dbname test"])
```

david22swan: I had them added onto the end due to it having been done that way in the original code.
Can try without them and see if it errors though

manifests/server/config_entry.pp
david22swan: Have removed the shell_escapes where the resultant string is being passed as an array.
Have also clarified the ones one full string is being passed by removing the array braces around them

spec/defines/validate_db_connection_spec.rb
david22swan: Seems to work fine

manifests/validate_db_connection.pp
david22swan: Maybe, there has been some talk of deferred functions lately.

manifests/server/passwd.pp
```diff
@@ -26,10 +26,11 @@
     #  without specifying a password ('ident' or 'trust' security). This is
     #  the default for pg_hba.conf.
     $escaped = postgresql::postgresql_escape($postgres_password)
+    $exec_command = "${shell_escape($psql_path)}${_dboption} -c \"ALTER ROLE \\\"${shell_escape($user)}\\\" PASSWORD \${NEWPASSWD_ESCAPED}\""
```
ekohl, line 29: So a few things happen here. First of all, `$_dboption` contains `$database` which is also user input. That should be escaped at line 19.

It's also important to realize this must use a shell because it passes the new password as an environment variable to hide the sensitive value. You could change `$_db_option` to an array at line 18 and then:
```suggestion
    $exec_command = [$psql_path] + $_dboption + ['-c', "ALTER ROLE \"${user}\" PASSWORD ${escaped}"]
```

But I don't know if if that properly generates a sensitive string. You'd also lose some clarity in what happens. So for now I'd escape the input in `$_dboption`.

spec/classes/server/config_spec.rb
```diff
@@ -14,7 +14,7 @@
       is_expected.to contain_package('policycoreutils-python').with(ensure: 'installed')
 
       is_expected.to contain_exec('/usr/sbin/semanage port -a -t postgresql_port_t -p tcp 5432')
-        .with(unless: '/usr/sbin/semanage port -l | grep -qw 5432')
+        .with(unless: ['/usr/sbin/semanage port -l | grep -qw 5432'])
```
ekohl, line 17: This was changed back to a string since it uses a shell anyway. Repeated in this file.

spec/classes/server_spec.rb
```diff
@@ -61,7 +61,7 @@ class { 'postgresql::globals':
       is_expected.to contain_postgresql_conn_validator('validate_service_is_running')
     end
     it 'sets postgres password' do
-      is_expected.to contain_exec('set_postgres_postgrespw').with('command' => '/usr/bin/psql -c "ALTER ROLE \"postgres\" PASSWORD ${NEWPASSWD_ESCAPED}"',
+      is_expected.to contain_exec('set_postgres_postgrespw').with('command' => ['/usr/bin/psql -c "ALTER ROLE \"postgres\" PASSWORD ${NEWPASSWD_ESCAPED}"'],
```
ekohl, line 64: I think this should also be a string (and thus unchanged in the patch).

ekohl (changes requested):

manifests/server/passwd.pp
david22swan: `$_dboption` has been escaped.
Have tried changing the full command to an array and it doesn't seem to parse cleanly

spec/classes/server_spec.rb
david22swan: fixed

spec/classes/server/config_spec.rb
david22swan: fixed

david22swan:
Spec test failures replicated on main branch

manifests/server/config_entry.pp
```diff
@@ -111,10 +114,13 @@
       # We need to force postgresql to stop before updating the port
       # because puppet becomes confused and is unable to manage the
       # service appropriately.
+      $stop_command = ['service', $postgresql::server::service_name, 'stop']
+      $stop_onlyif = ['service', $postgresql::server::service_name, 'status']
+      $stop_unless = "grep 'PGPORT=${shell_escape($value)}' /etc/sysconfig/pgsql/postgresql"
```
LukasAud, line 119: isn't this one missing the brackets [[ ]]?

LukasAud (approved):

manifests/server/config_entry.pp
david22swan: No, decided to go without as there is no practical need

david22swan: And as a way to clearly show it hasn't been divided

ekohl: The version without a shell:
```suggestion
      $stop_unless = [['grep', "PGPORT=${value}", '/etc/sysconfig/pgsql/postgresql']]
```

LukasAud: 👍

PR #1374, opened 2022-10-07
Rewrite dependency installation using native Puppet

ekohl:
Rather than hacking around in Ruby, this rewrites it in native Puppet.


david22swan:
Look's like the spec failure is replicated on the master

GSPatton (approved):
this looks good. The failing spec test isn't related to the changes in this PR

PR #1380, opened 2022-12-20
apt::source: configure repo only for current architecture

bastelfreak:
Without this setting, we tell apt to download release information for all architectures apt knows. This includes i386 on amd64 systems. postgresql doesn't publish i386 packages. because of that, it results in a warning during apt update:

N: Skipping acquire of configured file 'main/binary-i386/Packages' as repository 'https://apt.postgresql.org/pub/repos/apt jammy-pgdg InRelease' doesn't support architecture 'i386'

This is fixed with this change.

### DIFF
--- a/manifests/repo/apt_postgresql_org.pp
+++ b/manifests/repo/apt_postgresql_org.pp
@@ -15,14 +15,15 @@ class postgresql::repo::apt_postgresql_org inherits postgresql::repo {
     priority   => 500,
   }
   -> apt::source { 'apt.postgresql.org':
-    location => $_baseurl,
-    release  => "${facts['os']['distro']['codename']}-pgdg",
-    repos    => 'main',
-    key      => {
+    location     => $_baseurl,
+    release      => "${facts['os']['distro']['codename']}-pgdg",
+    repos        => 'main',
+    architecture => $facts['os']['architecture'],
+    key          => {
       id     => 'B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8',
       source => 'https://www.postgresql.org/media/keys/ACCC4CF8.asc',
     },
-    include  => {
+    include      => {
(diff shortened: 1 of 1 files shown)


kenyon (dismissed):

kenyon:
Commit message says nginx, but I think you mean postgresql here.

manifests/repo/apt_postgresql_org.pp
```diff
@@ -15,14 +15,15 @@
...
   }
   -> apt::source { 'apt.postgresql.org':
-    location => $_baseurl,
-    release  => "${facts['os']['distro']['codename']}-pgdg",
-    repos    => 'main',
-    key      => {
+    location     => $_baseurl,
+    release      => "${facts['os']['distro']['codename']}-pgdg",
+    repos        => 'main',
+    architecture =>  $facts['os']['architecture'],
```
smortex, line 21: Hum.  Real world example of a similar issue, just s/PostgreSQL/Apache/: If on an amd64 system you are using PostgreSQL with a proprietary lib which is only available for i386, you would add i386 architecture to your system (dpkg --add-architecture i386), install postgresql in its i386 flavor and be happy.

Not sure Puppet can handle installing from a specific arch that being said, but this seems perfectly valid.

bastelfreak: all this architecture mix is ugly :( and the upstream postgres repo doesnt even have i386 anymore.

bastelfreak:
@kenyon ah right. I fixed it in our nginx module and copied the text. fixed it.

kenyon:
@bastelfreak and PR title :)

bastelfreak:
@kenyon apt::source is the defined resource that's changed.

chelnak (approved):

PR #1382, opened 2023-01-11
Add multi instance support, refactoring config.pp (1/x)

SimonHoenscheid:
This commit is the first of many to add multi instance support to this module. The general idea is to first copy all classes which are used and create defines from them. These classes will use the defines as is. Necessary changes for the instances itself will be added to the classes and defined types at a later point. This ensures, the module will work as it does right now and there are no breaking changes.

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -1,267 +1,32 @@
 # @api private
 class postgresql::server::config {
-  $ip_mask_deny_postgres_user   = $postgresql::server::ip_mask_deny_postgres_user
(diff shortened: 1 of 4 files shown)


SimonHoenscheid:
I will add a spec test for the defined ressource and copy parameter docs from the server class to the defined ressource if available.

kenyon:
Typo in PR and commit title, "instance".

chelnak:
There seems to be a bigger issue with the acceptance tests causing hangs. Happy to merge this.

chelnak (approved):

PR #1386, opened 2023-01-24
fix concurrent_ruby gem install error

SimonHoenscheid:


ekohl:
This hits all code out there that uses puppet as a gem. I'd much rather have them release an updated gem with a stricter dependency or compatibility with 1.2.0.

bastelfreak:
Puppet 6.29.0 and 7.22.0 got released which pin the concurrent-ruby version, so this change shouldn't be required anymore.

PR #1388, opened 2023-01-26
Ubuntu 14/16/17: Drop code leftovers

bastelfreak:
This was previously done in
https://github.com/puppetlabs/puppetlabs-postgresql/pull/1331.


chelnak (approved):

ekohl (commented):
Similarly, https://endoflife.software/operating-systems/unix-like-bsd/openbsd says we can also clean that up. And https://endoflife.date/sles also says SLES 11 is EOL. I never bothered with these cleanups since it's barely any overhead to keep it in. It'll start getting interesting if you can actually drop code to support older PostgreSQL. Might be interesting to see if there's any PG < 9.3 code for example.

bastelfreak:
yep. I want to do some more cleanups. At the moment I was focused on stuff that's already deprecated, so we can drop it without a major release.

bastelfreak:
some of the nightlys tend to timeout. all important tests succeeded.

PR #1389, opened 2023-01-27
Add multi instance support, refactoring service.pp (4/x)

SimonHoenscheid:
This commit adds changes a class to add multi instance support to this module. The general idea is to first copy all classes which are used and create defines from them. These classes will use the defines as is. Necessary changes for the instances itself will be added to the classes and defined types at a later point. This ensures, the module will work as it does right now and there are no breaking changes.

### DIFF
--- a/manifests/server/instance_service.pp
+++ b/manifests/server/instance_service.pp
@@ -1,0 +1,62 @@
+# lint:ignore:140chars
+# @param service_ensure Ensure service is installed
+# @param service_enable Enable the PostgreSQL service
+# @param service_manage Defines whether or not Puppet should manage the service.
+# @param service_name Overrides the default PostgreSQL service name.
+# @param service_provider Overrides the default PostgreSQL service provider.
+# @param service_status Overrides the default status check command for your PostgreSQL service.
+# @param user Overrides the default PostgreSQL super user and owner of PostgreSQL related files in the file system.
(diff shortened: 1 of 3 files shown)


manifests/server/instance_service.pp
```diff
@@ -0,0 +1,62 @@
...
+  $service_name     = $postgresql::server::service_name,
+  $service_provider = $postgresql::server::service_provider,
+  $service_status   = $postgresql::server::service_status,
+  $user             = $postgresql::server::user,
+  $port             = $postgresql::server::port,
+  $default_database = $postgresql::server::default_database,
+  $psql_path        = $postgresql::server::psql_path,
+  $connect_settings = $postgresql::server::default_connect_settings,
+) {
+  anchor { 'postgresql::server::service::begin': }
```
bastelfreak, line 28: might be a good idea to add a spec test with two instances of postgresql::server::instance_service. I think the anchor will cause some problems because its name needs to be unique. It should probably stay in the postgresql::server::service class? Maybe worth checking which other classes depend on it

SimonHoenscheid: I was thinking about doing this, as soon I make other necessary changes to the define. Currently the define would break with a duplicate declaration warning for the service. I already tried the changes locally and it needs a few updates of spec tests anyhow.

chelnak (approved):

PR #1391, opened 2023-01-27
Add multi instance support, refactoring password.pp (5/x)

SimonHoenscheid:
This commit adds changes a class to add multi instance support to this module. The general idea is to first copy all classes which are used and create defines from them. These classes will use the defines as is. Necessary changes for the instances itself will be added to the classes and defined types at a later point. This ensures, the module will work as it does right now and there are no breaking changes.

### DIFF
--- a/manifests/server/instance_passwd.pp
+++ b/manifests/server/instance_passwd.pp
@@ -1,0 +1,63 @@
+# lint:ignore:140chars
+# @param user Overrides the default PostgreSQL super user and owner of PostgreSQL related files in the file system.
+# @param group Overrides the default postgres user group to be used for related files in the file system.
+#   Default value: 5432. Meaning the Postgres server listens on TCP port 5432.
+# @param psql_path Specifies the path to the psql command.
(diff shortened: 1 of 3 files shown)


manifests/server/instance_passwd.pp
```diff
@@ -0,0 +1,63 @@
...
+# @param postgres_password Sets the password for the postgres user to your specified value. By default, this setting uses the superuser account in the Postgres database, with a user called postgres and no password.
+# lint:endignore:140chars
+define postgresql::server::instance_passwd (
+  $user                                           = $postgresql::server::user,
+  $group                                          = $postgresql::server::group,
+  $psql_path                                      = $postgresql::server::psql_path,
+  $port                                           = $postgresql::server::port,
+  $database                                       = $postgresql::server::default_database,
+  $module_workdir                                 = $postgresql::server::module_workdir,
+  $postgres_password = $postgresql::server::postgres_password,
```
bastelfreak, line 18: the indend looks a bit odd here?

chelnak (approved):

PR #1396, opened 2023-02-02
Cleanup Systemd drop-in file handling

SimonHoenscheid:
* add puppet/systemd module
* remove systemd daemon reload and raise minimal puppet version to 6.1
* remove old "drop-in file" removal, was in place 3 years now
* move systemd drop-in file handling to seperate define
* Implement recent feedback
* define is now private
* rename parameter in define call
* fix unit tests
* fix rubocop complains
* fix path, set default fact
* fix systemd drop in file, adding template parameters to systemd define
* added reason for drop in file in a comment
* added reviewers feedback
* remove ensure parameter for systemd drop-in file handling, except
  for the define itself.
 * systemd_extra_args expects String, default is undef now

### DIFF
--- a/manifests/server/instance/config.pp
+++ b/manifests/server/instance/config.pp
@@ -245,58 +245,14 @@ define postgresql::server::instance::config (
     }
   }
   # lint:ignore:140chars
-  # RHEL 7 and 8 both support drop-in files for systemd units.  The old include directive is deprecated and may be removed in future systemd releases.
-  # Gentoo also supports drop-in files.
+  # RHEL 7 and 8 both support drop-in files for systemd units. Gentoo also supports drop-in files.
+  # Edit 02/2023 RHEL basedc Systems and Gentoo need Variables set for $PGPORT, $DATA_DIR or $PGDATA, thats what the drop-in file is for.
   # lint:endignore:140chars
   if $facts['os']['family'] in ['RedHat', 'Gentoo'] and $facts['service_provider'] == 'systemd' {
-    # While Puppet 6.1 and newer can do a daemon-reload if needed, systemd
-    # doesn't appear to report that correctly in all cases.
-    # One such case seems to be when an overriding unit file is removed from /etc
-    # and the original one from /lib *should* be used again.
-    #
-    # This can be removed when Puppet < 6.1 support is dropped *and* the file
-    # old-systemd-override is removed.
-    $systemd_command = ['systemctl', 'daemon-reload']
-    exec { 'restart-systemd':
-      command     => $systemd_command,
-      refreshonly => true,
-      path        => '/bin:/usr/bin:/usr/local/bin',
-      before      => Class['postgresql::server::service'],
-    }
-
-    file {
-      default:
-        ensure => file,
-        owner  => root,
-        group  => root,
-        notify => [Exec['restart-systemd'], Class['postgresql::server::service']],
-        before => Class['postgresql::server::reload'];
-
-      'systemd-conf-dir':
-        ensure => directory,
-        path   => "/etc/systemd/system/${service_name}.service.d";
-
-      # Template uses:
-      # - $facts['os']['name']
-      # - $facts['os']['release']['major']
-      # - $service_name
-      # - $port
-      # - $datadir
-      # - $extra_systemd_config
-      'systemd-override':
-        path    => "/etc/systemd/system/${service_name}.service.d/${service_name}.conf",
-        content => template('postgresql/systemd-override.erb'),
-        require => File['systemd-conf-dir'];
-    }
-
-    if $service_enable != 'mask' {
-      # Remove old unit file to avoid conflicts
-      file { 'old-systemd-override':
-        ensure => absent,
-        path   => "/etc/systemd/system/${service_name}.service",
-        notify => [Exec['restart-systemd'], Class['postgresql::server::service']],
-        before => Class['postgresql::server::reload'],
-      }
+    postgresql::server::instance::systemd { $service_name:
+      port                 => $port,
+      datadir              => $datadir,
+      extra_systemd_config => $extra_systemd_config,
     }
   }
 }
--- a/manifests/server/instance/systemd.pp
+++ b/manifests/server/instance/systemd.pp
@@ -1,0 +1,26 @@
+# @summary This define handles systemd drop-in files for the postgres main instance (default) or additional instances
+# @param service_name Overrides the default PostgreSQL service name.
+# @param drop_in_ensure sets the Systemd drop-in file to present or absent
+# @api private
+define postgresql::server::instance::systemd (
+  Variant[String[1], Stdlib::Port] $port,
+  Stdlib::Absolutepath $datadir,
+  Optional[String[1]] $extra_systemd_config = undef,
+  String[1] $service_name                   = $name,
+  Enum[present, absent] $drop_in_ensure     = 'present',
+
+) {
+  # Template uses:
+  # - $port
+  # - $datadir
+  # - $extra_systemd_config
+  systemd::dropin_file { "${service_name}.conf":
+    ensure  => $drop_in_ensure,
+    unit    => "${service_name}.service",
+    owner   => 'root',
+    group   => 'root',
+    content => template('postgresql/systemd-override.erb'),
+    notify  => Class['postgresql::server::service'],
+    before  => Class['postgresql::server::reload'],
+  }
+}
--- a/templates/systemd-override.erb
+++ b/templates/systemd-override.erb
@@ -1,6 +1,3 @@
-<%- if @os['name'] == 'Fedora' and @os['release']['major'] <= '31' -%>
-.include /usr/lib/systemd/system/<%= @service_name %>.service
-<% end -%>
 [Service]
 Environment=PGPORT=<%= @port %>
 <%- if @os['family'] == 'Gentoo' -%>
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -10,3 +10,4 @@ fixtures:
     puppet_agent: 'https://github.com/puppetlabs/puppetlabs-puppet_agent.git'
     stdlib: "https://github.com/puppetlabs/puppetlabs-stdlib.git"
     yumrepo_core: "https://github.com/puppetlabs/puppetlabs-yumrepo_core.git"
+    systemd: "https://github.com/voxpupuli/puppet-systemd.git"
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -26,7 +26,7 @@ class postgresql::params inherits postgresql::globals {
   $package_ensure               = 'present'
   $module_workdir               = pick($module_workdir,'/tmp')
   $password_encryption          = undef
-  $extra_systemd_config         = ''
+  $extra_systemd_config         = undef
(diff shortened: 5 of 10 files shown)


manifests/server/instance/config.pp
```diff
@@ -245,58 +245,11 @@
     }
   }
   # lint:ignore:140chars
-  # RHEL 7 and 8 both support drop-in files for systemd units.  The old include directive is deprecated and may be removed in future systemd releases.
+  # RHEL 7 and 8 both support drop-in files for systemd units.
   # Gentoo also supports drop-in files.
+  # Edit 02/2023 Lots of Linuxes use Systemd this time, do we want to use this on maybe all supported Systems?
```
SimonHoenscheid, line 250: @chelnak @bastelfreak came to my mind, I left a comment

manifests/server/instance/systemd.pp
```diff
@@ -0,0 +1,20 @@
+# @param service_name Overrides the default PostgreSQL service name.
```
bastelfreak, line 1: please add an `@summary` to describe what the define does.

chelnak (changes requested):
Small suggestion regarding puppet versions

manifests/server/instance/systemd.pp
ekohl: And consider making it `@api private`.

manifests/server/instance/systemd.pp
```diff
@@ -0,0 +1,20 @@
...
+define postgresql::server::instance::systemd (
+  String[1] $service_name = $name,
+) {
+  # Template uses:
+  # - $facts['os']['name']
+  # - $facts['os']['release']['major']
```
ekohl, line 7: I always said you can drop the Fedora specific bits there. It never made sense, but didn't want to block the PR from moving forward.

manifests/server/instance/systemd.pp
```diff
@@ -0,0 +1,20 @@
...
+  # - $service_name
+  # - $port
+  # - $datadir
+  # - $extra_systemd_config
+  systemd::dropin_file { "${service_name}.conf":
```
ekohl, line 12: Consider supporting ensuring the state (like `absent`).

manifests/server/instance/config.pp
ekohl: I think it depends on how the other systems use overrides. For example, on Debian it doesn't appear to support changing the data directory. That always uses `/var/lib/postgresql/%I` so I'd say there is no point in doing so. The same for changing ports.

In fact, I would check if changing the data directory and port is still up to date.

For example, changing the data directory via `postgresql.conf` is actually mostly invalid. In https://bugzilla.redhat.com/show_bug.cgi?id=1935301 I was told this breaks various tooling. In upstream (https://www.postgresql.org/message-id/3566642.1618422939%40sss.pgh.pa.us)  they recommended filtering it out of `postgresql.conf`.

The main reason is that the config (at least on RHEL) is present in the data directory, so you end up with a chicken-and-egg situation.

manifests/server/instance/systemd.pp
SimonHoenscheid: does @api private work with defines?

manifests/server/instance/config.pp
SimonHoenscheid: I checked this on a RHEL 7 with SCL Postgres 13, the port now works out of the box, the data directory ist still needed

manifests/server/instance/systemd.pp
SimonHoenscheid: @summary is there

manifests/server/instance/systemd.pp
SimonHoenscheid: I also removed the service name, because the block with the conditon using it, was removed

manifests/server/instance/systemd.pp
SimonHoenscheid: done

manifests/server/instance/systemd.pp
bastelfreak: `@api private` works also for defined resources.

manifests/server.pp
```diff
@@ -103,6 +104,7 @@
   $service_provider                                = $postgresql::params::service_provider,
   $service_reload                                  = $postgresql::params::service_reload,
   $service_status                                  = $postgresql::params::service_status,
+  Enum[present, absent] $systemd_drop_in_ensure    = $postgresql::params::systemd_drop_in_ensure,
```
bastelfreak, line 107: to make the puppet-strings documentation more beautiful I suggest to add the data here, and not in params.pp:
```suggestion
  Enum[present, absent] $systemd_drop_in_ensure    = 'present',
```

templates/systemd-override.erb
```diff
@@ -1,6 +1,3 @@
-<%- if @os['name'] == 'Fedora' and @os['release']['major'] <= '31' -%>
-.include /usr/lib/systemd/system/<%= @service_name %>.service
-<% end -%>
```
bastelfreak, line 3: in the future we should convert this to epp

manifests/server.pp
SimonHoenscheid: Yes, would break the existing style. In general, this is a topic on its own and should be done for the whole class/module in my opinion

bastelfreak: I'm not sure what's the Puppet rule here (if there is one); at Vox Pupuli we enforce new style guide (such as not using params.pp) before modules are converted.

templates/systemd-override.erb
SimonHoenscheid: we could also do this with systemd::manage_dropin. but I put the conversion to EPP on the list

manifests/server/instance/systemd.pp
SimonHoenscheid: done

manifests/server/instance/config.pp
SimonHoenscheid: Edited the comment

manifests/server.pp
SimonHoenscheid: quoted the present.

templates/systemd-override.erb
SimonHoenscheid: will refactor this in an other branch

SimonHoenscheid:
@bastelfreak @chelnak @ekohl this is ready for a final review, all previous comments have been addressed

manifests/server/instance/systemd.pp
```diff
@@ -0,0 +1,25 @@
+# @summary This define handles systemd drop-in files for the postgres main instance (default) or additional instances
+# @param service_name Overrides the default PostgreSQL service name.
+# @param systemd_drop_in_ensure sets the Systemd drop-in file to present or absent
+# @api private
+define postgresql::server::instance::systemd (
+  String[1] $service_name                       = $name,
+  Enum[present, absent] $systemd_drop_in_ensure = $postgresql::server::systemd_drop_in_ensure,
+  $port                                         = $postgresql::server::port,
+  $datadir                                      = $postgresql::server::datadir,
+  $extra_systemd_config                         = $postgresql::server::extra_systemd_config,
```
ekohl, line 10: Given this is a define, does it make sense to have any default at all (except perhaps for the ensure? You always pass them in for the specific instance anyway.

I'd also appreciate data types.

manifests/server/instance/systemd.pp
```diff
@@ -0,0 +1,25 @@
+# @summary This define handles systemd drop-in files for the postgres main instance (default) or additional instances
+# @param service_name Overrides the default PostgreSQL service name.
+# @param systemd_drop_in_ensure sets the Systemd drop-in file to present or absent
+# @api private
+define postgresql::server::instance::systemd (
+  String[1] $service_name                       = $name,
+  Enum[present, absent] $systemd_drop_in_ensure = $postgresql::server::systemd_drop_in_ensure,
```
ekohl, line 7: I'd keep the name simple (given the define is already named `systemd`):
```suggestion
  Enum[present, absent] $drop_in_ensure = $postgresql::server::systemd_drop_in_ensure,
```

manifests/server/instance/config.pp
```diff
@@ -57,6 +58,7 @@
   $logdir                                          = $postgresql::server::logdir,
   $service_name                                    = $postgresql::server::service_name,
   $service_enable                                  = $postgresql::server::service_enable,
+  Enum[present, absent] $systemd_drop_in_ensure    = $postgresql::server::systemd_drop_in_ensure,
```
ekohl, line 61: Does this need to be a separate ensure? I'd say we either have a global instance ensure (absent/present) or no parameter at all.

manifests/server/instance/systemd.pp
SimonHoenscheid: I just implemented your suggestions. 
Regarding the data types, I tried to be consistent with #1397

manifests/server/instance/config.pp
SimonHoenscheid: Currently the define does not have an ensure parameter, it is currently just the defined type of the former class. If I add an ensure for the define in the future, I will keep this in mind. Is that a solution which is OK for you @ekohl?

ekohl: Given there is no ensure parameter at all, I'd be tempted to not add this parameter either. Just keep it in the defined type.

I don't see an `@api private` (though perhaps it deserves one?) so I'm always hesitant to add more parameters than needed.

SimonHoenscheid: @ekohl I totally understand this and thanks for bringing this up. I would like to add an ensure but also want to make sure every resource within the define reflects the status correctly (otherwise its useless). 
And yes, the `@api private` is missing in all the new defines, except the systemd. 
I would like to do both topics in a later PR.

I will remove the `$drop_in_ensure`, except for the systemd defined type.

manifests/server/instance/systemd.pp
```diff
@@ -0,0 +1,26 @@
+# @summary This define handles systemd drop-in files for the postgres main instance (default) or additional instances
+# @param service_name Overrides the default PostgreSQL service name.
+# @param drop_in_ensure sets the Systemd drop-in file to present or absent
+# @api private
+define postgresql::server::instance::systemd (
+  Variant[String[1], Integer] $port,
```
ekohl, line 6: Do we still accept strings as ports? I'd at least narrow the integer down:
```suggestion
  Variant[String[1], Stdlib::Port] $port,
```

manifests/server/instance/systemd.pp
```diff
@@ -0,0 +1,26 @@
...
+  Stdlib::Absolutepath $datadir,
+  String[1] $extra_systemd_config,
```
ekohl, line 8: Isn't this optional? Or at least should be:
```suggestion
  Optional[String[1]] $extra_systemd_config = undef,
```

manifests/server/instance/systemd.pp
SimonHoenscheid: committed. I will add this to my review in #1397

manifests/server/instance/systemd.pp
SimonHoenscheid: committed. I will add this to my review in #1397

PR #1397, opened 2023-02-08
(CONT-361) Syntax update

LukasAud:
Code is now compliant with rules regarding:

parameter datatypes
parameter documentation


manifests/server.pp
```diff
@@ -89,79 +91,80 @@
...
-
-  Hash[String, Hash] $roles              = {},
-  Hash[String, Any] $config_entries      = {},
-  Postgresql::Pg_hba_rules $pg_hba_rules = {},
-
-  Boolean $backup_enable = $postgresql::params::backup_enable,
-  Hash $backup_options = {},
-  Enum['pg_dump'] $backup_provider = $postgresql::params::backup_provider,
+  String[1]                                    $package_name                 = $postgresql::params::server_package_name,
+  Enum['present', 'absent', 'latest']          $package_ensure               = $postgresql::params::package_ensure,
```
SimonHoenscheid, line 95: This will not work with versions, also affected: all other package resources

manifests/globals.pp
```diff
@@ -95,72 +96,72 @@
...
-  $datadir                                         = undef,
-  $confdir                                         = undef,
-  $bindir                                          = undef,
-  $xlogdir                                         = undef,
-  $logdir                                          = undef,
-  $log_line_prefix                                 = undef,
-  $manage_datadir                                  = undef,
-  $manage_logdir                                   = undef,
-  $manage_xlogdir                                  = undef,
+  Optional[String[1]] $datadir                     = undef,
```
SimonHoenscheid, line 130: It should be possible to use Stdlib::Absolutepath here. Is also an option for all the other parameters which can contain a path

manifests/server.pp
```diff
@@ -89,79 +91,80 @@
...
+  Boolean                                      $manage_recovery_conf         = $postgresql::params::manage_recovery_conf,
+  Boolean                                      $manage_postgresql_conf_perms = $postgresql::params::manage_postgresql_conf_perms,
+  Boolean                                      $manage_selinux               = $postgresql::params::manage_selinux,
+  String[1]                                    $module_workdir               = $postgresql::params::module_workdir,
+
+  Boolean                                      $manage_datadir               = $postgresql::params::manage_datadir,
+  Boolean                                      $manage_logdir                = $postgresql::params::manage_logdir,
+  Boolean                                      $manage_xlogdir               = $postgresql::params::manage_xlogdir,
+  Optional[String]                             $password_encryption          = $postgresql::params::password_encryption,
+  String                                       $extra_systemd_config         = $postgresql::params::extra_systemd_config,
```
SimonHoenscheid, line 155: Can be a string of minimal length 1 String[1]

SimonHoenscheid:
Great work, this module has been missing data types for a long time! I added a few review comments.

SimonHoenscheid (changes requested):

manifests/server.pp
ekohl: Isn't this `Optional[String[1]]`?

SimonHoenscheid: @ekohl I just transfered your suggestion from my PR. Thanks for pointing this out

manifests/server.pp
```diff
@@ -89,79 +91,80 @@
...
+  Boolean                                      $service_manage               = $postgresql::params::service_manage,
+  String[1]                                    $service_name                 = $postgresql::params::service_name,
+  Boolean                                      $service_restart_on_change    = $postgresql::params::service_restart_on_change,
+  Optional[String[1]]                          $service_provider             = $postgresql::params::service_provider,
+  String[1]                                    $service_reload               = $postgresql::params::service_reload,
+  String[1]                                    $service_status               = $postgresql::params::service_status,
+  String[1]                                    $default_database             = $postgresql::params::default_database,
+  Hash                                         $default_connect_settings     = $postgresql::globals::default_connect_settings,
+  Optional[String[1]]                          $listen_addresses             = $postgresql::params::listen_addresses,
+  Variant[String[1], Integer]                  $port                         = $postgresql::params::port,
```
SimonHoenscheid, line 111: @ekohl mentioned that in #1396 `Variant[String[1], Stdlib::Port] $port` would probably be the better fit

ekohl: And if this is a breaking change, I'd even suggest to tighten it to just `Stdlib::Port`.

LukasAud: We are trying to avoid turning this series of syntax updates into breaking changes. But thanks for the suggestion, will implement as soon as I can.

manifests/server.pp
LukasAud: Will double check this value when I get the chance. I might have misjudged the parameter.

manifests/server.pp
ekohl: That's fair, but you can safely use `Stdlib::Port` instead of `Integer`. Using an integer outside of the valid port range would result in the application failing to start up anyway so I'd not consider it breaking.

manifests/server.pp
LukasAud: Right, double checked extra_systemd_config and it looks like its defined initially at params.pp with a value of ''.
So it cannot be setup as Optional because its technically not an undef value and it doesnt get assigned an undef value either. And it cannot be a String[1] because it is indeed an empty-string.

Im not quite sure this is a change that can be implemented without breaking something else.

manifests/server.pp
LukasAud: Yeah, brought this topic up in a short discussion and figured out that my definition of $ensure was too limiting. Specially in comparison to official documentation. Will be updating all affected $ensure values to reflect better the valid values.

LukasAud: Note: regarding the context PR in stdlib... while we agree that the full Enum is, essentially, an alias for String[1], we still want to keep $ensure datatypes as Enums. The main reason here is to keep documentation as accurate as possible in the long run.

It also makes the variable easier to identify and track when looking at the code. And, of course, it can also be helpful when debugging.

LukasAud: Done

manifests/globals.pp
LukasAud: Done

manifests/server.pp
LukasAud: Done

LukasAud:
New commit. Have adjusted the PR to implement some suggestions.

ekohl (changes requested):

manifests/client.pp
```diff
@@ -12,10 +12,10 @@
...
 #   Ensure the client package is installed
 class postgresql::client (
-  Enum['file', 'absent'] $file_ensure        = 'file',
-  Stdlib::Absolutepath $validcon_script_path = $postgresql::params::validcon_script_path,
-  String[1] $package_name                    = $postgresql::params::client_package_name,
-  String[1] $package_ensure                  = 'present'
+  Enum['file', 'absent'] $file_ensure          = 'file',
+  Stdlib::Absolutepath   $validcon_script_path = $postgresql::params::validcon_script_path,
+  String[1]              $package_name         = $postgresql::params::client_package_name,
+  Variant[Enum['present', 'absent', 'purged', 'disabled', 'installed', 'latest'], Pattern[/./]] $package_ensure = 'present'
```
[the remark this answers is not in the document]
LukasAud: It is indeed Pattern[/./]. The reason was commented on another review above. Long story short, its mainly for documentation purposes.

manifests/server.pp
```diff
@@ -89,79 +91,80 @@
...
+  String[1]                                    $service_status               = $postgresql::params::service_status,
+  String[1]                                    $default_database             = $postgresql::params::default_database,
+  Hash                                         $default_connect_settings     = $postgresql::globals::default_connect_settings,
+  Optional[String[1]]                          $listen_addresses             = $postgresql::params::listen_addresses,
+  Variant[String[1], Stdlib::Port]             $port                         = $postgresql::params::port,
+  String[1]                                    $ip_mask_deny_postgres_user   = $postgresql::params::ip_mask_deny_postgres_user,
+  String[1]                                    $ip_mask_allow_all_users      = $postgresql::params::ip_mask_allow_all_users,
+  Array[String[1]]                             $ipv4acls                     = $postgresql::params::ipv4acls,
+  Array[String[1]]                             $ipv6acls                     = $postgresql::params::ipv6acls,
+
```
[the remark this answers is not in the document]
LukasAud: Will investigate

manifests/backup/pg_dump.pp
```diff
@@ -41,30 +41,30 @@
...
-  Stdlib::Absolutepath $success_file_path = '/tmp/pgbackup_success',
-  String[1] $template = 'postgresql/pg_dump.sh.epp',
-  Array $time = ['23', '5'],
-  String[1] $weekday = '*',
-  Optional[Variant[String, Sensitive[String]]] $db_password = undef,
-  Optional[String[1]] $db_user = undef,
-  Optional[String[1]] $package_name = undef,
-  Optional[String[1]] $post_script = undef,
-  Optional[String[1]] $pre_script = undef,
+  String[1]                                    $dir,
```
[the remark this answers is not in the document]
LukasAud: Understandable. However, we are currently favouring a more 'readable' and structured style.

manifests/server.pp
```diff
@@ -89,79 +91,80 @@
...
+  Optional[String[1]]                          $plpython_package_name        = $postgresql::params::plpython_package_name,
+
+  Variant[Enum['running', 'stopped'], Boolean] $service_ensure               = $postgresql::params::service_ensure,
+  Boolean                                      $service_enable               = $postgresql::params::service_enable,
+  Boolean                                      $service_manage               = $postgresql::params::service_manage,
+  String[1]                                    $service_name                 = $postgresql::params::service_name,
+  Boolean                                      $service_restart_on_change    = $postgresql::params::service_restart_on_change,
+  Optional[String[1]]                          $service_provider             = $postgresql::params::service_provider,
+  String[1]                                    $service_reload               = $postgresql::params::service_reload,
+  String[1]                                    $service_status               = $postgresql::params::service_status,
```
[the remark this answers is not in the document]
LukasAud: Correct. Will amend this.

manifests/client.pp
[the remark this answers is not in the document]
ekohl: I get the documentation purposes, but I still think `String[1]` is easier to understand than `Pattern[/./]`, mostly because the . is misleading. `String[1]` parses easily in my mind to "non-empty string". Regexes are harder.

manifests/server.pp
SimonHoenscheid: I already set   $extra_systemd_config  = undef in my commit. please rebase.

manifests/server.pp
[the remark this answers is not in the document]
SimonHoenscheid: I would recomment to go with Stdlib::Ensure::Package then. Everybody who is interested can follow this. Also possible values can be mentionen in the parameter docs.

manifests/client.pp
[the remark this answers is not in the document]
SimonHoenscheid: I would recomment to go with Stdlib::Ensure::Package then. Everybody who is interested can follow this. Also possible values can be mentioned in the parameter docs.

manifests/server.pp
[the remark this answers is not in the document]
LukasAud: It doesn't look like its possible to implement any Stdlib::IP for either $ipv4acls nor $ipv6acls as these are arrays os strings containing rules. They dont really match any of our existing IP types.

It should, however, be possible to apply Stdlib::IP::Address::V4 to $ip_mask_deny_postrges_user and $ip_mask_allow_all_users. Will have to test it first, tho.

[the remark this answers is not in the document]
LukasAud: Upon further investigation, it looks like mask_deny and mask_allow are variables being passed into pg_hba_rule.pp where they are expected as simple Strings. 

Furthermore, [documentation](https://www.postgresql.org/docs/current/auth-pg-hba-conf.html) regarding this rule suggests that valid values for these rules range from full IPv4 and IPv6 rules to shortened versions of the IPv6 rules and even keywords such as localhost or addresses (.example.com).

I dont believe its possible to use IP types here. At the very least not in an efficient way.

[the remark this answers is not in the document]
ekohl: That sounds like `Stdlib::Host` or even `Variant[Enum['samehost', 'samenet'], Stdlib::Host]` if you want to be explicit about the special values, except that it doesn't cover the `.example.com` case.

manifests/client.pp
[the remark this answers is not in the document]
LukasAud: Brought this up as a discussion and I think we can go ahead with the:

Variant[Enum['present', 'absent', 'purged', 'disabled', 'installed', 'latest'], String[1]] $package_ensure = 'present',

suggestion. It fairly easy to read and documentation should remain comprehensive enough.

manifests/backup/pg_dump.pp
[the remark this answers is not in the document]
kenyon: It doesn't help code readability much, but it makes diffs harder to read and review. +1 for not doing alignment.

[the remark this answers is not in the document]
ekohl: I'd even argue it's less readable if you have a long type (like the version you wrote for the package ensure) there is a lot of whitespace, so it becomes hard to see which data types belong to which variables.

david22swan (approved):

manifests/backup/pg_dump.pp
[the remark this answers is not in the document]
david22swan: Personally I find it best for when I have to search out a specific variable in the list, especially for larger modules that have upwards of a dozen.
I've always found it easier to track down that one that I'm looking for.

PR #1399, opened 2023-02-13
convert ERB templates to EPP

SimonHoenscheid:
rebase against #1396 after merge


ekohl (commented):
I like this, but wonder if it would make sense to add data types to the EPP files themselves. My biggest issue with ERB templates is that it's very easy to forget some (required) value and it silently proceeds. Data types are awesome to prevent that.

ekohl:
> You are thinking of a block like these right?

Yes exactly.

templates/pg_dump.sh.epp
```diff
@@ -1,3 +1,20 @@
...
+  Boolean                                  $delete_before_dump,
+  String[1]                                $dir,
+  Enum['plain','custom','directory','tar'] $format,
+  Array                                    $optional_args,
+  Optional[String[1]]                      $post_script,
+  Optional[String[1]]                      $pre_script,
+  String[1]                                $dir_group,
+  String[1]                                $dir_mode,
+  String[1]                                $dir_owner,
+  Enum['present','absent']                 $ensure,
```
ekohl, line 14: This doesn't show up in the template itself.

templates/pg_dump.sh.epp
```diff
@@ -1,3 +1,20 @@
...
+  Integer                                  $rotate,
```
ekohl, line 15: Given the usage in `find` I think this should be `Integer[0]`.

templates/pg_dump.sh.epp
```diff
@@ -1,3 +1,20 @@
...
+  Optional[String[1]]                      $db_user,
+  Boolean                                  $delete_before_dump,
+  String[1]                                $dir,
+  Enum['plain','custom','directory','tar'] $format,
+  Array                                    $optional_args,
+  Optional[String[1]]                      $post_script,
+  Optional[String[1]]                      $pre_script,
+  String[1]                                $dir_group,
+  String[1]                                $dir_mode,
+  String[1]                                $dir_owner,
```
ekohl, line 13: These dir_* variables are unused in the template

templates/pg_dump.sh.epp
```diff
@@ -1,3 +1,20 @@
+<%- |
+  Boolean                                  $compress,
+  Array                                    $databases,
+  Optional[String[1]]                      $db_user,
+  Boolean                                  $delete_before_dump,
+  String[1]                                $dir,
+  Enum['plain','custom','directory','tar'] $format,
+  Array                                    $optional_args,
```
ekohl, line 8: Is this `Array[String[1]]` or do we also expect `Integer` in there?

templates/pg_dump.sh.epp
```diff
@@ -1,3 +1,20 @@
...
+  Optional[String[1]]                      $post_script,
+  Optional[String[1]]                      $pre_script,
```
ekohl, line 10: It calls `flatten($pre_script)` and `flatten($post_script)` which implies an `Array` is supported/expected. In addition to that, it may even be nested (given flatten is called). If you're doing a breaking change, I'd suggest `Array[String[1]]` and remove the `flatten()` call.

templates/pg_dump.sh.epp
```diff
@@ -1,3 +1,20 @@
+<%- |
+  Boolean                                  $compress,
+  Array                                    $databases,
```
ekohl, line 3: You could use `Array[String[1]] $databases = []` and simplify the code. Right now it still deals with `undef` and it not being an `Array`.

templates/pg_hba_rule.conf.epp
```diff
@@ -0,0 +1,15 @@
+<%- |     
+  String                                    $name,
+  String                                    $description,
+  Variant[String, Integer]                  $order,
+  Postgresql::Pg_hba_rule_type              $type,
+  String                                    $database,
+  String                                    $user,
+  Optional[Postgresql::Pg_hba_rule_address] $address,
+  String                                    $auth_method,
+  Optional[String]                          $auth_option,
```
ekohl, line 10: Should the strings here use `String[1]` instead?

templates/recovery.conf.epp
```diff
@@ -0,0 +1,64 @@
...
+  Optional[String]    $restore_command,
+  Optional[String[1]] $archive_cleanup_command,
+  Optional[String[1]] $recovery_end_command,
+  Optional[String[1]] $recovery_target_name,
+  Optional[String[1]] $recovery_target_time,
+  Optional[String[1]] $recovery_target_xid,
+  Optional[Boolean]   $recovery_target_inclusive,
+  Optional[String[1]] $recovery_target,
+  Optional[String[1]] $recovery_target_timeline,
+  Optional[Boolean]   $pause_at_recovery_target,
```
ekohl, line 11: I always find `Optional[Boolean]` interesting because it's usually equal to `Boolean $var = false`. Not always, since you can have a tri-state where undefined is special cased, but not here. So I'd suggest to simplify it here.

templates/pg_dump.sh.epp
SimonHoenscheid: I took over the current types in #1397 or these which are currently in the module. If there are adjustments over all, I would do these later.

templates/pg_dump.sh.epp
SimonHoenscheid: see comment above

bastelfreak (approved):
I think this is a good way forward an can be merged. Yes the types could be a bit stricter in some places, but the PR is already huge. I think this is good to go, then #1397 is a good enhancement. We can always improve the datatypes further.

chelnak (approved):

PR #1400, opened 2023-02-14
deprecate old code managing sysconfig files

SimonHoenscheid:
* used on RHEL6
* old Debian and Ubuntu Versions
* Systemd takes care of that now

### DIFF
--- a/manifests/server/config_entry.pp
+++ b/manifests/server/config_entry.pp
@@ -2,7 +2,7 @@
 #
 # @param ensure Removes an entry if set to 'absent'.
 # @param value Defines the value for the setting.
-# @param path Path for postgresql.conf 
+# @param path Path for postgresql.conf
 #
 define postgresql::server::config_entry (
   Enum['present', 'absent'] $ensure = 'present',
@@ -70,10 +70,6 @@ define postgresql::server::config_entry (
     'max_pred_locks_per_transaction'      => undef,
   }
 
-  Exec {
-    logoutput => 'on_failure',
-  }
-
   if ! ($name in $requires_restart_until and (
       ! $requires_restart_until[$name] or
       versioncmp($postgresql::server::_version, $requires_restart_until[$name]) < 0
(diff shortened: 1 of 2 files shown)


manifests/server/config_entry.pp
```diff
@@ -91,73 +87,6 @@
     }
   }
 
-  # We have to handle ports and the data directory in a weird and
-  # special way.  On early Debian and Ubuntu and RHEL we have to ensure
-  # we stop the service completely. On RHEL 7 we either have to create
-  # a systemd override for the port or update the sysconfig file, but this
-  # is managed for us in postgresql::server::config.
-  if $facts['os']['name'] == 'Debian' or $facts['os']['name'] == 'Ubuntu' {
```
ekohl, line 99: This is no longer relevant?

ekohl (dismissed):
Looks like we dropped RHEL 6 support in https://github.com/puppetlabs/puppetlabs-postgresql/commit/32b770f903d584a9581f26b0c818f6508f29e2d1 (and other EL6 in other commits). If you can confirm the Debian part is no longer relevant then :+1:

manifests/server/config_entry.pp
SimonHoenscheid: I am pretty sure this is no longer needed. And the systemd drop-in is not managed on debian based systems. I would rely on the tests here, but I can also test a bit on a debian based system

ekohl: I really do wonder how often the data directory is changed. As I mentioned elsewhere, I'm not sure it's supported to have the data directory in postgresql.conf at all. So I think it may be ok to drop. Just a bit unsure about it.

SimonHoenscheid: I did this on many debian and ubuntu systems, changing it in postgresql.conf works

ekohl: Probably because they store it in `/etc` rather than `/var` like Red Hat does (which is crazy).

SimonHoenscheid: I just verified this on debian 10 with postgresql 11 (debians own package, default), changing port and data directory via postgresql.conf works fine.

bastelfreak (approved):

PR #1401, opened 2023-02-15
Clean up acceptance tests to minimize the spec helper

ekohl:
During cfgmgmtcamp it was mentioned that Puppet may want to switch back from Litmus to Beaker. This takes some small steps by reducing the size of `spec_helper_acceptance_local.rb`, which in my experience is always a good step towards alignment. Even if the switch doesn't happen, reducing local spec helpers is always good.

I'm going to rely on CI to make sure this is valid.

### DIFF
--- a/spec/acceptance/postgresql_psql_spec.rb
+++ b/spec/acceptance/postgresql_psql_spec.rb
@@ -73,7 +73,7 @@ describe 'postgresql_psql' do
       idempotent_apply(pp_five)
     end
 
-    pp_six = <<-MANIFEST.unindent
+    pp_six = <<~MANIFEST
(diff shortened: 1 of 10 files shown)


bastelfreak:
@ekohl can you rebase this one?

bastelfreak:
mhm can you apply some rubocop autofix magic?

jst-cyr:
@ekohl and @bastelfreak : It looks like the genesis of this cleanup was a possible move back to Beaker on the puppet side. From what I know, that didn't happen. Do we still need these changes?

ekohl:
No, this still uses Litmus. I'm sure there are still cleanups possible, but I don't intend to contribute those anymore.

PR #1404, opened 2023-02-24
Fix wrong Sensitive handling for updating role passwords

### LINKED ISSUE #1402: postgresql::server::role with sensitive passwords and enabled update_password are not working
## Describe the Bug
From the Datatype definition password_hash is allowed to be Sensitive but when setting it as one the password is set wrongly and ALTER password runs always.

Problem is https://github.com/puppetlabs/puppetlabs-postgresql/blob/v8.2.1/manifests/server/role.pp#L165 because is merged with string here https://github.com/puppetlabs/puppetlabs-postgresql/blob/main/manifests/server/role.pp#L182-L183

Resulting that `$pwd_hash_sql` will always replaced with `Sensitive [value redacted]` because it has not been unwrapped before.

## Expected Behavior
`postgresql::server::role` should be allowed to be used with a Sensitive password_hash and with update_password enabled.

## Steps to Reproduce
* Manifest with
```
postgresql::server::role { 'marmot':
  password_hash => postgresql::postgresql_password('marmot', 'mypasswd',true),
}
```
* run puppet
* run puppet again
  * shows that the password has been changed again
* try to login
```
psql --host=localhost --username=marmot --password
```
  * does not work because the password is not accepted

## Environment
 - puppetlabs-postgresql v8.2.1
 - postgresql 12.14
 - CentOS 8 Stream

## Additional Context
POC of the problem:
```
$password = Sensitive('Password')
$command = "/usr/bin/echo \"The password is: '${password}'\" > /tmp/data.txt"

exec { 'test':
  command => $command,
}
```

cruelsmith:
Before merging a String with a Sensitive the Sensitive need to be unwrapped else the value of the Sensitive will just `Sensitive [value redacted]`.

Since `$pwd_hash_sql` will be merged with `$pw_command` and `$unless_pw_command`  that also later will be Sensitive we just now ensure that `$pwd_hash_sql` will always be not a Sensitive.

Fixes #1402


manifests/server/role.pp
```diff
@@ -166,7 +166,7 @@
         $pwd_hash_sql = postgresql::postgresql_password(
           $username,
           $password_hash,
```
smortex, line 168: Calling `postgresql::postgresql_password()` with a variable `$password_hash` as the `password` parameter look wrong: as far as I can see, `postgresql::server::role` only has a `password_hash` parameter, so I would not expect it to hash anything itself (and be passed an already hashed password).

The `postgresql::server::db` class has a `password` parameter which is copied to the `password_hash` parameter of `postgresql::server::role` verbatim.  This is inconsistent, but I think this is for backward compatibility since when I check my (old) code I have things like:

```puppet
postgresql::server::db { $taiga::back::db_name:
  user     => $taiga::back::db_user,
  password => postgresql_password($taiga::back::db_user, $taiga::back::db_password),
}
```

The test suite mostly checks that values are sensitive ones, so basically that they match `Sensitive [value redacted]`, but not the actual unwrapped value.  [I opened a PR](https://github.com/puppetlabs/rspec-puppet/pull/42) to unwrap Sensitive values when running the test suite and as you guessed on Slack, some content seems to be wrapped twice.

cruelsmith: Yes, the hashing of a hash was also my first guess but it works since it returns the same value as long as the hashtype and salt does match.
Already have an idea to do a check like that to prevent duplicate hashing:
```
unless $password_hash_unsensitive =~ /^md5[0-9a-f]{32}$/ and $password_hash_unsensitive =~ /^scram-sha-256:/ { 
  $password_sql = postgresql_password([...])
} else {
  $password_sql = $password_hash_unsensitive
}
```
But i would say i create an additional PR since this PR is bug fixing and the duplicate hashing is more or less a maintenance change.

alexjfisher: Or just edit lines 182 and 183??

eg.
```puppet
$pw_command = "ALTER ROLE \"${username}\" ENCRYPTED PASSWORD '${pwd_hash_sql}'"
```

becomes
```puppet
$pw_command = "ALTER ROLE \"${username}\" ENCRYPTED PASSWORD '${pwd_hash_sql.unwrap}'"
```

cruelsmith: Sure but then we should also update the `Deferred` way to use Sensitive and rework to remove this, or?
https://github.com/puppetlabs/puppetlabs-postgresql/blob/ad330fc69fbaa54104934d7e0e05cf03655dd2ec/manifests/server/role.pp#L45-L49

ekohl: That part predates the Puppet 6.24 minimum version bound and can indeed be simplified.

smortex:
Hey! #1405 was merged in the `main` branch.  Rebasing your branch on top of it should fix the conflict.

From your working directory:

```
git fetch origin       # Download the latest code we have here
git rebase origin/main # Move your commits on top of the main branch
git push -f            # Send the changes (-f is required because we re-wrote history)
```

smortex (approved):
LGTM!

SimonHoenscheid (approved):

PR #1405, opened 2023-02-24
Unwrap sensitive values for testing

smortex:
When testing, sensitive values can be compared to the String `Sensitive [value redacted]`, but if we want to check that the redacted content is the one we expect, we should compare with another Sensitive wrapper.

This allows us to spot errors in the test suite (see #1404) more easily.

No functional change in this PR.


spec/defines/server/role_spec.rb
```diff
@@ -47,16 +47,18 @@
...
+        .with('command'     => sensitive(%(CREATE ROLE "test" ENCRYPTED PASSWORD 'new-pa$s' LOGIN NOCREATEROLE NOCREATEDB NOSUPERUSER  CONNECTION LIMIT -1)),
               'sensitive'   => 'true',
               'unless'      => "SELECT 1 FROM pg_roles WHERE rolname = 'test'",
               'port'        => '5432')
     end
     it 'has alter role for "test" user with password as ****' do
       is_expected.to contain_postgresql_psql('ALTER ROLE test ENCRYPTED PASSWORD ****')
-        .with('command'     => 'Sensitive [value redacted]',
+        .with('command'     => sensitive(%(ALTER ROLE "test" ENCRYPTED PASSWORD 'Sensitive [value redacted]')),
+              # FIXME: This is obviously wrong                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^
```
ekohl, line 58: Any idea what's needed to fix this? Or is it actually a bug?

alexjfisher: yeah, the sensitive values definitely need unwrapping before interpolating into another string.

```puppet
$secret = Sensitive('SECRET')

exec { 'test':
  command => Sensitive("/bin/echo My secret is ${secret} > /tmp/test_secret"),
}
```

```shell
cat /tmp/test_secret
My secret is Sensitive [value redacted]
```

alexjfisher: Given the minimum supported version in the metadata is 6.24.0 it should be fine to just throw some `unwrap`s in there without even checking if the value is `Sensitive` first.

See https://github.com/puppetlabs/puppet/commit/41991b2dc65fd00f484523b915958bd82f0ff836

smortex: Yeah, it is a bug.  I opened this PR in order to make it easier to diagnose the issues and fix it.  For that reason I do not change the tested value in this PR so that it can hopefully be merged quickly because it does not change the current (bad) behavior.  Yet, adding a comment for the obvious does not hurt.

The bug itself is tracked in #1404.

smortex:
I see this:
image

Is this an issue?  In another PR I though I could fix it by rebasing and all reviews went away.  I think this is ready for merging and this will allow to move on with #1404.

While here, any info on the failing jobs in CI?  I see similar EL-8 failures in other PRs…

bastelfreak (approved):

deric (approved):

SimonHoenscheid (approved):

SimonHoenscheid:
@smortex is this ready to go after a rebase?

smortex:
@SimonHoenscheid this is ready to go!  I rebased the PR, I will merge it when CI has passed.

PR #1406, opened 2023-02-27
Defaulting password encryption for version above 14

### LINKED ISSUE #1402: postgresql::server::role with sensitive passwords and enabled update_password are not working

cruelsmith:
* Fix missing handling of allowed undef value of parameter hash of postgresql::postgresql_password
* Fix edgecase where passwords starting with md5 but are not followed by 32 base64 chars will not be hashed by postgresql::postgresql_password
* Fix case of postgresql::postgresql_password where sensitive hashes are wrongly handled
  * Extend spec tests for 'postgresql_password function' to cover this
* Add respecting password_encryption for all internal postgresql::postgresql_password calls
* Add respecting password_encryption for postgresql::backup::pg_dump
* Add spec tests for new hash type handling of postgresql::server::role
* Update REFERENCE.md
  * Fixing lint 140chars inside REFERENCE.md

See https://www.postgresql.org/docs/14/runtime-config-connection.html#GUC-PASSWORD-ENCRYPTION

**Note:** That this change is will upgrade the password hashes to use `scram-sha-256` if a postgres version 14 or higher is used, `update_password` is set to true and cleartext passwords are used.
It will not change pre-hashed passwords.
If you want to force a default password_encryption you can set `postgresql::server::password_encryption` to the wanted one.

**Note:** The function `postgresql::postgresql_password` still has a hard default of `md5` as hash type. Use `postgresql::server::password_encryption` as fourth parameter when using it on your own to ensure you use the version based default one.

----
Fixes #1402
Includes and Closes #1404 #1405


lib/puppet/functions/postgresql/postgresql_password.rb
```diff
@@ -22,18 +22,21 @@
...
+      return Puppet::Pops::Types::PSensitiveType::Sensitive.new(password) if sensitive
+      return password
+    end
+    pass = case hash
+           when 'md5', nil # ensure default value when definded with nil
              "md5#{Digest::MD5.hexdigest(password.to_s + username.to_s)}"
-           else
+           when 'scram-sha-256'
              pg_sha256(password, (salt || username))
            end
```
smortex, line 41: Adding an `else` that raise an exception for the `case` statement will help adding new hashing methods if we need to in the future.

cruelsmith: Ok, added it but in normal usage it is a kind of useless code since the dispatch already ensures that only known hash types are handed over.

SimonHoenscheid:
@cruelsmith can you please rebase?

bastelfreak (approved):

Neustradamus:
@cruelsmith: Good job!

Linked to:
- https://github.com/scram-sasl/info/issues/1

PR #1410, opened 2023-03-27
Fix default value for $service_status on ArchLinux

### LINKED ISSUE #1408: postgresql::params does not have a valid default value for $service_status on ArchLinux
## Describe the Bug
On ArchLinux `$service_status` isn't defined and the module fails hard.
Error message:
`
Evaluation Error: Error while evaluating a Resource Statement, Postgresql::Server::Instance::Reload[main]:
parameter 'service_status' expects a String value, got Undef 
`

## Expected Behavior
`$service_status` should provide a valid value for ArchLinux out of the box.

## Steps to Reproduce
On ArchLinux:
`include postgresql::server`

## Additional Context
A quick fix could be to change the value for `$service_status` ([line 138 in params.pp](https://github.com/puppetlabs/puppetlabs-postgresql/blob/main/manifests/params.pp#L138)) to `"systemctl status ${service_name}"`

smortex:
The value can be overriden from `postgresql::globals`, but if no
override is provided by the user, the module should use a valid default.
ArchLinux use systemd, so use the relevant `systemctl` command.

Fixes #1408


ekohl (changes requested):
Isn't it better to change the type definition to allow `Undef`? Then the service provider can figure it out by itself.

Actually something we should do for most OSes anyway instead of writing out the systemd command.

smortex:
That make sense: this module has a load of legacy and I trend to contribute to it by copying the patterns I see in the other OS code and are missing rather than killing everything with fire and make everything simple.  I'll work on it on another PR.

ekohl:
@SimonHoenscheid is working on that, but this isn't going to work when it's defined in the class. Perhaps it needs to be modified to something interpolated, though a safe implementation would use an array. I wonder what's best to do here.

smortex:
While the PR is not perfect, it can unbreak the module for ArchLinux and that would unbreak CI for some of our Voxpupuli modules.  Maybe this fix can go as it is for now (and be released as a patch version), and updated again when we have some cluster management code which would be at least a minor version bump?

genebean (dismissed):
This seems inline with other code within the file. Refactoring the process seems like something for a different PR to me.

ekohl (dismissed):
Fair enough.

smortex:
Rebased on top of `main` so that the branch does not appear out-of-date for GitHub (Is that a repo requirement or maybe I got fooled by GitHub, this nuked all reviews :unamused:)?

Phil-Friderici:
@puppetlabs/modules any chance to review this very tiny PR that is only one line and would help me/us greatly ?

bastelfreak:
I rebased it against main to fix the SLES errors.

bastelfreak (approved):
as mentioned in the comments, that's not perfect but will fix Arch Linux.

Phil-Friderici:
@bastelfreak THANK YOU, Thank You, thank you

PR #1414, opened 2023-04-19
(CONT-792) - Add Puppet 8/Drop Puppet 6

jordanbreen28:
This PR adds support for Puppet 8 and Drops support for puppet 6.


LukasAud (approved):
LGTM

.rubocop.yml
```diff
@@ -1,10 +1,15 @@
...
 require:
 - rubocop-performance
 - rubocop-rspec
 AllCops:
+  NewCops: enable
   DisplayCopNames: true
-  TargetRubyVersion: '2.5'
+  ExtraDetails: true
+  DisplayStyleGuide: true
+  TargetRubyVersion: '2.7'
```
alexjfisher, line 12: Puppetserver **7** is only jruby 9.3 and therefore only compatible with ruby 2.6.  If you try to fix any of the rubocop_todo.yml contents you will likely break this module for _all_ users (since Puppet 8 isn't even out yet)

jordanbreen28: @alexjfisher thanks for being thorough and bringing this to our attention.. You have prompted an important conversation in the team.

PR #1418, opened 2023-04-21
Prevent nested sensitive wrap of password alter query

### LINKED ISSUE #1417: Passing in a Sensitive to postgresql::server::db as "password" causes broken user
## Describe the Bug
When passing in a `Sensitive[String]` to `postgresql::server::db`, the resulting user has a broken password.
Also each Puppet run resets the password to a different, but still broken, string.

This bug surfaced in https://github.com/puppetlabs/puppetlabs-postgresql/pull/1313

## Expected Behavior

The password should be set correctly and only once.

## Steps to Reproduce

1. Use the following code to trigger this problem:

```puppet
postgresql::server::db { $database_name:
  user     => "testuser",
  password => Sensitive("testpassword"),
  grant    => 'all',
}
```

## Environment
 - Version v8.2.1

## Additional Context
The reason for this problem is, that the above mentioned MR started using the `postgres_password` function to generate the password hash.  
There the password is returned as a `Sensitive` whenever the input was also a `Sensitive` (or more precisely: when `sensitive` was set to true). Then the `ALTER ROLE` SQL query gets constructed, **converted to a `Sensitive` (again!!)** and passed to `postgresql_psql` (see https://github.com/puppetlabs/puppetlabs-postgresql/blob/main/manifests/server/role.pp#L186-L187). 

Basically the logic constructs a Sensitive inside Sensitive which never gets unwrapped resulting in the following SQL query (yes exactly like that):
```
ALTER ROLE "<username>" ENCRYPTED PASSWORD 'Sensitive [value redacted]'
```

laugmanuel:
Fixes #1417

### DIFF
--- a/manifests/server/role.pp
+++ b/manifests/server/role.pp
@@ -160,7 +160,7 @@ define postgresql::server::role (
             $hash,
             $salt,
           ]
-        )
+        ).unwrap
       }
       else {
         $pwd_hash_sql = postgresql::postgresql_password(
@@ -169,7 +169,7 @@ define postgresql::server::role (
           false,
           $hash,
           $salt,
-        )
+        ).unwrap
       }
       if $pwd_hash_sql =~ Deferred {
         $pw_command = Deferred('sprintf', ["ALTER ROLE \"%s\" ENCRYPTED PASSWORD '%s'", $username, $pwd_hash_sql])


ekohl:
There are some related PRs. I think this is a better fix than https://github.com/puppetlabs/puppetlabs-postgresql/pull/1404 but https://github.com/puppetlabs/puppetlabs-postgresql/pull/1405 would improve the tests to catch this. Can you take a look and see if you can combine those tests into your PR?

laugmanuel:
> There are some related PRs. I think this is a better fix than #1404 but #1405 would improve the tests to catch this. Can you take a look and see if you can combine those tests into your PR?

Thanks for the heads-up. Somehow I did miss the existing Issue and PR - sorry for that.

I've cherry-picked the tests from #1404 to this PR.

spec/defines/server/role_spec.rb
```diff
@@ -50,18 +50,20 @@
...
-        .with('command' => 'Sensitive [value redacted]',
-              'sensitive' => 'true',
-              'unless' => 'Sensitive [value redacted]',
-              'port' => '5432')
+        .with('command'     => sensitive(%(ALTER ROLE "test" ENCRYPTED PASSWORD 'Sensitive [value redacted]')),
+              # FIXME: This is obviously wrong                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^
+              'sensitive'   => 'true',
+              'unless'      => sensitive(%(SELECT 1 FROM pg_shadow WHERE usename = 'test' AND passwd = 'Sensitive [value redacted]')),
+              # FIXME: This is obviously wrong                                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^
+              'port'        => '5432')
```
smortex, line 66: Isn't this PR about fixing these two FIXME :wink:?

I rebased #1405 on top of the main branch (it should fix the `Layout/HashAlignment` issues we see above), can you rebase your first commit on top of my branch and adjust the test suite which should now work as expected, and force push so that the PR is updated?

laugmanuel: You are absolutely right! The result of a rebase without fully reviewing every single line. I've rebased again and fixed the tests.

laugmanuel:
Broken tests for RHEL9 and SLES15 seem to be infrastructure related...

ekohl (approved):

cruelsmith:
Sry to be the party popper but make it even sense to call the `postgresql::postgresql_password` function with forcing it to be sensitive and instantly unwarping its result?

That is why in #1404 i just turned the sensitive return of the value of: https://github.com/puppetlabs/puppetlabs-postgresql/blob/e522f51fb074203d155d38c5155cf62b28c4ec4b/manifests/server/role.pp#L169
This also means that the unwrap for the `Deferred` is not needed in this change because the function already returns a not sensitive value.
https://github.com/puppetlabs/puppetlabs-postgresql/blob/83be16ed29ec681acce4b7839f45d3490cfa5acf/lib/puppet/functions/postgresql/postgresql_password.rb#L39-L43

laugmanuel:
> 

I guess both ways are valid and it's mostly personal preference. However, I do prefer unwrapping in Puppet, as passing in a Sensitive also returns a Sensitive in this case - so the function itself does not change the datatype.

ekohl:
I've merged https://github.com/puppetlabs/puppetlabs-postgresql/pull/1404 because it was a quick fix, but I still wouldn't mind merging this. If you're still interested in this then please rebase.

laugmanuel:
> I've merged #1404 because it was a quick fix, but I still wouldn't mind merging this. If you're still interested in this then please rebase.

Rebased this MR against current `main`.

smortex (commented):
I am confused. Can you provide a non-working unit test that this change fix?

laugmanuel:
> I am confused. Can you provide a non-working unit test that this change fix?
> 
> I feel like this got fixed by accident while fixing other misbehavior.

This PR fixes the same problem as #1404 does, but in a different way (as I wrote on Apr 28). Because of that, there are no broken tests at the moment.

I just followed the advice of @ekohl to rebase. However, I'm fine with closing this as well...

smortex:
I see, it looks like the bugfix was also in `main` and the commit in this PR now only show the differences between the two bugfix attempts.  Since the problem is now fixed, I will close this PR and the corresponding issue.

Thanks!

PR #1420, opened 2023-04-23
Fix wrong data type for `data_checksums` parameter

### LINKED ISSUE #1419: [9.0.0] data_checksums settings inconsistent and breaks existing code
After the update to 9.0.0 my code stop working with:
Class[Postgresql::Globals]: parameter 'data_checksums' expects a value of type Undef or String, got Boolean
When I take an look in the code of the module, now the option is sometimes an boolean (for example postgresql::server::instance::initdb) and sometimes an String (for example postgresql::globals)

smortex:
When adding data types, this `Boolean` value has been assigned a `String[1]` data type in some places, which break existing setup.

Consistently use an `Optional[Boolean]` data type.

Fixes #1419

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -563,7 +563,7 @@ Default value: `undef`
 
 ##### <a name="-postgresql--globals--data_checksums"></a>`data_checksums`
 
-Data type: `Optional[String[1]]`
+Data type: `Optional[Boolean]`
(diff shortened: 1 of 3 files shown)


jordanbreen28 (approved):
Thanks for getting this fix in so quick @smortex.

PR #1421, opened 2023-04-23
Add default version for Fedora 37, 38

lweller:
I added the default configuration for Fedora 37.


ekohl (approved):
This is correct according to https://packages.fedoraproject.org/pkgs/postgresql/postgresql/ but perhaps we can add F38 while we're at it.

lweller:
@ekohl I also added F38

Issue #1426, opened 2023-04-29
[9.0.1] breaks listen_addresses and increase the chance off an miss configuration


tuxmaster5000:
Before version 9 the parameter listen_addresses was an array, so existing code will not work any more:
`Class[Postgresql::Server]: parameter 'listen_addresses' expects a value of type Undef or String, got Tuple`
Change the array to an string will increase the chance of an miss configuration, because when set multiple address they must be an  comma-separated list.  So the old behaviour joining the array with an  comma as the seperator makes more sense.
The last working one was 8.2.1

jordanbreen28:
Thanks for raising this @tuxmaster5000! Will work to get this fixed and released ASAP.

Introduced in https://github.com/puppetlabs/puppetlabs-postgresql/pull/1397

tuxmaster5000:
@jordanbreen28  9.0.2 don't fix this :(
`Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Postgresql::Server::Config_entry[listen_addresses]: parameter 'value' expects a value of type Undef, String, or Numeric, got Tuple (file: /etc/puppetlabs/code/environments/production/modules/postgresql/manifests/server/instance/config.pp, line: 152)`

jordanbreen28:
@tuxmaster5000 ah! Let’s take a look.. can you check that your passing the values to listen_addresses in the form of an Array  of strings `['0.0.0.0', '127.0.0.1']` or string `'127.0.0.1'`

tuxmaster5000:
The  listen_addresses is an array in my case. Like it was working in the 8'er series.

ekohl:
https://github.com/puppetlabs/puppetlabs-postgresql/pull/1430#discussion_r1187332194

jordanbreen28:
@tuxmaster5000 okay this should be it now... 
v9.0.3 on the forge should fix this! 🤞 
Thanks for being thorough and spotting my oversight!

PR #1434, opened 2023-05-09
(GH-1426) - Update value to accept array

jordanbreen28:
Fixes https://github.com/puppetlabs/puppetlabs-postgresql/issues/1426

### DIFF
--- a/manifests/server/config_entry.pp
+++ b/manifests/server/config_entry.pp
@@ -5,9 +5,9 @@
 # @param path Path for postgresql.conf
 #
 define postgresql::server::config_entry (
-  Enum['present', 'absent']             $ensure = 'present',
-  Optional[Variant[String[1], Numeric]] $value  = undef,
-  Variant[Boolean, String[1]]           $path   = false
+  Enum['present', 'absent']                               $ensure = 'present',
+  Optional[Variant[String[1], Numeric, Array[String[1]]]] $value  = undef,
+  Variant[Boolean, String[1]]                             $path   = false
 ) {
   $postgresql_conf_path = $postgresql::server::postgresql_conf_path
 


GSPatton (approved):

manifests/server/config_entry.pp
```diff
@@ -5,9 +5,9 @@
 # @param path Path for postgresql.conf
 #
 define postgresql::server::config_entry (
-  Enum['present', 'absent']             $ensure = 'present',
-  Optional[Variant[String[1], Numeric]] $value  = undef,
-  Variant[Boolean, String[1]]           $path   = false
+  Enum['present', 'absent']                               $ensure = 'present',
+  Optional[Variant[String[1], Numeric, Array[String[1]]]] $value  = undef,
```
ekohl, line 9: Shouldn't this also an array of `Numeric`?

ekohl (commented):
Not objecting to having merged this, but I only just got the notification (something's up with GH). It would have been nice to add a (regression) test.

manifests/server/config_entry.pp
jordanbreen28: @ekohl I think that `Array[String[1]]` is suitable in the case of passing config values for listen_addresses (i understand value , curious on your thoughts of updating this to accept `Array[Numeric]` values as well.

I think continuing to update this is a bit out of scope, as initially this was implemented as a quick bug fix to ensure that existing manifests with values for `listen_addresses` did not break after the syntax update carried out on this module. but I'm open to your suggestions. 💯

ekohl: It's probably fine, since I don't know how many array values there are where it is numeric. You can indeed use the following in YAML as a workaround:
```yaml
---
value:
  - '1'
  - '2'
```

But very often you see some consistency, like `Variant[X, Y, Array[Variant[X, Y]]]` so that's something that always goes through my head. Still, it's not always true. For example, the current type can be written as `Variant[Undef, String[1], Numeric, Array[String[1]]]` (which is actually shorter than `Optional[]`) and you don't want to accept `Array[Undef]`.

That's at least how I look at these data types and assess if it fits what's needed.

The config entry is supposed to reflect a setting, so https://www.postgresql.org/docs/current/config-setting.html is a good resource. That states:

> All parameter names are case-insensitive. Every parameter takes a value of one of five types: boolean, string, integer, floating point, or enumerated (enum).

So String and Numeric certainly are good. Enum is covered by String. Then for booleans there's:

> Boolean: Values can be written as on, off, true, false, yes, no, 1, 0 (all case-insensitive) or any unambiguous prefix of one of these.

So accepting `Boolean` is also good.

There's no real array type, but there are types which are lists. https://www.postgresql.org/docs/current/runtime-config.html is quite long. Most are comma joined with `ssl_ciphers` as an exception. We can ignore that one and require the user to specify it as a single string.

In conclusion: I think the current type will work.

jordanbreen28: @ekohl as always, thanks for your great work and investigation into keeping our modules in top condition, it is appreciated!!

Completely understand your reasoning now thanks to all that info, I'll add a ticket to our backlog to further investigate this and implement it if required/we think it will be nice (which I must admit I do now..).

So thanks again 💯

PR #1449, opened 2023-06-21
puppetlabs/stdlib: Require 9.x

bastelfreak:
also contains #1448


SimonHoenscheid (approved):

deric (approved):

bastelfreak:
This is now marked as maintenance so it doesn't appear in the changelog. We do this because we reverted the change in #1480 . That allows us to do a minor release.

PR #1450, opened 2023-06-21
Add Server Instance Feature

SimonHoenscheid:
## Summary
This PR adds server instances to the module. 

## Additional Context
- Currently the server_instances define relies on the server class providing a baseline
- instances with different versions are not possible, except for upgrade scenarios (e.g 13 was managed before, 14 is the new managed version)


lib/puppet/provider/postgresql_conf/ruby.rb
```diff
@@ -0,0 +1,118 @@
...
+  def delete_header(lines)
+    header_regex = %r{^# HEADER:.*}
+    lines.delete_if do |entry|
+      entry.match?(header_regex)
+    end
+  end
+
+  def add_header(lines)
+    timestamp = Time.now.strftime('%F %T %z')
+    header = ["# HEADER: This file was autogenerated at #{timestamp}\n",
```
tuxmea, line 34: Will the timestamp be changed on every puppet run?

manifests/server/config_entry.pp
```diff
@@ -3,21 +3,16 @@
...
 define postgresql::server::config_entry (
-  Enum['present', 'absent']                               $ensure = 'present',
-  String[1]                                               $key    = $name,
-  Optional[Variant[String[1], Numeric, Array[String[1]]]] $value  = undef,
-  Variant[Boolean, String[1]]                             $path   = false
+  Enum['present', 'absent']                               $ensure  = 'present',
+  String[1]                                               $key     = $name,
+  Optional[Variant[String[1], Numeric, Array[String[1]]]] $value   = undef,
+  Optional[String[1]]                                     $comment = undef,
+  Variant[Boolean, String[1]]                             $path    = $postgresql::server::postgresql_conf_path
```
tuxmea, line 14: Should this be a Boolean or String or a Stdlib::AbsoluteUnixPath?

Ramesh7: I think this should be string now as earlier its was expecting either bool or string but now we have remove below code 
```
$target = $path ? {
    false   => $postgresql_conf_path,
    default => $path,
  }
```
so should be string?

lib/puppet/provider/postgresql_conf/ruby.rb
SimonHoenscheid: the write_config function is called on create and delete operations and deletes the existing header (if any), adds a new one with up to date timestamp and writes the changes to file. As far as I tested this right now, this works reliable.

manifests/server_instance.pp
```diff
@@ -0,0 +1,53 @@
+# lint:ignore:140chars
+# @param instance_name
```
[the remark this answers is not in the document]
SimonHoenscheid: sure, I will add detailed description and documentation as soon as the code is able to spin up and configure a working instance

lib/puppet/provider/postgresql_conf/ruby.rb
```diff
@@ -0,0 +1,118 @@
...
+Puppet::Type.type(:postgresql_conf).provide(:ruby) do
+  desc 'Set key/values in a postgresql config file.'
+  confine kernel: 'Linux'
+
+  # file parsing in a seperate function
+  def parse_config
+    file = File.open(resource[:target])
+    active_values_regex = %r{^\s*(?<key>[\w.]+)\s*=?\s*(?<value>.*?)(?:\s*#\s*(?<comment>.*))?\s*$}
+    active_settings = []
+
```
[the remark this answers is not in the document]
SimonHoenscheid: I added comments and a debug option.

manifests/server/instance/initdb.pp
```diff
@@ -22,28 +22,35 @@
...
-  Boolean                                  $manage_xlogdir = $postgresql::server::manage_xlogdir,
-  String[1]                                $module_workdir = $postgresql::server::module_workdir,
-  Boolean                                  $needs_initdb   = $postgresql::server::needs_initdb,
-  String[1]                                $user           = $postgresql::server::user,
-  Optional[String[1]]                      $username       = $postgresql::server::username,
-  Optional[String[1]]                      $xlogdir        = $postgresql::server::xlogdir,
+  Optional[String[1]]            $auth_host         = $postgresql::server::auth_host,
+  Optional[String[1]]            $auth_local        = $postgresql::server::auth_local,
+  Optional[Boolean]              $data_checksums    = $postgresql::server::data_checksums,
+  Stdlib::Absolutepath           $datadir           = $postgresql::server::datadir,
```
bastelfreak, line 37: we can pull all those path adjustments into a new PR

manifests/server/instance/late_initdb.pp
```diff
@@ -18,13 +18,15 @@
   Stdlib::Absolutepath                      $psql_path      = $postgresql::server::psql_path,
   Stdlib::Port                              $port           = $postgresql::server::port,
   String[1]                                 $module_workdir = $postgresql::server::module_workdir,
+
```
bastelfreak, line 21: this empty line can go away

manifests/server/instance/initdb.pp
```diff
@@ -22,28 +22,35 @@
 # @param module_workdir Working directory for the PostgreSQL module
 # @param needs_initdb Explicitly calls the initdb operation after server package is installed
 #   and before the PostgreSQL service is started.
+# @param port
```
bastelfreak, line 25: we can pull the port change into a new PR as well

manifests/server/instance/passwd.pp
```diff
@@ -42,7 +42,7 @@
     #  the default for pg_hba.conf.
     $escaped = postgresql::postgresql_escape($real_postgres_password)
     $exec_command = "${stdlib::shell_escape($psql_path)}${_dboption} -c \"ALTER ROLE \\\"${stdlib::shell_escape($user)}\\\" PASSWORD \${NEWPASSWD_ESCAPED}\"" # lint:ignore:140chars
-    exec { 'set_postgres_postgrespw':
+    exec { "set_postgres_postgrespw_${name}":
```
bastelfreak, line 45: bugfix for a new PR

manifests/server/reassign_owned_by.pp
```diff
@@ -8,19 +8,22 @@
...
 define postgresql::server::reassign_owned_by (
   String $old_role,
   String $new_role,
   String $db,
-  String $psql_user                      = $postgresql::server::user,
-  Stdlib::Port $port                     = $postgresql::server::port,
-  Hash $connect_settings                 = $postgresql::server::default_connect_settings,
+  String $psql_user               = $postgresql::server::user,
+  String[1] $group                = $postgresql::server::group,
+  Stdlib::Absolutepath $psql_path = $postgresql::server::psql_path,
```
bastelfreak, line 20: can also be moved into its own PR

manifests/server/schema.pp
```diff
@@ -8,22 +8,30 @@
 # @param owner Sets the default owner of the schema.
 # @param schema Sets the name of the schema.
 # @param connect_settings Specifies a hash of environment variables used when connecting to a remote server.
+# @param user Sets the OS user to run psql
+# @param group Sets the OS group to run psql
+# @param psql_path Sets path to psql command
+# @param module_workdir
+#   Specifies working directory under which the psql command should be executed.
+#   May need to specify if '/tmp' is on volume mounted with noexec option.
+# @param version Deprecated. Sets PostgreSQL version
```
bastelfreak, line 17: is that actually deprecated? can we remove? all those changes should be moved into its own PR as well

manifests/server/schema.pp
```diff
@@ -8,22 +8,30 @@
...
-  Optional[String[1]] $owner            = undef,
-  String[1]           $schema           = $title,
-  Hash                $connect_settings = $postgresql::server::default_connect_settings,
+  String[1]                                $db               = $postgresql::server::default_database,
+  Optional[String[1]]                      $owner            = undef,
+  String[1]                                $schema           = $title,
+  Hash                                     $connect_settings = $postgresql::server::default_connect_settings,
+  String[1]                                $user             = $postgresql::server::user,
+  String[1]                                $group            = $postgresql::server::group,
+  Variant[String[1], Stdlib::Absolutepath] $psql_path        = $postgresql::server::psql_path,
```
bastelfreak, line 30: should we only allow Stdlib::Absolutepath?

manifests/server/schema.pp
```diff
@@ -8,22 +8,30 @@
...
+  String[1]                                $module_workdir   = $postgresql::server::module_workdir,
```
bastelfreak, line 31: Stdlib::Absolutepath?

manifests/server_instance.pp
```diff
@@ -0,0 +1,135 @@
...
+  Hash $passwd_settings                   = {},
+  Hash $roles                             = {},
+  Hash $config_entries                    = {},
+  Hash $pg_hba_rules                      = {},
+  Hash $databases_and_users               = {},
+  Hash $databases                         = {},
+  Hash $database_grants                   = {},
+  Hash $table_grants                      = {},
+) {
+  if $instance_user != undef {
```
bastelfreak, line 38:
```suggestion
  if $instance_user {
```

spec/unit/type/postgresql_conf_spec.rb
```diff
@@ -1,4 +1,4 @@
-#! /usr/bin/env ruby # rubocop:disable Lint/ScriptPermission
+#! /usr/bin/env ruby
```
bastelfreak, line 1: we should remove the executeable bit and remove the shebang

README.md
```diff
@@ -71,7 +72,161 @@ psql -h my.postgres.server -U
 If you get an error message from these commands, your permission settings restrict access from the location you're trying to connect from. Depending on whether you want to allow connections from that location, you might need to adjust your permissions.
 
 For more details about server configuration parameters, consult the [PostgreSQL Runtime Configuration documentation](http://www.postgresql.org/docs/current/static/runtime-config.html).
+### Configure an instance
```
bastelfreak, line 75: to please the markdown gods, we need an empty line above the heading

README.md
```diff
@@ -71,7 +72,161 @@ psql -h my.postgres.server -U
...
+
+creating a new instance has the following advantages:
+* files are owned by the postgres user
+* instance is running under a different user, if the instance is hacked, the hacker has no access to the file system
+* the instance user can be an LDAP user, higher security because of central login monitoring, password policies, password rotation policies
+* main instance can be disabled
+
+
+Here is a profile which can be used to create instaces
+```
```
bastelfreak, line 87:
```suggestion
```puppet
```

README.md
```diff
@@ -71,7 +72,161 @@ psql -h my.postgres.server -U
...
+
+  $instances.each |String $instance, Hash $instance_settings| {
+    postgresql::server_instance { $instance:
+      *               => $instance_settings,
+    }
+  }
+}
+```
+And here is data to create an instance called test1:
+```
```
bastelfreak, line 110:
```suggestion
```yaml
```

manifests/server/db.pp
```diff
@@ -11,9 +11,10 @@
 # @param template Specifies the name of the template database from which to build this database. Defaults value: template0.
 # @param istemplate Specifies that the database is a template, if set to true.
 # @param owner Sets a user as the owner of the database.
-# @param port Specifies the port where the PostgreSQL server is listening on.
+# @param instance The name of the Postgresql database instance. Default value: 'main'.
+# @param port Specifies the port for the PostgreSQL server to listen on.
 # @param psql_user Overrides the default PostgreSQL super user and owner of PostgreSQL related files in the file system.
-# @param psql_group Overrides the default PostgreSQL user group to be used for related files in the file system.
+# @param psql_group Overrides the default postgres user group to be used for related files in the file system.
```
bastelfreak, line 17: was that change on purpose?

manifests/server/db.pp
```diff
@@ -11,7 +11,8 @@
 # @param template Specifies the name of the template database from which to build this database. Defaults value: template0.
 # @param istemplate Specifies that the database is a template, if set to true.
 # @param owner Sets a user as the owner of the database.
-# @param port Specifies the port where the PostgreSQL server is listening on.
+# @param instance The name of the Postgresql database instance. Default value: 'main'.
+# @param port Specifies the port for the PostgreSQL server to listen on.
```
bastelfreak, line 15: Is the change for the `port` param documentation intentional?

manifests/server/grant.pp
```diff
@@ -18,6 +18,7 @@
 # @param ensure Specifies whether to grant or revoke the privilege. Default is to grant the privilege. Valid values: 'present', 'absent'.
 # @param group Sets the OS group to run psql
 # @param psql_path Sets the path to psql command
+# @param instance The name of the Postgresql database instance. Default value: 'main'.
```
bastelfreak, line 21: we don't need to mention the default value, puppet-strings will pick it up automatically

manifests/server/instance/systemd.pp
```diff
@@ -5,26 +5,36 @@
 define postgresql::server::instance::systemd (
   Stdlib::Port $port,
   Stdlib::Absolutepath $datadir,
+  String[1] $service_name,
+  String[1] $instance_name                  = $name,
   Optional[String[1]] $extra_systemd_config = undef,
-  String[1] $service_name                   = $name,
```
bastelfreak, line 9: we should flip $service_name two lines up

manifests/server/tablespace.pp
```diff
@@ -5,13 +5,14 @@
...
 # @param connect_settings Specifies a hash of environment variables used when connecting to a remote server.
-# @param port the port of the postgresql instance that sould be used.
 # @param user Sets the OS user to run psql
 # @param group Sets the OS group to run psql
 # @param psql_path Sets path to psql command
 # @param module_workdir
 #   Specifies working directory under which the psql command should be executed.
 #   May need to specify if '/tmp' is on volume mounted with noexec option.
+# @param instance The name of the Postgresql database instance. Default value: 'main'.
+# @param port Port to use when connecting.
```
bastelfreak, line 15: can you drop this line ad readd the old parameter documentation

manifests/server_instance.pp
```diff
@@ -0,0 +1,136 @@
+# @summary define to install and manage additional postgresql instances
+# @param instance_name The name of the instance.
+# @param instance_user The user to run the instance as. Defaults to $instance_name if not specified.
+# @param instance_group The group to run the instance as. Defaults to $instance_name if not specified.
```
bastelfreak, line 4: we don't need to list the default values here

manifests/server_instance.pp
```diff
@@ -0,0 +1,136 @@
...
+  Hash $table_grants                      = {},
+) {
+  unless($facts['os']['family'] == 'RedHat' and $facts['os']['major'] == '8') {
+    warning('This define postgresql::server_instance is only tested on RHEL8')
+  }
+  if $instance_user {
+    $real_instance_user = $instance_user
+  } else {
+    $real_instance_user = $instance_name
+  }
```
bastelfreak, line 43:
```puppet
$instance_user = pick($instance_user, $instance_name)
```
or probably even better update this:

```puppet
Optional[String[1]] $instance_user      = undef,
```

to this:

```puppet
String[1] $instance_user      = $instance_name,
```

manifests/server_instance.pp
```diff
@@ -0,0 +1,136 @@
...
+
+  file { $instance_directories:
+    ensure => directory,
+  }
+
+  if $manage_instance_user_and_group {
+    User { $instance_user:
+      managehome => true,
+      home       => "/opt/pgsql/data/home/${instance_user}",
```
bastelfreak, line 57: I think we should make the home path configureable with a default of `"/opt/pgsql/data/home/${instance_user}"`

manifests/server_instance.pp
```diff
@@ -0,0 +1,136 @@
...
+  } else {
+    $real_instance_group = $instance_name
+  }
+
+  file { $instance_directories:
+    ensure => directory,
+  }
+
+  if $manage_instance_user_and_group {
+    User { $instance_user:
```
bastelfreak, line 55:
```suggestion
    user { $instance_user:
```

manifests/server_instance.pp
```diff
@@ -0,0 +1,136 @@
...
+      managehome => true,
+      home       => "/opt/pgsql/data/home/${instance_user}",
+      gid        => $instance_group,
+    }
+    Group { $instance_group: }
```
bastelfreak, line 60:
```suggestion
    group { $instance_group: }
```

and I suggest to set `system => true` (same for the user)

lib/puppet/provider/postgresql_conf/ruby.rb
```diff
@@ -0,0 +1,160 @@
...
+# to add, remove or modify settings.
+#
+# The provider is able to parse postgresql.conf files with the following format:
+# key = value # comment
+
+Puppet::Type.type(:postgresql_conf).provide(:ruby) do
+  desc 'Set keys, values and comments in a postgresql config file.'
+  confine kernel: 'Linux'
+
+  # file parsing in a seperate function
```
bastelfreak, line 14: I had a chat with @ekohl about it, I think we should use YARD to document the methods (see the line method for examples)

bastelfreak (commented):
I think this looks good at the moment. Can you maybe add an acceptance test to spin up multiple instances?

manifests/server/schema.pp
```diff
@@ -29,20 +30,22 @@
...
+    psql_user        => $user,
+    psql_group       => $group,
+    psql_path        => $psql_path,
+    port             => $port_override,
+    cwd              => $module_workdir,
     connect_settings => $connect_settings,
+    instance         => $instance,
   }
 
   postgresql_psql { "${db}: CREATE SCHEMA \"${schema}\"":
```
bzed, line 51: All the lines similar to 
`postgresql_psql { "${db}: CREATE SCHEMA \"${schema}\"":`
need to have the instance name included, otherwise you won't be able to create the same {schema, role, whatever} in different instances.
Typical error is

```Duplicate declaration: Postgresql_psql[CREATE ROLE foo ENCRYPTED PASSWORD ****] is already declared at....```

**Comment by bzed** [comment]
Also in my opinion the option that you have several postgresql version running in parallel is not really supported. You are introducing "instances", although the (imho) right thing to implement would be "clusters".
See for example the 'pg_createcluster' manpage on a Debian system for details. These things are also (slowly) coming to redhat and other more manually managed distributions.

File: manifests/server/schema.pp (hunk 29 above)
Comment (SimonHoenscheid) [reply manifests/server/schema.pp:51]: I am aware of this. We thought about addressing this with an issue for a feature, because having the same objects in different instances on the same node looked like an edge case for us. Would you be OK with this solution?

**Comment by SimonHoenscheid** [comment]
@bzed
yes, the PR description mentions running multiple versions is not supported, but has always been an update scenario. Managing the same instance in multiple versions would otherwise end up in a duplicate declaration.

Regarding the name, I am not sure which way to go here. I know the script wrappers you mentioned. These are the ones that prevent Debian and Ubuntu support right now. I have not seen them on other Linuxes so far.

**Comment by georg-h** [comment]
> Also in my opinion the option that you have several postgresql version running in parallel is not really supported. You are introducing "instances", although the (imho) right thing to implement would be "clusters". See for example the 'pg_createcluster' manpage on a Debian system for details. These things are also (slowly) coming to redhat and other more manually managed distributions.

@bzed in general you're right that the administrative ressource unit is called "cluster" in case of postgreSQL. But any other database flavors (Db2, Oracle, ...) call this thing "instance" so any DBA (like me) understands that "an instance contains and manages databases" and this relates also to a postgres cluster and reverse. 
On the other hand DBAs that are new to postgreSQL are often confused by the term "cluster" for this as usually a cluster are several single units that are interconnected and working as one big unit. It becomes worse if you put patroni on top of postgreSQL "clusters" as Patroni calls its multi-pg-instance HA construct also cluster (and means the above described multi single unit thing). 
When talking to non-DBAs I handle this in the way that I call it always "instance aka cluster" or "cluster aka instance" until they have understand that this is the same in case of postgreSQL.
kind regards
Georg

**Comment by bzed** [comment]
> > Also in my opinion the option that you have several postgresql version running in parallel is not really supported. You are introducing "instances", although the (imho) right thing to implement would be "clusters". See for example the 'pg_createcluster' manpage on a Debian system for details. These things are also (slowly) coming to redhat and other more manually managed distributions.
> 
> @bzed in general you're right that the administrative ressource unit is called "cluster" in case of postgreSQL. But any other database flavors (Db2, Oracle, ...) call this thing "instance" so any DBA (like me) understands that "an instance contains and manages databases" and this relates also to a postgres cluster and reverse. 

At the end its a name for something and I'm fine with everything as long as its obvious :)

**Comment by bzed** [comment]
> @bzed yes, the PR description mentions running multiple versions is not supported, but has always been an update scenario. Managing the same instance in multiple versions would otherwise end up in a duplicate declaration.

Thats not an update scenario, thats - at least here - daily business. We regularly run more than one version for a longer time on database servers until all applications are migrated. Thats a kind of update scenario, yes, but one that needs to be managed.

> Regarding the name, I am not sure which way to go here. I know the script wrappers you mentioned. These are the ones that prevent Debian and Ubuntu support right now. I have not seen them on other Linuxes so far.

All of the commercial packages (from EDB and others) already come with them, and also the redhat repositories of the community edition. Supporting that would actually be much easier as pg_lscluster has a --json option you can feed into facter.


You could just use 14/main as "instance", then you would have the version in the instance and could handle it appropriately.

**Comment by SimonHoenscheid** [comment]
> > @bzed yes, the PR description mentions running multiple versions is not supported, but has always been an update scenario. Managing the same instance in multiple versions would otherwise end up in a duplicate declaration.
> 
> Thats not an update scenario, thats - at least here - daily business. We regularly run more than one version for a longer time on database servers until all applications are migrated. Thats a kind of update scenario, yes, but one that needs to be managed.

The module did not support this before, even with one default/main instance. But I would love to see this as a feature.

> All of the commercial packages (from EDB and others) already come with them, and also the redhat repositories of the community edition. Supporting that would actually be much easier as pg_lscluster has a --json option you can feed into facter.
> 
> You could just use 14/main as "instance", then you would have the version in the instance and could handle it appropriately.

I dont use commercial Postgres packages besides RHEL. Sorry for not being aware of this. 
Anyhow this would also require changes to the initdb define, making use of pg_cluster if its available. It's a nice new feature, but I would also see this as a seperate extension.

File: lib/puppet/provider/postgresql_conf/ruby.rb (hunk 28 above)
Comment (SimonHoenscheid) [reply lib/puppet/provider/postgresql_conf/ruby.rb]: please check if thats ok

**Comment by SimonHoenscheid** [comment]
> I think this looks good at the moment. Can you maybe add an acceptance test to spin up multiple instances?

acceptance test added

**Review by Ramesh7 — APPROVED** [review]
The code looks good but need more review from postgres experts.

File: README.md [hunk 30]
```diff
@@ -72,6 +73,184 @@ If you get an error message from these commands, your permission settings restri
...
+  }
+}
+```
+
+And here is data to create an instance called test1:
+
+```yaml
+# stop default main instance
+postgresql::server::service_ensure: "stopped"
+postgresql::server::service_enable: false
```
smortex, line 127: This should go in the above profile rather than in Hiera.

README.md
```diff
@@ -72,6 +73,184 @@ If you get an error message from these commands, your permission settings restri
...
+        database: "replication"
+        user: "ins_test1"
+        auth_method: "peer"
+        order: 6
+      "local replication REPLICATION user":
+        type: "local"
+        database: "replication"
+        user: "rep_test1"
+        auth_method: "peer"
+        order: 7
```
smortex, line 252: These 120 lines of boilerplate are scary :scream:

I guess that if it was trivial it would have been done, but just to be sure, we have no way to skip some parameters that feel redundant (e.g. pass a datadir and infer the path to config files, generate the service name based on the instance name, etc)?  That would require to change a lot of parameter default values and use the value of previous parameters instead of taking a value directly from params.pp?

smortex (approved):
Looks globally good.  I have some concerns about the interface which feels very complex and repetitive but I don't think there is much we can do unfortunately.

README.md
```diff
@@ -72,6 +73,184 @@ If you get an error message from these commands, your permission settings restri
...
+      "/opt/pgsql/data/home":
+        ensure: directory
+      "/opt/pgsql/wal":
+        ensure: directory
+      "/opt/pgsql/log":
+        ensure: directory
+      "/opt/pgsql/log/13":
+        ensure: directory
+      "/opt/pgsql/log/13/test1":
+        ensure: directory
```
tuxmea, line 152: Can we place the directory names in an array? This would reduce the number of lines.

Another option would be to make use of the dirtree module and the dirtree type/provider inside the instance defined type.
Using dirtree option should be done in a different PR as this will raise different discussion.

README.md
```diff
@@ -72,6 +73,184 @@ If you get an error message from these commands, your permission settings restri
...
+    config_settings:
+      pg_hba_conf_path: "/opt/pgsql/data/13/test1/pg_hba.conf"
```
tuxmea, line 154: Here we have to replicate the version in many occasions.
Maybe version should become a hiera data, so it can be reused where needed.
Hardcoding version might make sense, if one is able to install multiple postgresql versions in parallel.

tuxmea (approved):

lib/puppet/provider/postgresql_conf/ruby.rb
```diff
@@ -0,0 +1,167 @@
...
+# This provider is used to manage postgresql.conf files
+# It uses ruby to parse the config file and
+# to add, remove or modify settings.
+#
+# The provider is able to parse postgresql.conf files with the following format:
+# key = value # comment
+
+Puppet::Type.type(:postgresql_conf).provide(:ruby) do
+  desc 'Set keys, values and comments in a postgresql config file.'
+  confine kernel: 'Linux'
```
smortex, line 12: (re-adding this comment because I was not reporting it where I though)

This seems to break the module on all non-linux OS:

```sh-session
root@agrajag /usr/home/romain # facter os
{
  architecture => "amd64",
  family => "FreeBSD",
  hardware => "amd64",
  name => "FreeBSD",
  release => {
    branch => "RC4",
    full => "14.0-RC4",
    major => "14",
    minor => "0"
  }
}
root@agrajag /usr/home/romain # puppet agent -t -E romain --tags Postgresql
[...]
Notice: /Stage[main]/Postgresql::Server::Service/Postgresql::Server::Instance::Service[main]/Anchor[postgresql::server::service::begin::main]: Dependency Postgresql_conf[port_for_instance_main] has failures: true
Notice: /Stage[main]/Postgresql::Server::Service/Postgresql::Server::Instance::Service[main]/Anchor[postgresql::server::service::begin::main]: Dependency Postgresql_conf[password_encryption_for_instance_main] has failures: true
Notice: /Stage[main]/Postgresql::Server::Service/Postgresql::Server::Instance::Service[main]/Anchor[postgresql::server::service::begin::main]: Dependency Postgresql_conf[data_directory_for_instance_main] has failures: true
[...]
Error: Could not find a suitable provider for postgresql_conf
```

PR #1452, opened 2023-06-29
Fix acceptance tests on SLES 15

smortex:
The net-tools package which contain netstat(1) has been deprecated and
is now part of the legacy modules.

Enable and install the required tooling to fix CI.


bastelfreak:
I think that's a duplicate of https://github.com/puppetlabs/puppetlabs-postgresql/pull/1453 ?

smortex:
> I think that's a duplicate of #1453 ?

w00t, I could not finish to test this yesterday but it looks like I was on the right way.  So it's fixed now, let's close this one and have passing CI!

PR #1460, opened 2023-07-25
Add Debian 12 compatibility

bastelfreak:
## Summary
Provide a detailed description of all the changes present in this pull request.

## Additional Context
Add any additional context about the problem here. 
- [ ] Root cause and the steps to reproduce. (If applicable)
- [ ] Thought process behind the implementation.

## Related Issues (if any)
Mention any related issues or pull requests.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified. (For example `puppet apply`)

PR #1461, opened 2023-07-25
maint: removing debian 12 from metadata

malikparvez:
Partially reverts puppetlabs/puppetlabs-postgresql#1460

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -57,8 +57,7 @@
       "operatingsystem": "Debian",
       "operatingsystemrelease": [
         "10",
-        "11",
-        "12"
+        "11"
       ]
     },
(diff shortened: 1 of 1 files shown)


ekohl (commented):
Can we do a partial revert and only remove it from metadata.json?

bastelfreak:
why does it need to be reverted? It works fine on Debian 12.

malikparvez:
@ekohl yes that might work, thanks for the suggestion

ekohl:
I think it's fair if it isn't tested in acceptance tests (which right now it isn't).

jordanbreen28 (approved):

PR #1469, opened 2023-08-25
uniform port parameter and $port_override value assignment

### LINKED ISSUE #1468: Identical $port_override behaviour
## Describe the Bug
Varous defines use  the $port_override variable, parameter and condition should behave thje same

See: Slack thread
https://puppetcommunity.slack.com/archives/C0W298S9G/p1692947310149489

Diskussed this with David Sandilands and he asked for an issue and PR.

SimonHoenscheid:
fixes #1468

### DIFF
--- a/manifests/server/database.pp
+++ b/manifests/server/database.pp
@@ -9,6 +9,7 @@
 # @param locale Overrides the locale during creation of the database.
 # @param istemplate Defines the database as a template if set to true.
 # @param connect_settings Specifies a hash of environment variables used when connecting to a remote server.
+# @param port Specifies the port for the PostgreSQL server port to connect to, default is 5432.
 define postgresql::server::database (
   Optional[String[1]] $comment          = undef,
   String[1]           $dbname           = $title,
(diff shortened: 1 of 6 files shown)


manifests/server/database.pp
```diff
@@ -19,6 +20,7 @@
   Optional[String[1]] $locale           = $postgresql::server::locale,
   Boolean             $istemplate       = false,
   Hash                $connect_settings = $postgresql::server::default_connect_settings,
+  Variant[String[1], Stdlib::Port, Integer] $port = $postgresql::params::port,
```
ekohl, line 23: This means you need to always specify it when you override it in the server. Perhaps it can default to `undef` and in the body use some `pick()` logic?

manifests/server/database.pp
```diff
@@ -34,11 +36,10 @@
...
   }
 
-  # If the connection settings do not contain a port, then use the local server port
   if $connect_settings != undef and 'PGPORT' in $connect_settings {
-    $port = undef
+    $port_override = $port
   } else {
-    $port = $postgresql::server::port
+    $port_override = $port
   }
```
ekohl, line 43: Both branches are the same, so it could be simplified to just `$port_override = $port`.

ekohl (changes requested):
I have a hard time processing this change, but the current implementation is a regression IMHO and needs some work. I'd also consider this backwards incompatible.

SimonHoenscheid:
@ekohl thanks for the feedback, I made some changes.

SimonHoenscheid:
drop in favor of #1487

PR #1479, opened 2023-08-28
flexible value for auth_method in pg_hba.conf if passwords are used

### LINKED ISSUE #1465: Default HBA does not take scram into account
Hello,

the default HBA rules are still hardcoded with [md5](https://github.com/puppetlabs/puppetlabs-postgresql/blob/2a4e7bb949def65585d2387151d25681e5240852/manifests/server/instance/config.pp#L116C11-L116C23), couldn't we use the $password_encryption variable of [config.pp](https://github.com/puppetlabs/puppetlabs-postgresql/blob/main/manifests/server/instance/config.pp) to initialize it ?

Best regards,
Olivier

SimonHoenscheid:
fixes #1465

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -25,7 +25,7 @@ class postgresql::params inherits postgresql::globals {
   $manage_selinux               = pick($manage_selinux, false)
   $package_ensure               = 'present'
   $module_workdir               = pick($module_workdir,'/tmp')
-  $password_encryption          = if versioncmp($version, '14') >= 0 { 'scram-sha-256' } else { 'md5' }
+  $password_encryption          = versioncmp($version, '14') ? { -1 => 'md5', default => 'scram-sha-256' }
   $extra_systemd_config         = undef
   $manage_datadir               = true
   $manage_logdir                = true
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -25,7 +25,7 @@ class postgresql::params inherits postgresql::globals {
   $manage_selinux               = pick($manage_selinux, false)
   $package_ensure               = 'present'
   $module_workdir               = pick($module_workdir,'/tmp')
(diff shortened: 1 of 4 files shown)


cruelsmith (approved):
Fine in my 👀
Maybe we also should add a spec test for that one? Currently we not really checking the hba rules in spec that comes out of the server config class.

manifests/params.pp
```diff
@@ -25,7 +25,7 @@
   $manage_selinux               = pick($manage_selinux, false)
   $package_ensure               = 'present'
   $module_workdir               = pick($module_workdir,'/tmp')
-  $password_encryption          = if versioncmp($version, '14') >= 0 { 'scram-sha-256' } else { undef }
+  $password_encryption          = if versioncmp($version, '14') >= 0 { 'scram-sha-256' } else { 'md5' }
```
bastelfreak, line 28: This should be a selector statement I think:
```suggestion
  $password_encryption          = versioncmp($version, '14') ? { -1 => 'md5', default => scram-sha-256' }
```

cruelsmith (approved):

SimonHoenscheid:
@cruelsmith I have problems getting this PR green, would you mind having a look?

cruelsmith:
Ah found it ... the issue is that the `postgresql::postgresql_password()` still defaults to `md5` and we manual need to set it to `scram-sha-256` in case of version >= 14.
Question is if it is possible to respect the `$postgresql::server::password_encryption` in the `postgresql::postgresql_password()` in some way. Else every code that uses that function needs to be updated to hand over that value.

When you want to review see here since it changes also some code: https://github.com/puppetlabs/puppetlabs-postgresql/commit/cd20e360c7247212bada2a35c3e914abab11a611 and https://github.com/cruelsmith/puppetlabs-postgresql/actions/runs/6100912336 (Yes the SLES and RHEL are failing since i test in my clone ENV but the important Ubuntu 22.04 is working)

PR #1480, opened 2023-08-29
Revert "puppetlabs/stdlib: Require 9.x"

bastelfreak:
Reverts puppetlabs/puppetlabs-postgresql#1449

To ensure we can do a minor release, we will try to revert this, do a minor release, apply it again.

This reverts e142268e9aa9f802a861d42bed2f7039bb8d6295

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -10,7 +10,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 9.0.0 < 10.0.0"
+      "version_requirement": ">= 5.2.0 < 10.0.0"
     },
     {
       "name": "puppetlabs/apt",
--- a/manifests/backup/pg_dump.pp
+++ b/manifests/backup/pg_dump.pp
@@ -68,13 +68,13 @@ class postgresql::backup::pg_dump (
 ) {
   # Install required packages
   if $package_name {
-    stdlib::ensure_packages($package_name)
+    ensure_packages($package_name)
   }
   if $install_cron {
     if $facts['os']['family'] == 'RedHat' {
-      stdlib::ensure_packages('cronie')
+      ensure_packages('cronie')
     } elsif $facts['os']['family'] != 'FreeBSD' {
-      stdlib::ensure_packages('cron')
+      ensure_packages('cron')
     }
   }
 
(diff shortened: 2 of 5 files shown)


bastelfreak:
I'm not sure if it's a good idea to have a check for latest modules that's enforced.

metadata.json
```diff
@@ -10,19 +10,19 @@
...
     },
     {
       "name": "puppet/systemd",
-      "version_requirement": ">= 4.0.1 < 6.0.0"
+      "version_requirement": ">= 4.0.1 < 5.0.0"
     },
     {
       "name": "puppetlabs/concat",
-      "version_requirement": ">= 4.1.0 < 10.0.0"
+      "version_requirement": ">= 4.1.0 < 9.0.0"
```
ekohl, line 25: I don't think you need to revert these 2.

ekohl (commented):
I agree that sometimes you just aren't ready for latest yet. Makes me wonder if there's a way to suppress it.

Technically you can call it compatible with the latest version, you just get deprecation warnings. On Puppet 7 that's no problem and on Puppet 8 you can work around it now (I think).

ekohl (commented):
:+1: content wise, but perhaps label it as `Bring back compatibility with stdlib < 9`.

bastelfreak:
now I'm not sure yet how we properly reflect this in the changelog, but the tests should pass.

ekohl (approved):
I wouldn't be too worried. Especially if we quickly release a new major after which brings proper compatibility.

deric (dismissed):

metadata.json
bastelfreak: I will revert those two. That allows us to skip this PR and the original one in the changelog.

bastelfreak:
this now reverts #1449 completely. That allows us to skip both PRs for the changelog.

metadata.json
bastelfreak: ah no! those were added in another PR so we don't need to revert it. #1448

ekohl (commented):
I miss the reference that `git revert e142268e9aa9f802a861d42bed2f7039bb8d6295` would generate, but otherwise :+1: for the approach.

bastelfreak:
I added the sha to the PR message

ekohl (approved):

PR #1486, opened 2023-08-30
Drop postgresql 8.1/RHEL5 specific code

bastelfreak:
## Summary

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -196,7 +196,6 @@ class postgresql::globals (
         '8'     => '10',
         '7'     => '9.2',
         '6'     => '8.4',
-        '5'     => '8.1',
         default => undef,
       },
     },
@@ -248,7 +247,6 @@ class postgresql::globals (
   }
 
   $default_postgis_version = $globals_version ? {
-    '8.1'   => '1.3.6',
     '8.4'   => '2.0',
     '9.0'   => '2.1',
     '9.1'   => '2.1',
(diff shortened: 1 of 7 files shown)


ekohl (commented):
If you do this, I'd also drop setting the default to 8.1 for some OSes and also remove test code.

And dropping 8.1 is probably fine, but I think Puppetlabs will want to continue supporting RHEL 6 while that's supported by Red Hat and that means keeping 8.4 I think.

bastelfreak:
I dropped all postgresql 8.1 or RHEL 5 specific code. The metadata.json only mentions RHEL7 and newer so I think that's fine.

Ramesh7 (approved):

ekohl:
> The metadata.json only mentions RHEL7 and newer so I think that's fine.

In the past Puppet has had this weird policy where they didn't list something in `metadata.json` but then did support it, sort of. I don't like it, because it makes life harder but since it's in their namespace I have tried to respect it.

alexjfisher:
@bastelfreak Following this change, I don't think https://github.com/puppetlabs/puppetlabs-postgresql/blob/04a3fbe6358b1f897444cb520e488ebdaff5cbaf/manifests/server/database.pp#L35 does anything?

PR #1487, opened 2023-08-30
Cleanup useless $connect_setting validation

bastelfreak:
In the past we checked if $connect_setting isn't undef. That is not required because the default is {} and the Datatype is hash. Undef isn't allowed.

## Summary

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/manifests/server/default_privileges.pp
+++ b/manifests/server/default_privileges.pp
@@ -37,7 +37,7 @@ define postgresql::server::default_privileges (
 ) {
   # If possible use the version of the remote database, otherwise
   # fallback to our local DB version
-  if $connect_settings != undef and 'DBVERSION' in $connect_settings {
+  if 'DBVERSION' in $connect_settings {
     $version = $connect_settings['DBVERSION']
   } else {
     $version = $postgresql::server::_version
@@ -62,15 +62,15 @@ define postgresql::server::default_privileges (
   #
   # Port, order of precedence: $port parameter, $connect_settings[PGPORT], $postgresql::server::port
   #
-  if $port != undef {
+  if $port {
     $port_override = $port
-  } elsif $connect_settings != undef and 'PGPORT' in $connect_settings {
+  } elsif 'PGPORT' in $connect_settings {
     $port_override = undef
   } else {
     $port_override = $postgresql::server::port
   }
 
-  if $target_role != undef {
+  if $target_role {
     $_target_role = " FOR ROLE ${target_role}"
     $_check_target_role = "/${target_role}"
   } else {
@@ -178,11 +178,11 @@ define postgresql::server::default_privileges (
     environment      => 'PGOPTIONS=--client-min-messages=error',
   }
 
-  if($role != undef and defined(Postgresql::Server::Role[$role])) {
+  if defined(Postgresql::Server::Role[$role]) {
     Postgresql::Server::Role[$role] -> Postgresql_psql["default_privileges:${name}"]
   }
 
-  if($db != undef and defined(Postgresql::Server::Database[$db])) {
+  if defined(Postgresql::Server::Database[$db]) {
     Postgresql::Server::Database[$db] -> Postgresql_psql["default_privileges:${name}"]
   }
 }
--- a/manifests/server/extension.pp
+++ b/manifests/server/extension.pp
@@ -77,9 +77,9 @@ define postgresql::server::extension (
   #
   # Port, order of precedence: $port parameter, $connect_settings[PGPORT], $postgresql::server::port
   #
-  if $port != undef {
+  if $port {
     $port_override = $port
-  } elsif $connect_settings != undef and 'PGPORT' in $connect_settings {
+  } elsif 'PGPORT' in $connect_settings {
     $port_override = undef
   } else {
     $port_override = $postgresql::server::port
--- a/manifests/server/grant.pp
+++ b/manifests/server/grant.pp
@@ -77,7 +77,7 @@ define postgresql::server::grant (
   #
   # Port, order of precedence: $port parameter, $connect_settings[PGPORT], $postgresql::server::port
   #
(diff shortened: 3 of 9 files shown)


manifests/server/default_privileges.pp
```diff
@@ -62,15 +62,15 @@
...
   #
-  if $port != undef {
+  if $port {
     $port_override = $port
-  } elsif $connect_settings != undef and 'PGPORT' in $connect_settings {
+  } elsif 'PGPORT' in $connect_settings {
     $port_override = undef
   } else {
     $port_override = $postgresql::server::port
   }
```
bastelfreak, line 71: the whole block doesn't make sense? Port is always set and cannot be undef. so `$port_override` is always set to `$port`? I don't thinky anybody uses `$connect_settings`. The whole setup seems broken?

manifests/server/default_privileges.pp
```diff
@@ -178,11 +178,11 @@
     environment      => 'PGOPTIONS=--client-min-messages=error',
   }
 
-  if($role != undef and defined(Postgresql::Server::Role[$role])) {
```
bastelfreak, line 181: `$role` cannot be undef

manifests/server/default_privileges.pp
```diff
@@ -178,11 +178,11 @@
...
+  if defined(Postgresql::Server::Role[$role]) {
     Postgresql::Server::Role[$role] -> Postgresql_psql["default_privileges:${name}"]
   }
 
-  if($db != undef and defined(Postgresql::Server::Database[$db])) {
```
bastelfreak, line 185: `$db` cannot be undef.

manifests/server/extension.pp
```diff
@@ -77,9 +77,9 @@
   #
   # Port, order of precedence: $port parameter, $connect_settings[PGPORT], $postgresql::server::port
   #
-  if $port != undef {
+  if $port {
     $port_override = $port
-  } elsif $connect_settings != undef and 'PGPORT' in $connect_settings {
+  } elsif 'PGPORT' in $connect_settings {
```
bastelfreak, line 82: Here it makes sense because $port is undef by default

manifests/server/grant.pp
```diff
@@ -77,7 +77,7 @@
   #
   # Port, order of precedence: $port parameter, $connect_settings[PGPORT], $postgresql::server::port
   #
-  if $port != undef {
+  if $port {
     $port_override = $port
   } elsif $connect_settings != undef and 'PGPORT' in $connect_settings {
     $port_override = undef
```
bastelfreak, line 83: whole block doesn't make sense because `$port` isn't undef?

SimonHoenscheid (approved):

bastelfreak:
I think the changes in this PR don't change the current behaviour so they should be safe to merge. But I think they point out the connect_setting implementation is currently broken, at least in some classes.

manifests/server/default_privileges.pp
```diff
@@ -62,15 +62,15 @@
...
     $port_override = $port
-  } elsif $connect_settings != undef and 'PGPORT' in $connect_settings {
+  } elsif 'PGPORT' in $connect_settings {
     $port_override = undef
   } else {
     $port_override = $postgresql::server::port
   }
 
-  if $target_role != undef {
+  if $target_role {
```
Ramesh7, line 73: I think we need update [here](https://github.com/puppetlabs/puppetlabs-postgresql/blob/ece77516a4f9f4badd18e25c8b3ae974ef5be459/manifests/server/default_privileges.pp#L36) as well, as its being optional param we should also update its data type?

manifests/server/grant_role.pp
```diff
@@ -39,7 +39,7 @@
     connect_settings => $connect_settings,
   }
 
-  if ! $connect_settings or empty($connect_settings) {
```
Ramesh7, line 42: Observed bit of confusing due to the DSL precedence as I thought negate (!) is for `$connect_settings or empty($connect_settings)` but I was wrong. Thanks to PDK console :p,
Thanks for the cleanup.

manifests/server/reassign_owned_by.pp
```diff
@@ -21,15 +21,10 @@
   $group     = $postgresql::server::group
   $psql_path = $postgresql::server::psql_path
 
-  #
-  # Port, order of precedence: $port parameter, $connect_settings[PGPORT], $postgresql::server::port
-  #
-  if $port {
-    $port_override = $port
-  } elsif $connect_settings != undef and 'PGPORT' in $connect_settings {
+  if 'PGPORT' in $connect_settings {
```
Ramesh7, line 24: 👍

Ramesh7 (approved):
except minor comments LGTM

manifests/server/role.pp
```diff
@@ -53,17 +53,17 @@
   #
   # Port, order of precedence: $port parameter, $connect_settings[PGPORT], $postgresql::server::port
   #
-  if $port != undef {
+  if $port {
```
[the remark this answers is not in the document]
bastelfreak: In this case, the priority doesn't change. `$port` is still `undef` by default. In some classes the logic seems to be broken because `$port` doesn't default to undef, but that problem exists on the main branch already.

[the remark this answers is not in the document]
bastelfreak: I think this is fine to merge because it only simplifies the code, it doesn't change the logic. but the logic is broken and should probably be checked in a dedicated issue.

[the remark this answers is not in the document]
ekohl: #1469 was aimed at making it more consistent. I haven't checked back on the most recent implementation, but you're right that it is inconsistent between files.

Issue #1467, opened 2023-08-25
Check if datatypes are reasonable


SimonHoenscheid:
## Use Case
It seems like some datatypes could be more accurate

## Describe the Solution You Would Like
correct datatypes if reasonable

## Additional Context
example:

`Variant[String[1], Stdlib::Port, Integer] $port = $postgresql::server::port,`

Since Integer accepts all integers and String[1] all non empty strings the specialized  Stdlib:Port will be without effect if it is a Pattern.
It does not make sense to have both Integer and Stdlib::Port since it is defined as Integer[0,65535] - should probably remove Integer from the variant unless it is possible to have a port that is > 65535.

ekohl:
> It does not make sense to have both Integer and Stdlib::Port since it is defined as Integer[0,65535] - should probably remove Integer from the variant unless it is possible to have a port that is > 65535.

Excellent example: it's impossible to have a port > 65535 because there simply aren't more bits in the TCP header. it's 16 bits and 2¹⁶ is 65536. If you start at 0 you end up with 65535 as the upper bound.

Similarly, we shouldn't accept strings for ports, unless the software allows specifying a range (like `10-20`). Even then a `Pattern` would be better.

SimonHoenscheid:
TIL! Thanks @ekohl

SimonHoenscheid:
Removing Strings support for the $port parameter would be a breaking change. I am fine with that. Currently there are tests which cover this behaviour too.

PR #1490, opened 2023-08-31
Provide a default for config_entry's path and enforce absolute path

SimonHoenscheid:


manifests/server/config_entry.pp
```diff
@@ -9,15 +9,8 @@
   Enum['present', 'absent']                               $ensure = 'present',
   String[1]                                               $key    = $name,
   Optional[Variant[String[1], Numeric, Array[String[1]]]] $value  = undef,
-  Variant[Boolean, String[1]]                             $path   = false
+  Variant[String[1], Stdlib::Absolutepath]                $path   = $postgresql::server::postgresql_conf_path
```
ekohl, line 12: Why `Stdlib::Absolutepath` if `String[1]` is also accepted? Then I'd simplify it to just `String[1]`. Having said that: does a relative path ever work or should it be simplified to an absolute path?

ekohl (commented):
Technically this doesn't make it configurable because it already was. It just changes the default.

Looking at the old implementation I think it would have been clearer if you used `Optional[String[1]]` $path = undef` and the `pick($path, $postgresql::server::postgresql_conf_path)` to get the actual value. Booleans for an empty value just feel awkward.

manifests/server/config_entry.pp
bastelfreak: Agreed. Thinking about it `Stdlib::Absolutepath` probably makes more sense. The `pick()` idea is nice, but I think the current implementation is easier to read.

ekohl: > The `pick()` idea is nice, but I think the current implementation is easier to read.

In theory it allows you to stop caring about the compile order, if you also add `include postgresql::server` but that's not there today either.

SimonHoenscheid: Regarding the Data Type I just took the Type from the `postgresql::server` class. I would suggest changing this in the whole module is part of #1467. I am happy to create  a PR for that.

SimonHoenscheid: PR ist open #1491

ekohl (commented):
This PR should be titled "Provide a default for config_entry's path and enforce absolute path", but maybe a bit shorter.

Technically this already enforces `$postgresql::server::postgresql_conf_path` to be an absolute path, so I think you should change the data type to `Stdlib::Absolutepath` here if you do so. Because it will be impossible to set it to a relative path. On the other hand, that never really was valid anyway. But I may be a bit pedantic here.

PR #1493, opened 2023-08-31
server::database_grant: Always set default user/group and expose port as parameter

bastelfreak:
This enables us to make it configureable. When a user doesn't specify it, the defaults will be passed, without breaking existing behaviour.


ekohl (commented):
One idea that gets lost here is that with `connect_settings` you could set defaults for a remote server. I think that makes it backwards incompatible.

bastelfreak:
@ekohl It's already broken because the defined resource postgresql::server::grant has a port param with a default value and won't use `$connect_settings` anyways.

bastelfreak:
@ekohl I could reverse the logic and always use $connect_settings when it's not empty?

SimonHoenscheid (dismissed):
Given the already broken conditions this LGTM

bastelfreak:
I made the port optional, so this should be fine now

SimonHoenscheid (approved):

PR #1496, opened 2023-08-31
server::grant: Prefer $connect_settings over default port

bastelfreak:
Previously the port parameter had a default value. This broke the logic
to identify the port. The port is now optional. If it's not set,
$connect_settings is checked for PGPORT. If that also doesn't exist, the
default from the server class is used.

### DIFF
--- a/manifests/server/grant.pp
+++ b/manifests/server/grant.pp
@@ -41,7 +41,7 @@ define postgresql::server::grant (
   Array[String[1],0]                             $object_arguments  = [],
   String                                         $psql_db           = $postgresql::server::default_database,
   String                                         $psql_user         = $postgresql::server::user,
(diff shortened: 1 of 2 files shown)


manifests/server/grant.pp
```diff
@@ -75,15 +75,8 @@
   }
 
   #
-  # Port, order of precedence: $port parameter, $connect_settings[PGPORT], $postgresql::server::port
```
ekohl, line 78: Isn't it confusing to have `$port` but not use it? In that sense I think the old logic was better: make `it Optional[Stdlib::Port]` and fall back to `connect_settings` and otherwise the server port.

bastelfreak: updated it

SimonHoenscheid (approved):

PR #1498, opened 2023-08-31
Prefer $connect_settings over explicit parameters

bastelfreak:
This is the outcome of a large refactoring in #1450 and discussions on IRC and slack. We noticed that the behaviour of `$connect_settings` is currently inconsistent. Sometimes it's preferred over explicity parameters, sometimes not.

Reminder: The idea of `$connect_settings` is to provide a hash with environment variable to tell puppet to manage a remote database instead of a local instance.

One example is: https://github.com/puppetlabs/puppetlabs-postgresql/blob/93386b48861ff41d5f47033afee592e0506526c5/manifests/server/grant.pp#L80-L86

```puppet
if $port {
  $port_override = $port
} elsif 'PGPORT' in $connect_settings {
  $port_override = undef
} else {
  $port_override = $postgresql::server::port
}
```

Here the local $port parameter is preferred over `$connect_settings`. The problem is that we now cannot set a default in the resource for `$port`. The default is hardcoded to `$postgresql::server::port`. This becomes. Other classes handle it in a different way:

https://github.com/puppetlabs/puppetlabs-postgresql/blob/93386b48861ff41d5f47033afee592e0506526c5/manifests/server/database.pp#L41-L46

```puppet
if 'PGPORT' in $connect_settings {
  $port_override = undef
} else {
  $port_override = $port
}
```

Here `$connct_settings` is checked first. It defaults to an empty hash.
If `PGPORT` isn't in it, `$port` is used. The logic is shorter and
enables us to expose $port as a parameter with a default value. With
this approach users can decide if they pass `$port` or
`$connect_settings`.

At the moment the remote database support is broken because the logic to
parse `$connect_settings` isn't consistent. The goal of this PR is to
unify the logic and prefer `$connect_settings` if it's provided by the
user.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -8,18 +8,18 @@
 
 #### Public Classes
 
-* [`postgresql::client`](#postgresqlclient): Installs PostgreSQL client software. Set the following parameters if you have a custom version you would like to install.
-* [`postgresql::globals`](#postgresqlglobals): Class for setting cross-class global overrides.
-* [`postgresql::lib::devel`](#postgresqllibdevel): This class installs postgresql development libraries.
-* [`postgresql::lib::docs`](#postgresqllibdocs): Installs PostgreSQL bindings for Postgres-Docs. Set the following parameters if you have a custom version you would like to install.
-* [`postgresql::lib::java`](#postgresqllibjava): This class installs the postgresql jdbc connector.
-* [`postgresql::lib::perl`](#postgresqllibperl): This class installs the perl libs for postgresql.
-* [`postgresql::lib::python`](#postgresqllibpython): This class installs the python libs for postgresql.
-* [`postgresql::server`](#postgresqlserver): This installs a PostgreSQL server
-* [`postgresql::server::contrib`](#postgresqlservercontrib): Install the contrib postgresql packaging.
-* [`postgresql::server::plperl`](#postgresqlserverplperl): This class installs the PL/Perl procedural language for postgresql.
-* [`postgresql::server::plpython`](#postgresqlserverplpython): This class installs the PL/Python procedural language for postgresql.
-* [`postgresql::server::postgis`](#postgresqlserverpostgis): Install the postgis postgresql packaging.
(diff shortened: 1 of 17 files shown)


bastelfreak:
I raised this to explain the current situation. This PR doesn't solve all the problems. If we agree on preferring `$connect_settings` I will update this PR and fix the other code pieces.

SimonHoenscheid (approved):

REFERENCE.md
```diff
@@ -8,18 +8,18 @@
...
-* [`postgresql::lib::java`](#postgresqllibjava): This class installs the postgresql jdbc connector.
-* [`postgresql::lib::perl`](#postgresqllibperl): This class installs the perl libs for postgresql.
-* [`postgresql::lib::python`](#postgresqllibpython): This class installs the python libs for postgresql.
-* [`postgresql::server`](#postgresqlserver): This installs a PostgreSQL server
-* [`postgresql::server::contrib`](#postgresqlservercontrib): Install the contrib postgresql packaging.
-* [`postgresql::server::plperl`](#postgresqlserverplperl): This class installs the PL/Perl procedural language for postgresql.
-* [`postgresql::server::plpython`](#postgresqlserverplpython): This class installs the PL/Python procedural language for postgresql.
-* [`postgresql::server::postgis`](#postgresqlserverpostgis): Install the postgis postgresql packaging.
+* [`postgresql::client`](#postgresql--client): Installs PostgreSQL client software. Set the following parameters if you have a custom version you would like to install.
```
Ramesh7, line 11: This will break navigation on Forge, and I think that's the reason where existing one skipped double `hyphen` (--).

Ramesh7: Here is the tracking ticket for Forge to fix the same : https://perforce.atlassian.net/browse/PF-2807

bastelfreak: It was already used in https://github.com/puppetlabs/puppetlabs-postgresql/pull/1500 but Simon used an older puppet-strings version by accident to regenerate it.

Ramesh7 (approved):
LGTM

PR #1501, opened 2023-09-04
pdksync - (CAT-1366) - Fix issue url from old jira to github

malikparvez:
(CAT-1366) - Fix issue url from old jira to github
pdk version: `2.7.0`


rajat-puppet (approved):
LGTM!

PR #1502, opened 2023-09-04
initdb better parameter naming

SimonHoenscheid:


SimonHoenscheid:
user and username were to close in my opinion so I renamed username to instance_username. I also set the username explicitly, which is no difference to the default if the username was not set before: 

```
--username=username
Selects the user name of the database superuser. This defaults to the name of the effective user running initdb. 
```

PR #1507, opened 2023-09-06
variables to parameters for tablespace/schema/reassign_owned_by

SimonHoenscheid:


manifests/server/schema.pp
```diff
@@ -9,35 +9,38 @@
...
+  Stdlib::Port         $port             = $postgresql::server::port,
+  String[1]            $user             = $postgresql::server::user,
+  String[1]            $group            = $postgresql::server::group,
+  Stdlib::Absolutepath $psql_path        = $postgresql::server::psql_path,
+  Stdlib::Absolutepath $module_workdir   = $postgresql::server::module_workdir,
 ) {
-  $user           = $postgresql::server::user
-  $group          = $postgresql::server::group
-  $psql_path      = $postgresql::server::psql_path
-  $version        = $postgresql::server::_version
```
bastelfreak, line 26: we should merge #1506 first

manifests/server/schema.pp
```diff
@@ -9,35 +9,38 @@
...
-  $module_workdir = $postgresql::server::module_workdir
-
   Postgresql::Server::Db <| dbname == $db |> -> Postgresql::Server::Schema[$name]
 
-  # If the connection settings do not contain a port, then use the local server port
```
bastelfreak, line 31: please bring that back

manifests/server/schema.pp
```diff
@@ -9,35 +9,38 @@
...
-    psql_group => $group,
-    psql_path  => $psql_path,
-    port       => $port_override,
-    cwd        => $module_workdir,
+    db               => $db,
+    psql_user        => $user,
+    psql_group       => $group,
+    psql_path        => $psql_path,
+    port             => $port_override,
+    cwd              => $module_workdir,
```
bastelfreak, line 43: I'm always unsure if we should changethe alignment here. but also odd that puppet-lint on main didn't complain :thinking:

bastelfreak: ah puppet-lint doesn't complain because this sets resource defaults, it's not a normal resource declaration. I suggest to undo the change here.

manifests/server/tablespace.pp
```diff
@@ -5,20 +5,25 @@
...
 # @param spcname Specifies the name of the tablespace.
 # @param connect_settings Specifies a hash of environment variables used when connecting to a remote server.
-# @param port the port of the postgresql instance that sould be used.
+# @param user Sets the OS user to run psql
+# @param group Sets the OS group to run psql
+# @param psql_path Sets path to psql command
+# @param module_workdir
+#   Specifies working directory under which the psql command should be executed.
+#   May need to specify if '/tmp' is on volume mounted with noexec option.
+# @param port Port to use when connecting.
```
bastelfreak, line 14: you deleted the documentation in line 7 and added a different one here, please undo that

bastelfreak (approved):

PR #1512, opened 2023-09-06
Flexible password encryption in pg hba conf

### LINKED ISSUE #1465: Default HBA does not take scram into account

cruelsmith:
## Summary
* ensure that default password encryption is also used in default pg hba conf
* ensure that `postgresql::postgresql_password` respect the module default password encryption

## Additional Context

## Related Issues (if any)
fixes #1465

replace #1479

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -4429,6 +4429,8 @@ Data type: `Optional[Optional[Postgresql::Pg_password_encryption]]`
 
 Set type for password hash
 
+Default value comes from `postgresql::params::password_encryption` and changes based on the `postgresql::globals::version`.
+
 ##### `salt`
 
 Data type: `Optional[Optional[Variant[String[1], Integer]]]`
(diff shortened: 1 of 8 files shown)


bastelfreak (approved):
looks good to me. maybe label that as bugfix like 'fix auth hashing method for old postgresql versions'?

SimonHoenscheid (approved):

PR #1514, opened 2023-09-07
Drop code for Debian without systemd

bastelfreak:
FreeBSD 9.4/9.5 are EoL since ages and not listed as supported in metadata.json. Also no linux operating system without Systemd is supported, so we can drop the old code.

## Summary

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -164,13 +164,8 @@ class postgresql::params inherits postgresql::globals {
       $bindir                 = pick($bindir, "/usr/lib/postgresql/${version}/bin")
       $datadir                = pick($datadir, "/var/lib/postgresql/${version}/main")
       $confdir                = pick($confdir, "/etc/postgresql/${version}/main")
-      if pick($service_provider, $facts['service_provider']) == 'systemd' {
-        $service_reload = "systemctl reload ${service_name}"
-        $service_status = pick($service_status, "systemctl status ${service_name}")
-      } else {
-        $service_reload = "service ${service_name} reload"
-        $service_status = pick($service_status, "service ${service_name} status")
-      }
+      $service_reload         = "systemctl reload ${service_name}"
+      $service_status         = pick($service_status, "systemctl status ${service_name}")
       $psql_path              = pick($psql_path, '/usr/bin/psql')
       $postgresql_conf_mode   = pick($postgresql_conf_mode, '0644')
     }


manifests/params.pp
```diff
@@ -104,8 +99,6 @@
 
       if $postgresql::globals::postgis_package_name {
         $postgis_package_name = $postgresql::globals::postgis_package_name
-      } elsif $facts['os']['release']['major'] == '5' {
-        $postgis_package_name = 'postgis'
       } elsif $postgis_version and versioncmp($postgis_version, '2') < 0 {
```
ekohl, line 102: Is there still PostGIS 1 support for Red Hat distros? Feels like this could be simplified.

manifests/params.pp
```diff
@@ -169,13 +162,8 @@
       $bindir                 = pick($bindir, "/usr/lib/postgresql/${version}/bin")
       $datadir                = pick($datadir, "/var/lib/postgresql/${version}/main")
       $confdir                = pick($confdir, "/etc/postgresql/${version}/main")
-      if pick($service_provider, $facts['service_provider']) == 'systemd' {
```
ekohl, line 172: > Also no linux operating system without Systemd is supported, so we can drop the old code.

Technically Debian can still run without systemd and the previous code did handle that at essentially no cost.

manifests/params.pp
bastelfreak: I have zero knowledge about postgis so I didn't want to touch it

bastelfreak: ack, I dropped it

spec/spec_helper_local.rb
```diff
@@ -155,7 +155,7 @@ def param(type, title, param)
 # See https://github.com/voxpupuli/voxpupuli-test/blob/master/lib/voxpupuli/test/facts.rb
 add_custom_fact :service_provider, ->(_os, facts) do
```
ekohl, line 156: A quick look at the `service_provider` fact:
```console
$ rg service_provider manifests/
manifests/server/service.pp
8:    service_provider => $postgresql::server::service_provider,

manifests/server/instance/service.pp
7:# @param service_provider Overrides the default PostgreSQL service provider.
25:  Optional[String[1]]                          $service_provider = $postgresql::server::service_provider,
40:      provider  => $service_provider,

manifests/globals.pp
22:# @param service_provider Overrides the default PostgreSQL service provider.
119:  Optional[String[1]] $service_provider            = undef,

manifests/server.pp
19:# @param service_provider Overrides the default PostgreSQL service provider.
131:  Optional[String[1]]                                $service_provider             = $postgresql::params::service_provider,

manifests/params.pp
20:  $service_provider             = $postgresql::globals::service_provider

manifests/server/instance/config.pp
247:  if $facts['os']['family'] in ['RedHat', 'Gentoo'] and $facts['service_provider'] == 'systemd' {
```

The last one probably can't be simplified because Gentoo also supports non-systemd (like openrc).

manifests/params.pp
```diff
@@ -204,18 +186,9 @@
...
-        }
-        default: {
-          $user                 = pick($user, 'postgres')
-          $group                = pick($group, 'postgres')
-          $datadir              = pick($datadir, "/var/db/postgres/data${version}")
-        }
-      }
+      $user                 = pick($user, 'postgres')
+      $group                = pick($group, 'postgres')
+      $datadir              = pick($datadir, "/var/db/postgres/data${version}")
```
smortex, line 191: Regarding the FreeBSD part, it's :100: to me.

manifests/params.pp
```diff
@@ -97,15 +97,7 @@
...
-      if $postgresql::globals::postgis_package_name {
-        $postgis_package_name = $postgresql::globals::postgis_package_name
-      } elsif $facts['os']['release']['major'] == '5' {
-        $postgis_package_name = 'postgis'
-      } elsif $postgis_version and versioncmp($postgis_version, '2') < 0 {
-        $postgis_package_name = "postgis${package_version}"
-      } else {
-        $postgis_package_name = "postgis2_${package_version}"
-      }
+      $postgis_package_name = pick($postgresql::globals::postgis_package_name, "postgis2_${package_version}")
```
ekohl, line 100: I think this also requires its own PR. If I look at EL8 & EL9 then I don't find postgis packages. In Fedora it's just called `postgis`. No idea what it's called in the official postgresql repositories.

manifests/params.pp
bastelfreak: @Ramesh7 can you tell me if we can drop support for Debian without systemd as init system?

ekohl: I've at least changed the PR title. I'd like the final commit message (if it is merged) to also reflect it.

manifests/params.pp
```diff
@@ -164,13 +164,8 @@
...
       $confdir                = pick($confdir, "/etc/postgresql/${version}/main")
-      if pick($service_provider, $facts['service_provider']) == 'systemd' {
-        $service_reload = "systemctl reload ${service_name}"
-        $service_status = pick($service_status, "systemctl status ${service_name}")
-      } else {
-        $service_reload = "service ${service_name} reload"
-        $service_status = pick($service_status, "service ${service_name} status")
-      }
+      $service_reload         = "systemctl reload ${service_name}"
+      $service_status         = pick($service_status, "systemctl status ${service_name}")
```
ekohl, line 168: Just opening the debate: isn't this the default implementation? Can we make `service_status` undef?

spec/spec_helper_local.rb
bastelfreak: yeah. Ideally the whole file would go away :(

Ramesh7 (approved):
lgtm

PR #1515, opened 2023-09-11
Fix password_encryption for DBVERSION in server::role

cruelsmith:
## Summary
After #1512 the `postgresql::postgresql_password()` must not default to 'md5' anymore which creates a edge case in `postgresql::server::role`. When `DBVERSION` in paramter `connect_settings` is set to a version before 14 and the global version is 14 or above, the used `hash` parameter of `postgresql::postgresql_password()`  will be set to undef instead of 'md5'.

This PR fixes that edge case by setting the wanted hash type and do not relay anymore on the default hash type of `postgresql::postgresql_password()` itself.

## Additional Context

## Related Issues (if any)
Fixes #1512

## Checklist
- [ ] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.

### DIFF
--- a/manifests/server/role.pp
+++ b/manifests/server/role.pp
@@ -142,7 +142,7 @@ define postgresql::server::role (
     $_hash = if $hash {
       $hash
     } elsif $connect_settings != undef and 'DBVERSION' in $connect_settings {
-      if (versioncmp($version, '14') >= 0) { 'scram-sha-256' } else { undef }
+      versioncmp($version, '14') ? { -1 => 'md5', default => 'scram-sha-256' }
     } else {
       $postgresql::server::password_encryption
     }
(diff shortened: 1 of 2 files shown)


bastelfreak:
@cruelsmith can you add a unit test to ensure the logic correct now?

manifests/server/role.pp
```diff
@@ -142,7 +142,7 @@
     $_hash = if $hash {
       $hash
     } elsif $connect_settings != undef and 'DBVERSION' in $connect_settings {
-      if (versioncmp($version, '14') >= 0) { 'scram-sha-256' } else { undef }
+      versioncmp($version, '14') ? { -1 => 'md5', default => 'scram-sha-256' }
```
bastelfreak, line 145: derp, I missed that earlier :(

cruelsmith:
Added acceptance unit test that now works as wanted. Needed some test since i was not sure how to do it.

The failing `(RedHat-9, puppet7-nightly)` run is unrelated and a general issue that i saw also on other runs. Some times the `Install module` phase is simply failing.

bastelfreak (approved):

PR #1516, opened 2023-09-25
add $manage_selinux as a parameter, keep default, simpler condition

SimonHoenscheid:


manifests/server/instance/config.pp
```diff
@@ -76,6 +77,7 @@
   Postgresql::Pg_password_encryption             $password_encryption          = $postgresql::server::password_encryption,
   Optional[Postgresql::Pg_password_encryption]   $pg_hba_auth_password_encryption = $postgresql::server::pg_hba_auth_password_encryption,
   Optional[String]                               $extra_systemd_config         = $postgresql::server::extra_systemd_config,
+  Optional[Boolean]                              $manage_selinux               = $postgresql::server::manage_selinux,
```
bastelfreak, line 80: it's a Boolean in postgresql::server so it cannot be Undef so `Optional` doesn't make sense and we can switch to `Boolean`?

bastelfreak (approved):

PR #1521, opened 2023-09-27
postgis: Drop EL5 leftovers and fix package name for Fedora

bastelfreak:
## Summary

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -99,6 +99,8 @@ class postgresql::params inherits postgresql::globals {
 
       if $postgresql::globals::postgis_package_name {
         $postgis_package_name = $postgresql::globals::postgis_package_name
+      } elsif $facts['os']['name'] == 'Fedora' {
+        $postgis_package_name = 'postgis'
       } elsif $postgis_version and versioncmp($postgis_version, '2') < 0 {
         $postgis_package_name = "postgis${package_version}"
       } else {
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -99,8 +99,6 @@ class postgresql::params inherits postgresql::globals {
 
       if $postgresql::globals::postgis_package_name {
         $postgis_package_name = $postgresql::globals::postgis_package_name
-      } elsif $facts['os']['release']['major'] == '5' {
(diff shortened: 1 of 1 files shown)


SimonHoenscheid (dismissed):

manifests/params.pp
```diff
@@ -99,8 +99,6 @@
 
       if $postgresql::globals::postgis_package_name {
         $postgis_package_name = $postgresql::globals::postgis_package_name
-      } elsif $facts['os']['release']['major'] == '5' {
```
ekohl, line 102: Perhaps change this line to be `elsif $facts['os']['name'] == 'Fedora'` since that still uses the postgis package name.

manifests/params.pp
```diff
@@ -99,8 +99,6 @@
...
-        $postgis_package_name = 'postgis'
       } elsif $postgis_version and versioncmp($postgis_version, '2') < 0 {
```
ekohl, line 102: As I noted elsewhere: in EL 8 & EL 9 I didn't find a postgis package. For the official packages I see there's postgis32 and postgis33 (https://download.postgresql.org/pub/repos/yum/15/redhat/rhel-9-x86_64/) so the else condition isn't quite true either.

manifests/params.pp
bastelfreak: done

manifests/params.pp
bastelfreak: I'm not really sure how we should move forward. It seems like there are no acceptance tests for this code and I don't know on which systems they work / are supposed to work. My goal was to simplify the conditions here and clean it up. I can raise a ticket about the mismatch.

ekohl: Sounds good to me.

ekohl (approved):

manifests/params.pp
bastelfreak: raised https://github.com/puppetlabs/puppetlabs-postgresql/issues/1522

PR #1529, opened 2023-10-11
Fix the `postgresql::postgresql_password()` function

### LINKED ISSUE #1528: Regression: postgresql::postgresql_password() broken on 10.0.0
## Describe the Bug
Updating to 10.0.0 breaks the usage of  `postgresql::postgresql_password()`. Might be related to #1515 (`md5` not supported on modern systems)

## Steps to Reproduce
After updating to 10.0.0 all calls to `postgresql::postgresql_password()` fail with `Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Function Call, failed to coerce org.jruby.RubySymbol to java.lang.String`

## Environment
 - Version of postgresql-Mod: 10.0.0
 - Version of puppet: 8.2.0
 - Platform: Rocky Linux 9

## Additional Context
Did not yet debug all of the code changes, but I suspect the switch of using `md5` as the default hash-method now. Seeing a Java-error there rings a least a bell with the modern security policies that are part of newer systems like EL9.

smortex:
The changes in #1512 introduced a regression when a hash value is not
explicitly passed:

```
Puppet Server Error: Evaluation Error: Error while evaluating a Function Call, failed to coerce org.jruby.RubySymbol to java.lang.String
```

Fixes #1528

### DIFF
--- a/lib/puppet/functions/postgresql/postgresql_password.rb
+++ b/lib/puppet/functions/postgresql/postgresql_password.rb
@@ -30,7 +30,7 @@ Puppet::Functions.create_function(:'postgresql::postgresql_password') do
   end
 
   def default_impl(username, password, sensitive = false, hash = nil, salt = nil)
-    hash = call_function(:'postgresql::default', 'password_encryption') if hash.nil?
+    hash = call_function('postgresql::default', 'password_encryption') if hash.nil?
     password = password.unwrap if password.respond_to?(:unwrap)
     if password.is_a?(String) && password.match?(%r{^(md5[0-9a-f]{32}$|SCRAM-SHA-256\$)})
       return Puppet::Pops::Types::PSensitiveType::Sensitive.new(password) if sensitive


bastelfreak:
@smortex thanks for the fix, I noticed the error as well. Any chance you can add an easy test for it?

smortex:
> @smortex thanks for the fix, I noticed the error as well. Any chance you can add an easy test for it?

Not sure how to do this given `call_function` belongs to Puppet itself.  Do you have an idea how we can improve this (other that adding checks on the Puppet side)?

Also while investigating this I found another bug, which I am working on ATM.  Don't release a 10.0.1 too soon :upside_down_face: !

smortex:
> @smortex thanks for the fix, I noticed the error as well. Any chance you can add an easy test for it?

Also there is already extensive testing of the `postgresql::postgresql_password()` function where the call without explicit hash does not fail. Maybe some code run by jruby fail but the bad behavior is not triggered by MRI :shrug:?  I have not found how to reproduce this outside of the context of puppetserver compiling a catalog.

kenyon:
I still get this error with this PR's branch, is that the other bug you found @smortex?

`Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: no parameter named 'instance' (file: /etc/puppetlabs/code/environments/production/modules/postgresql/manifests/server/extension.pp, line: 83) on Postgresql_psql[puppetdb: CREATE EXTENSION "pg_trgm"] (file: /etc/puppetlabs/code/environments/production/modules/postgresql/manifests/server/extension.pp, line: 83) on node lithium.kenyonralph.com`

kenyon:
Ah nevermind, just needed to run `sudo puppet generate types` (non-PE). Need to add that to my regular deploy command.

SimonHoenscheid (approved):

Ramesh7 (approved):

PR #1530, opened 2023-10-11
Fix `postgresql::default()` return value for unknown parameters

smortex:
When calling `postgresql::default()` with a name that has no corresponding parameter, the function returns the string "postgresql::globals::<parameter-name>" which is probably not intended given the comment above the code.

The usage of pick seems to indicate that an exception should be raised when a parameter is not found in `params` nor `globals.pp`, so adjust the code accordingly.

### DIFF
--- a/functions/default.pp
+++ b/functions/default.pp
@@ -10,6 +10,5 @@ function postgresql::default(
 
   #search for the variable name in params first
   #then fall back to globals if not found
-  pick( getvar("postgresql::params::${parameter_name}"),
-  "postgresql::globals::${parameter_name}")
+  pick(getvar("postgresql::params::${parameter_name}"), getvar("postgresql::globals::${parameter_name}"))
 }
--- a/functions/default.pp
+++ b/functions/default.pp
@@ -8,7 +8,7 @@ function postgresql::default(
 ) {
   include postgresql::params
 
-  #search for the variable name in params first
-  #then fall back to globals if not found
-  pick(getvar("postgresql::params::${parameter_name}"), getvar("postgresql::globals::${parameter_name}"))
+  # Search for the variable name in params.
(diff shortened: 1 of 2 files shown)


functions/default.pp
```diff
@@ -8,8 +8,7 @@ function postgresql::default(
 ) {
   include postgresql::params
 
-  #search for the variable name in params first
-  #then fall back to globals if not found
-  pick( getvar("postgresql::params::${parameter_name}"),
-  "postgresql::globals::${parameter_name}")
+  # Search for the variable name in params.
+  # params inherits from globals, so it will also catch these variables.
+  pick(getvar("postgresql::params::${parameter_name}"))
```
bastelfreak, line 13: is there a situation where a variable is defined in globales but not params? with your change we wouldn't find a value anymore.

smortex: params inherits from globals. `postgresql::globals::default_connect_settings` value can be read with `$postgresql::params::default_connect_settings` as shown by the unit test ;-)

bastelfreak: ah right :D

bastelfreak (approved):

SimonHoenscheid (approved):

deric (approved):

functions/default.pp
ekohl, line 13: Late to the party, but what's the point of using `pick()` with a single parameter? Isn't just `getvar()` the same?

smortex: `getvar('nothing_here')` returns `undef`.  Since pick was already there, I assumed we wanted to have a strong failure with an exception in that case.

ekohl: I suspected that, but good to hear that explicitly.

Issue #1533, opened 2023-10-14
Postgresql_conf[ssl] resource broken with Server Error: no parameter named 'key' after bump to 10.0.1


corporate-gadfly:
## Describe the Bug
```
Error: Could not retrieve catalog from remote server:
Error 500 on SERVER: Server Error: no parameter named 'key'
(file: /etc/puppetlabs/code/environments/production/modules/postgresql/manifests/server/config_entry.pp, line: 87)
on Postgresql_conf[ssl]
(file: /etc/puppetlabs/code/environments/production/modules/postgresql/manifests/server/config_entry.pp, line: 87)
on node xxx.tld
```
For reference: https://github.com/puppetlabs/puppetlabs-postgresql/blob/05bc7156d1d1ac841e09c632a41cbc97af54157c/manifests/server/config_entry.pp#L87-L94

I don't use `puppetlabs-postgresql` directly. Instead, my usage is via `puppetlabs-puppetdb` (version 7.13.0) as follows:
```
  class { 'puppetdb::globals':
    version => '8.1.1-1.el8',
  }
  class { 'puppetdb':
    open_ssl_listen_port    => true,
    postgres_version        => '11',
    postgresql_ssl_on       => true,
    database_host           => $trusted['certname'],
    database_listen_address => '0.0.0.0',

    # set low TTL values resulting in low resource usage
    # (as we are not interested in reports/facts/dashboard)

    # Mark as 'expired' nodes that haven't seen any activity
    # (no new catalogs, facts, or reports) in the specified amount of time.
    # if unset, the default is 7d
    node_ttl                => '1d',
    # Automatically delete nodes that have been deactivated
    # or expired for the specified amount of time.
    # if unset, the default is 14d
    node_purge_ttl          => '1d',
    # Automatically delete reports that are older than
    # the specified amount of time.
    # if unset, the default is 14d
    report_ttl              => '1d',
    # JAVA args
    java_args               => {
      '-Xmx' => '2g',
      '-Xms' => '2g',
    },
  }

```
Feel free to close if this should be filed against puppetdb.

bastelfreak:
Hi @corporate-gadfly ,
any chance you didn't run `puppet generate types` after the code deployment? If you are using r10k, you can enable it in the config file (/etc/puppetlabs/r10k/r10k.yaml):

```yaml
---
pool_size: 8
deploy:
  generate_types: true
  exclude_spec: true
  incremental: true
[...]

PR #1534, opened 2023-10-18
add test to cover basic server configuration

SimonHoenscheid:
## Summary
create test to cover #1533 behaviour

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/spec/acceptance/postgres_server_spec.rb
+++ b/spec/acceptance/postgres_server_spec.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'spec_helper_acceptance'
+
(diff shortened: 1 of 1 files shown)


spec/acceptance/postgres_server_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    class { 'postgresql::globals':
+      encoding            => 'UTF-8',
+      locale              => 'en_US.UTF-8',
+      manage_package_repo => true,
+      version             => '14',
+    }
+    include postgresql::server
+  MANIFEST
+  it 'installs postgres' do
+    apply_manifest(pp, catch_failures: true)
```
ekohl, line 18: Should it run a second time (with `catch_changes: true` instead) as an idempotency check?

PR #1538, opened 2023-10-19
(#1532) Replace ParserError with Puppet::Error

bastelfreak:
I'm not sure how we ended up with ParserError in the provider. This exception doesn't exist in Ruby. Puppet ships ther own exception, Puppet::Error. It probably makes sense to raise that instead.

Fixes 179472ba261a7d0847fdb93fdcb1d43741c4cdde

## Summary
Provide a detailed description of all the changes present in this pull request.

## Additional Context
Add any additional context about the problem here. 
- [ ] Root cause and the steps to reproduce. (If applicable)
- [ ] Thought process behind the implementation.

## Related Issues (if any)
Mention any related issues or pull requests.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified. (For example `puppet apply`)

Issue #1532, opened 2023-10-12
Postgresql::Server::Config_entry uninitialized constant ParserError


jlindquist-godaddy:
## Describe the Bug
```
Error: /Stage[main]/Postgresql::Server/Postgresql::Server::Config_entry[shared_buffers]/Postgresql_conf[shared_buffers]: Could not evaluate: uninitialized constant ParserError
Did you mean?  ParseError
Notice: /Stage[main]/Postgresql::Server::Service/Postgresql::Server::Instance::Service[main]/Anchor[postgresql::server::service::begin::main]: Dependency Postgresql_conf[shared_buffers] has failures: true
```

## Expected Behavior
Catalog compilation does not fail, and config entry is properly validated and set as necessary

## Steps to Reproduce
Include server config entries from Hiera
```
postgresql::server::config_entries:
  max_connections: 300
  shared_buffers: 4GB
```

## Environment
 - Version 10.0.1
 - Platform CentOS 7

## Additional Context
It looks like we're triggering the error [here](https://github.com/puppetlabs/puppetlabs-postgresql/blob/v10.0.1/lib/puppet/provider/postgresql_conf/ruby.rb#L76-L84) based on a config such as this which does have multiple entries that need to be resolved, but it should raise a proper error:
```
# grep shared_buffers /var/lib/pgsql/14/data/postgresql.conf
shared_buffers = 128MB # min 128kB
#wal_buffers = -1			# min 32kB, -1 sets based on shared_buffers
shared_buffers = 4GB
```

bastelfreak:
@jlindquist-godaddy can you take a look  at https://github.com/puppetlabs/puppetlabs-postgresql/pull/1538 and test it?

jlindquist-godaddy:
Looks good, thank you!

ekolkman:
I just ran into this bug with version 10.0.3 of the module.
Easily reproducable: have multiple listen_addresses lines in the postgresql.conf file (values don't matter) and configure a server with a listen_addresses paramter set to a value.
(Postgres itself usually doesn't mind multiple entries for the same parameter, as only the last one is used. But this version of the module doesn't accept it and now gives an error about giving the error.)

ekolkman:
Problem persists in 10.1.0.
Reproducable by adding any entry in the postgresql.conf a second time (doesn't matter if the value differs).
[...]

PR #1540, opened 2023-10-19
(#1532) Replace ParserError with Puppet::Error 

bastelfreak:
I'm not sure how we ended up with ParserError in the
provider. This exception doesn't exist in Ruby.
Puppet ships ther own exception, Puppet::Error. It
probably makes sense to raise that instead.

Fixes https://github.com/bastelfreak/puppetlabs-postgresql/commit/179472ba261a7d0847fdb93fdcb1d43741c4cdde

Alternative implementation for https://github.com/puppetlabs/puppetlabs-postgresql/pull/1538

## Summary

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/lib/puppet/provider/postgresql_conf/ruby.rb
+++ b/lib/puppet/provider/postgresql_conf/ruby.rb
@@ -74,7 +74,7 @@ Puppet::Type.type(:postgresql_conf).provide(:ruby) do
   # check, if resource exists in postgresql.conf file
   def exists?
     select = parse_config.select { |hash| hash[:key] == resource[:key] }
-    raise ParserError, "found multiple config items of #{resource[:key]} found, please fix this" if select.length > 1
+    raise Puppet::Error, "found multiple config items of #{resource[:key]}, please fix this" if select.length > 1
     return false if select.empty?
 
     @result = select.first
--- a/spec/unit/provider/postgresql_conf/ruby_spec.rb
+++ b/spec/unit/provider/postgresql_conf/ruby_spec.rb
@@ -4,13 +4,14 @@ require 'spec_helper'
 provider_class = Puppet::Type.type(:postgresql_conf).provider(:ruby)
 
 describe provider_class do
-  let(:resource) { Puppet::Type.type(:postgresql_conf).new(name: 'foo', value: 'bar') }
+  let(:resource) { Puppet::Type.type(:postgresql_conf).new(name: 'foo', key: 'foo', value: 'bar') }
   let(:provider) { resource.provider }
 
   before(:each) do
     allow(provider).to receive(:file_path).and_return('/tmp/foo')
     allow(provider).to receive(:read_file).and_return('foo = bar')
     allow(provider).to receive(:write_file).and_return(true)
+    allow(provider).to receive(:resource).and_return(key: 'your_key', line_number: 1, value: 'foo')
   end
   # rubocop:enable RSpec/ReceiveMessages
 
@@ -26,8 +27,27 @@ describe provider_class do
     expect(provider).to respond_to(:add_header)
   end
 
-  it 'has a method exists?' do
-    expect(provider).to respond_to(:exists?)
+  describe '#exists?' do
+    it 'returns true when a matching config item is found' do
+      config_data = [{ key: 'your_key', value: 'your_value' }]
+      expect(provider).to receive(:parse_config).and_return(config_data)
+
+      expect(provider.exists?).to be true
+    end
+
+    it 'returns false when no matching config item is found' do
+      config_data = [{ key: 'other_key', value: 'other_value' }]
+      expect(provider).to receive(:parse_config).and_return(config_data)
+
+      expect(provider.exists?).to be false
+    end
+
+    it 'raises an error when multiple matching config items are found' do
+      config_data = [{ key: 'your_key', value: 'value1' }, { key: 'your_key', value: 'value2' }]
+      expect(provider).to receive(:parse_config).and_return(config_data)
+
+      expect { provider.exists? }.to raise_error(Puppet::Error, 'found multiple config items of your_key, please fix this')
+    end
   end
 
   it 'has a method create' do


spec/unit/provider/postgresql_conf/ruby_spec.rb
```diff
@@ -29,6 +29,32 @@
   it 'has a method exists?' do
     expect(provider).to respond_to(:exists?)
   end
+  context '#exists?' do
+    before(:each) do
+      # Stub the resource method to return a specific value
+      allow(provider).to receive(:resource).and_return(key: 'your_key')
```
ekohl, line 35: Would it be better to initialize it properly on line 7?

bastelfreak: I considered it, but I'm unsure what's best practice. I assumed that this is currently only required here, so it makes sense to initialize it here and not globally? I'm fine with both. if you like it I move it.

ekohl: Essentially `provider` is the subject and stubbing/mocking AFAIK your subject is frowned upon. I'd initialize it properly. I also don't think the provider can work well without the `key` parameter so it's the right thing.

bastelfreak: ah yes, sounds reasonable.

spec/unit/provider/postgresql_conf/ruby_spec.rb
```diff
@@ -11,6 +11,7 @@
     allow(provider).to receive(:file_path).and_return('/tmp/foo')
     allow(provider).to receive(:read_file).and_return('foo = bar')
     allow(provider).to receive(:write_file).and_return(true)
+    allow(provider).to receive(:resource).and_return(key: 'your_key', line_number: 1, value: 'foo')
```
ekohl, line 14: I meant to change line 7 to:
```ruby
let(:resource) { Puppet::Type.type(:postgresql_conf).new(name: 'foo', key: 'foo', value: 'bar') }
```

spec/unit/provider/postgresql_conf/ruby_spec.rb
```diff
@@ -29,6 +30,28 @@
   it 'has a method exists?' do
     expect(provider).to respond_to(:exists?)
   end
+  context '#exists?' do
```
ekohl, line 33: I'd write it as:
```suggestion
  describe '#exists?' do
```

That it responds to `exists?` is a weaker test than the ones you're adding. It's also custom to use `describe` if you're describing a method.

spec/unit/provider/postgresql_conf/ruby_spec.rb
bastelfreak: ah okay. will update

bastelfreak: mhm this line is still requires to get the tests passing

ekohl: This confuses me and now I wonder what's wrong. Perhaps tomorrow I'll have some idea.

SimonHoenscheid (dismissed):

spec/unit/provider/postgresql_conf/ruby_spec.rb
ekohl: So I think I know what is going on. Line 7 instantiates a new Puppet type with the desired state. The provider has the actual state. You're mocking that here. So by mocking this, you know it is already in place.

So at the very least I'd expect you to only mock it when you want it to exist, not before each test.

It also makes me wonder about something else. `key` is a parameter, not a property. So I'm not sure it should be loaded from the provider.

lib/puppet/provider/postgresql_conf/ruby.rb
```diff
@@ -75,7 +75,7 @@ def write_config(file, lines)
   # check, if resource exists in postgresql.conf file
   def exists?
     select = parse_config.select { |hash| hash[:key] == resource[:key] }
-    raise ParserError, "found multiple config items of #{resource[:key]} found, please fix this" if select.length > 1
+    raise Puppet::Error, "found multiple config items of #{resource[:key]} found, please fix this" if select.length > 1
```
rwaffen, line 78: I think one `found` is enough for this sentence?

bastelfreak: indeed, updated it.

lib/puppet/provider/postgresql_conf/ruby.rb
```diff
@@ -74,7 +74,7 @@ def write_config(file, lines)
   # check, if resource exists in postgresql.conf file
   def exists?
     select = parse_config.select { |hash| hash[:key] == resource[:key] }
-    raise ParserError, "found multiple config items of #{resource[:key]} found, please fix this" if select.length > 1
+    raise Puppet::Error, "found multiple config items of #{resource[:key]}, please fix this" if select.length > 1
```
joshcooper, line 77: FWIW, providers can call the `fail` method which will raise `Puppet::Error`, for example see the base exec provider: https://github.com/puppetlabs/puppet/blob/fce49baa3767fffb000f1b26d3a7b4edbec58999/lib/puppet/provider/exec.rb#L105

spec/unit/provider/postgresql_conf/ruby_spec.rb
```diff
@@ -26,8 +27,27 @@
...
+      expect(provider).to receive(:parse_config).and_return(config_data)
+
+      expect(provider.exists?).to be false
+    end
+
+    it 'raises an error when multiple matching config items are found' do
+      config_data = [{ key: 'your_key', value: 'value1' }, { key: 'your_key', value: 'value2' }]
+      expect(provider).to receive(:parse_config).and_return(config_data)
+
+      expect { provider.exists? }.to raise_error(Puppet::Error, 'found multiple config items of your_key found, please fix this')
```
ekohl, line 49:
```suggestion
      expect { provider.exists? }.to raise_error(Puppet::Error, 'found multiple config items of your_key, please fix this')
```

spec/unit/provider/postgresql_conf/ruby_spec.rb
ekohl: This line is now redundant, right?
```suggestion
```

lib/puppet/provider/postgresql_conf/ruby.rb
bastelfreak: I just wanted to fix a simple typo. I don't know how to properly write tests for the `fail()` method. If someone wants to do this: feel free.

Ramesh7 (approved):

PR #1542, opened 2023-10-23
Extend postgresql_conf to support multiple config files using ParsedFile

ekohl:
To avoid a custom postgresql_conf implementation, this reverts back to ParsedFile. It relies on https://github.com/puppetlabs/puppet/pull/9130 since there were bugs in it, which is why it's now a draft.


lib/puppet/type/postgresql_conf.rb
```diff
@@ -2,40 +2,36 @@
...
+    [ /^(.+\.conf):([\w.]+)$/m, [ [:target], [:key] ] ],
+    # TODO: make target optional
+    #[ /^([\w.]+)$/m, [ [:key] ] ],
+  ]
   end
 
-  newparam(:key) do
-    desc 'The Postgresql parameter to manage.'
+
+  newparam(:key, namevar: true) do
```
ekohl, line 17: I kept `key` instead of `name` to force me to implement `ParsedFile` agnostic of the `name` parameter. At this point I think it could revert back to `name`.

lib/puppet/type/postgresql_conf.rb
```diff
@@ -2,40 +2,36 @@
...
   ensurable
 
-  newparam(:name) do
-    desc 'A unique title for the resource.'
-    newvalues(%r{^[\w.]+$})
+  def self.title_patterns
+  [
+    [ /^(.+\.conf):([\w.]+)$/m, [ [:target], [:key] ] ],
+    # TODO: make target optional
+    #[ /^([\w.]+)$/m, [ [:key] ] ],
```
ekohl, line 12: I didn't have time to figure out how `title_patterns` works with matching, but it probably isn't hard.

bastelfreak: that really depends on the point of view :D

ekohl: It comes down to understanding https://github.com/puppetlabs/puppet/blob/945beb6ef0200b1bc7879b62131282a61fc9a935/lib/puppet/resource.rb#L625-L663 and finding the right combination.

ekohl: Ok, couldn't help myself. I think the commented code should work. The reason it's commented is that I didn't know if it would really work. But this part suggests the patterns are correct:
```ruby
require 'test/unit'

TITLE_PATTERNS = [ 
  [ /^(.+\.conf):([\w.]+)$/m, [ [:target], [:key] ] ],
  [ /^([\w.]+)$/m, [ [:key] ] ],
]

def match(title)
  h = {}
  TITLE_PATTERNS.each do |regexp, symbols_and_lambdas|
    captures = regexp.match(title.to_s)
    if captures
      symbols_and_lambdas.zip(captures[1..-1]).each do |symbol_and_lambda,capture|
        symbol, proc = symbol_and_lambda
        # Many types pass "identity" as the proc; we might as well give
        # them a shortcut to delivering that without the extra cost.
        # 
        # Especially because the global type defines title_patterns and
        # uses the identity patterns.
        # 
        # This was worth about 8MB of memory allocation saved in my
        # testing, so is worth the complexity for the API.
        if proc then
          h[symbol] = proc.call(capture)
        else
          h[symbol] = capture
        end
      end
      return h 
    end
  end
end

class MatcherTest < Test::Unit::TestCase
  def test_with_conf
    assert_equal({target: '/postgresql.conf', key: 'foo'}, match('/postgresql.conf:foo'))
  end

  def test_without_conf
    assert_equal({key: 'foo'}, match('foo'))
  end
end
```

```console
$ ruby title_patterns.rb 
Loaded suite title_patterns
Started
Finished in 0.000252056 seconds.
--------------------------------------------------------------------------------------
2 tests, 2 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
--------------------------------------------------------------------------------------
7934.74 tests/s, 7934.74 assertions/s
```

jst-cyr:
@ekohl : Do you think this is still an issue with current postgresql module releases? I'm just wondering if it's okay to close this older draft PR off or if there is still a need to figure out a way to solve this.

ekohl:
I think it's still an issue that it reinvents the `ParsedFile` but given it needs a bugfix in Puppet itself which is now closed source I can't contribute that. And I have no intention to contribute to closed source software.

PR #1544, opened 2023-10-25
postgresql_conf: Fix regex for value param and add tests

SimonHoenscheid:
## Summary
Fixes #1536

PR #1545, opened 2023-10-25
postgresql_conf: add spec tests for value param

bastelfreak:
This cherry-picks the tests from #1544 so we can see how the currently fail on main.


SimonHoenscheid:
@bastelfreak this one is obsolete right?

PR #1557, opened 2023-11-23
(#1556) Fix Python package name for Ubuntu >= 22.04

### LINKED ISSUE #1556: Tries to install the wrong Python PostgreSQL package on Ubuntu 22.04
## Describe the Bug

Module tries to install package `python-psycopg2` under Ubuntu 22.04 where that package no longer exists. It is available as `python3-psycopg2` instead.

## Expected Behavior

Module should distinguish between Ubuntu versions >= 22.04 and < 22.04 and install the correctly named Python PostgreSQL package.

## Steps to Reproduce

1. `include postgresql::lib::python` on Ubuntu 22.04
2.  Watch Puppet agent run fail: 

        E: Unable to locate package python-psycopg2
        Error: /Stage[main]/Postgresql::Lib::Python/Package[python-psycopg2]/ensure: change from 'purged' to 'present' failed:   Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold install python-psycopg2' returned 100: Reading package lists...
        Building dependency tree...
        Reading state information...
        E: Unable to locate package python-psycopg2 (corrective)

## Environment
 - Module version 9.2.0 and later
 - Ubuntu 22.04

antaflos:
## Summary

For Ubuntu 22.04 and later the Python PostgreSQL package is called "python3-psycopg2" so make that distinction in params.pp.

## Additional Context

This is missing spec tests but apparently there are no examples for Ubuntu past 18.04 so not sure where I would add tests to cover this change.

## Related Issues (if any)

Fixes #1556 

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -159,7 +159,15 @@ class postgresql::params inherits postgresql::globals {
       $perl_package_name      = pick($perl_package_name, 'libdbd-pg-perl')
       $plperl_package_name    = pick($plperl_package_name, "postgresql-plperl-${version}")
       $plpython_package_name  = pick($plpython_package_name, "postgresql-plpython-${version}")
-      $python_package_name    = pick($python_package_name, 'python-psycopg2')
+
+      $_ubuntu_2204 = ($facts['os']['name'] == 'Ubuntu' and versioncmp($facts['os']['release']['full'], '22.04') >= 0)
+      $_debian_12 = ($facts['os']['name'] == 'Debian' and versioncmp($facts['os']['release']['full'], '12') >= 0)
+
+      if $_ubuntu_2204 or $_debian_12 {
+        $python_package_name = pick($python_package_name, 'python3-psycopg2')
+      } else {
+        $python_package_name = pick($python_package_name, 'python-psycopg2')
+      }
 
       $bindir                 = pick($bindir, "/usr/lib/postgresql/${version}/bin")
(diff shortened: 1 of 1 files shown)


manifests/params.pp
```diff
@@ -159,7 +159,12 @@
       $perl_package_name      = pick($perl_package_name, 'libdbd-pg-perl')
       $plperl_package_name    = pick($plperl_package_name, "postgresql-plperl-${version}")
       $plpython_package_name  = pick($plpython_package_name, "postgresql-plpython-${version}")
-      $python_package_name    = pick($python_package_name, 'python-psycopg2')
+
+      if $facts['os']['name'] == 'Ubuntu' and versioncmp($facts['os']['release']['full'], '22.04') >= 0 {
```
ekohl, line 163: Looks like the same is needed at least for Debian 12. For Debian 11 there still is `python-psycopg2` as well as `python3-psycopg2`. Would you mind doing the same for Debian >= 12 at least while you're at it?

ekohl (approved):
In itself this is correct.

manifests/params.pp
antaflos: I have added such a test for Debian 12 as well, trying to make the logic somewhat readable. Not sure if this is in tune with the module's coding style though.

ekohl (approved):

ekohl:
I don't see how the failing checks are related so I'm merging this

PR #1561, opened 2023-12-19
support for a custom apt source release

h0tw1r3:
Postgresql no longer specifically supports Ubuntu 18.04, but still supports Debian 10 which can be used until Puppet drops support for Ubuntu 18.04.

## Summary
Postgresql recommends using the Debian buster repo for Ubuntu bionic.

## Additional Context
Temporary until Puppet no longer supports Ubuntu 18.04.

## Related Issues (if any)
https://github.com/voxpupuli/puppet-puppetboard/pull/402

## Checklist
- 🟢 Spec tests.
- 🟢 Acceptance tests.

### DIFF
--- a/data/os/Ubuntu/18.04.yaml
+++ b/data/os/Ubuntu/18.04.yaml
@@ -1,0 +1,3 @@
+---
+postgresql::repo::baseurl: https://apt-archive.postgresql.org/pub/repos/apt/
+postgresql::repo::release: "%{facts.os.distro.codename}-pgdg-archive"
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -60,6 +60,7 @@
 #
 # @param repo_baseurl Sets the baseurl for the PostgreSQL repository. Useful if you host your own mirror of the repository.
 # @param yum_repo_commonurl Sets the url for the PostgreSQL common Yum repository. Useful if you host your own mirror of the YUM repository.
+# @param apt_source_release Overrides the default release for the apt source.
 #
 # @param needs_initdb
 #   Explicitly calls the initdb operation after the server package is installed and before the PostgreSQL service is started.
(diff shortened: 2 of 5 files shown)


ekohl (dismissed):
IMHO we shouldn't do this and instead drop 18.04. The standard support of 18.04 has ended and clearly PostgreSQL no longer wants to supports it since they dropped it from their repositories (on 2023-11-05 according to https://wiki.postgresql.org/wiki/Apt#News).

If anything, we can introduce a parameter to `apt_postgresql_org` for the release so you can override it via Hiera. In fact, I think that'd be wise for everything including the GPG keys.

h0tw1r3:
> If anything, we can introduce a parameter to `apt_postgresql_org` for the release so you can override it via Hiera. In fact, I think that'd be wise for everything including the GPG keys.

Reworked. Thanks for the constructive feedback. I think the overall design pattern could use attention which is why I'm _not_ touching key management _yet_.

data/os/Ubuntu/18.04.yaml
```diff
@@ -0,0 +1,3 @@
+---
+postgresql::repo::baseurl: https://apt-archive.postgresql.org/pub/repos/apt/
+postgresql::repo::release: "%{facts.os.distro.codename}-pgdg-archive"
```
ekohl, line 3: Having archive repos is much cleaner than using a different distro. I much prefer this.

ekohl (commented):
I can't properly review this right now, but at first glance this looks like it addresses my concerns.

ekohl (approved):
> I think the overall design pattern could use attention which is why I'm _not_ touching key management _yet_.

I think https://github.com/puppetlabs/puppetlabs-postgresql/pull/1563 would be a major step in that direction. If you just need to provide a URL without an ID then it's way easier.

PR #1547, opened 2023-10-26
reenable acceptance test for server encoding

bastelfreak:
## Summary
Provide a detailed description of all the changes present in this pull request.

## Additional Context
Add any additional context about the problem here. 
- [ ] Root cause and the steps to reproduce. (If applicable)
- [ ] Thought process behind the implementation.

## Related Issues (if any)
Mention any related issues or pull requests.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.
- [ ] Manually verified. (For example `puppet apply`)

Issue #1541, opened 2023-10-20
postgresql::server::config_entry - dependency cycles detected (after upgrade to 10.0.0)


augustocarvalho:
## Describe the Bug
After upgrade from 9.2.0 to 10.0.0 or 10.0.1 puppet fails with dependency cycle error:
`Error: Found 1 dependency cycle:
(Anchor[postgresql::server::service::begin::main] => Postgresql_psql[Set template1 encoding to UTF8] => Postgresql::Server::Instance::Late_initdb[main] => Postgresql::Server::Instance::Initdb[main] => Class[Postgresql::Server::Initdb] => Postgresql_conf[wal_level] => Class[Postgresql::Server::Service] => Postgresql::Server::Instance::Service[main] => Anchor[postgresql::server::service::begin::main])\nCycle graph written to /opt/puppetlabs/puppet/cache/state/graphs/cycles.dot.
Error: Failed to apply catalog: One or more resource dependency cycles detected in graph
`

## Expected Behavior
Update the config entry to postgresql.conf

## Steps to Reproduce
# For example change the wal_level
  postgresql::server::config_entry { 'wal_level':
    value   => 'archive',    
  }

Apply puppet and it fails.

## Environment
 - Version [Puppet 7]
 - Platform [Ubuntu 20.04]
 - Postgresql-14
 - puppetlab-postgresql (since 10.0.0)

bastelfreak:
Hi,
thanks for the Ticket, we're able to reproduce it in https://github.com/puppetlabs/puppetlabs-postgresql/pull/1547

ShadNex:
Any updates on this? We're experiencing the same issue after upgrading the module and had to revert to 9.2.0

bastelfreak:
@ShadNex this should be fixed in https://github.com/puppetlabs/puppetlabs-postgresql/pull/1564 . Can you try that branch and comment in the PR please?

PR #1564, opened 2024-01-03
Unbreak Debian support with custom encoding

smortex:
On Debian, version 10.0.0+ of the module cause a broken catalog due to dependecy cycles:

```
Anchor[postgresql::server::service::begin::main] =>
Postgresql_psql[Set template1 encoding to UTF8] =>
Postgresql::Server::Instance::Late_initdb[main] =>
Postgresql::Server::Instance::Initdb[main] =>
Class[Postgresql::Server::Initdb] =>
Postgresql_conf[port_for_instance_main] =>
Class[Postgresql::Server::Service] =>
Postgresql::Server::Instance::Service[main] =>
Anchor[postgresql::server::service::begin::main]
```

Reenable the test that trigger this issue (taken from #1547) and try to fix it.


antaflos:
We are successfully running this branch and it works as advertised. No more dependency cycles. :+1:

smortex:
@antaflos Beware that this is unfortunately still WIP because it breaks EL for reasons I had not time to discover yet.  I have been busy the last week with other $WORK related things, and am about to go on vacation for a week, so will not be able to have a look before I am back.

But thank your for your feedback :+1: : for now I only tested it with the CI and as you can see with [919fc9b](https://github.com/puppetlabs/puppetlabs-postgresql/pull/1564/commits/919fc9b0cefb2a1d75550e514652cfa3310c7613) there is some nasty bug because A -> B + A -> B -> C and A -> B + B -> C should not change much regarding dependencies.

lib/puppet/type/postgresql_psql.rb
```diff
@@ -135,11 +135,7 @@ def matches(value)
   end
 
   autorequire(:anchor) do
```
bastelfreak, line 137: as an alternative, can we maybe autorequire the postgresql_conn_validator resource?

spec/acceptance/utf8_encoding_spec.rb
```diff
@@ -15,8 +16,15 @@ class { 'postgresql::server': }
...
-    expect(port(5432)).to be_listening
+    idempotent_apply(pp, debug: true)
+    puts '-------------------------------'
+    puts LitmusHelper.instance.run_shell('ss -lntp').stdout
+    puts '-------------------------------'
+    puts LitmusHelper.instance.run_shell('journalctl -u postgresql').stdout
+    puts '-------------------------------'
+    puts LitmusHelper.instance.run_shell('systemctl status postgresql*').stdout
+    puts '-------------------------------'
+    expect(port(5432)).to be_listening.on('127.0.0.1').with('tcp')
```
bastelfreak, line 27: the problem I had in the past: some systems report tcp, some tcp4. that was really ugly so I switched to matching the IP address. and that sometimes didn't work because some CI systems have IPv6 so it binds to ::1 but sometimes it's only 127.0.0.1. all ugly...

lib/puppet/type/postgresql_psql.rb
```diff
@@ -135,11 +135,7 @@ def matches(value)
...
-    ["postgresql::server::service::begin::#{self[:instance]}"]
-  end
-
-  autorequire(:service) do
```
ekohl, line 141: I'm trying to understand the logic. Don't you **always** need the server to be running to execute any SQL statements?

ekohl (commented):
Looking at this, I think the `initdb` (which doesn't need a DB running) shouldn't include `late_initdb` (which does need a server running). Instead, `postgresql::server_instance` should be modified to split it.

The reason you run into this is that Debian has `$needs_initdb` set to false by default:

https://github.com/puppetlabs/puppetlabs-postgresql/blob/5e6e99484d8ac6b5386fd63d15e4cbd50de9dcf8/manifests/params.pp#L145

So this code path is triggered:
https://github.com/puppetlabs/puppetlabs-postgresql/blob/5e6e99484d8ac6b5386fd63d15e4cbd50de9dcf8/manifests/server/instance/initdb.pp#L195-L202

Wouldn't it make more sense to move it to `postgresql::server_instance` as:
```puppet
if !$initdb_settings['needs_initdb'] and $initdb_settings['encoding'] {
  postgresql::server::instance::late_initdb { $instance_name:
    encoding       => $initdb_settings['encoding'],
    user           => $initdb_settings['user'],
    group          => $initdb_settings['group'],
    module_workdir => $initdb_settings['module_workdir'],
    psql_path      => $initdb_settings['psql_path'],
    port           => $initdb_settings['port'],
  }
}
```

It may break because those values may not be `undef`, but I hope the idea is clear.

lib/puppet/type/postgresql_psql.rb
smortex: That quite make sense.  I'll dig into this.

spec/acceptance/utf8_encoding_spec.rb
smortex: I am doing more test because this is quite weird.  Looking at the code it *should* work, looking at `ss` output it *seems* to be working, but it is not and that makes me feel angry.  Similar code work in other areas of the project so I guess something is very wrong…  I would like to nail it rather than ignoring it if possible.

lib/puppet/type/postgresql_psql.rb
ekohl: Oh yes, that makes sense. Though with the newest versions (Fedora 38 with PostgreSQL 15) I see the systemd unit is `Type=notify` so I'd assume that there it will only really be ready once it's listening (spoiler: the patch sends `READY=1` right after logging `database system is ready to accept connections`).

Digging into this, it was introduced in https://github.com/postgres/postgres/commit/7d17e683fcc28a1b371c7dd02935728cd2cbf9bf so with PostgreSQL 9.6 it became possible to support. Looking at EL8 that's built with `Type=notify`, but EL7 probably isn't. Sadly Debian 11 & 12 are also `Type=forking`. That means you're right and we still need to depend on the connection validator.

kenyon: > Oh yes, that makes sense. Though with the newest versions (Fedora 38 with PostgreSQL 15) I see the systemd unit is `Type=notify` so I'd assume that there it will only really be ready once it's listening (spoiler: the patch sends `READY=1` right after logging `database system is ready to accept connections`).
> 
> Digging into this, it was introduced in [postgres/postgres@7d17e68](https://github.com/postgres/postgres/commit/7d17e683fcc28a1b371c7dd02935728cd2cbf9bf) so with PostgreSQL 9.6 it became possible to support. Looking at EL8 that's built with `Type=notify`, but EL7 probably isn't. Sadly Debian 11 & 12 are also `Type=forking`. That means you're right and we still need to depend on the connection validator.

These seem to be the relevant bugs about this:

* https://bugs.launchpad.net/ubuntu/+source/postgresql-common/+bug/1862138
* https://bugs.debian.org/872915

ShadNex:
I've tested this PR as suggested by @bastelfreak and indeed it seems to solve the dependency issue for our environment as seen in #1541 👍 Would be great if this gets released soon!

manifests/server/instance/service.pp
```diff
@@ -42,6 +42,10 @@
       status    => $service_status,
     }
 
+    Anchor["postgresql::server::service::begin::${name}"]
```
davidsandilands, line 45: I may deeply regret asking when I do not have a full background on this but my understanding was Anchor was a redundant pattern as the contain function came in?

ShadNex:
Is there someting missing for this PR to get merged? Would be nice to have it in the next relase.

bastelfreak:
@ShadNex yes, passing tests :)

@smortex can you rebase again please?

manifests/server/instance/service.pp
smortex: Yes anchors are from the past.  Not sure if attempting to remove them in this PR will help however (and that would be a backwards incompatible change).  But in the long term, having them gone is probably a good idea.

smortex:
> Please bear with me while my fat fingers type crap on the keyboard 😒 .

Thank you for your patience.

Rebasing done, I pushed each commit individually to have a step by step view of what works or not.  I obviously lost track since January, and was (and still am) out of ideas so if you want to try something you are very welcome to do so.  It seems that failures are test-order dependent…

span786:
Failures seems unreleted to me, we can ge this merged.

amitkarsale:
Merged based on review comments.

PR #1577, opened 2024-02-21
don't set data_directory in config when we configure it via env

### LINKED ISSUE #1576: don't set data_directory in the config by default
## Describe the Bug

Today, we always set `data_directory = …` in `postgresql.conf` via
https://github.com/puppetlabs/puppetlabs-postgresql/blob/411e7bce3f2d5256c315351144dc525b6c0afbbe/manifests/server/instance/config.pp#L222-L225

However we also set it in the systemd override:
https://github.com/puppetlabs/puppetlabs-postgresql/blob/411e7bce3f2d5256c315351144dc525b6c0afbbe/manifests/server/instance/systemd.pp#L29-L34
https://github.com/puppetlabs/puppetlabs-postgresql/blob/411e7bce3f2d5256c315351144dc525b6c0afbbe/templates/systemd-override.conf.epp#L11

And pass it when calling `initdb`:
https://github.com/puppetlabs/puppetlabs-postgresql/blob/411e7bce3f2d5256c315351144dc525b6c0afbbe/manifests/server/instance/initdb.pp#L153

The problem is that setting `data_directory` in the config actually overrides the `-D` parameter on the CLI and the `PGDATA` environment variable, as can be seen in e.g. https://bugzilla.redhat.com/show_bug.cgi?id=1935301 thus possibly breaking future actions against this DB (see the link for an example).

## Expected Behavior
`data_directory` not set in `postgresql.conf` when systemd is used and thus the environment variable is present

## Steps to Reproduce
install a postgresql setup :)

## Environment
 - Version 10.0.3
 - Platform CentOS Stream 8

## Additional Context
This is the exact opposite of #510 :woman_shrugging:

evgeni:
Fixes: #1576

## Summary
see subject and linked issue

## Additional Context

## Related Issues (if any)
#1576

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/manifests/server/instance/config.pp
+++ b/manifests/server/instance/config.pp
@@ -219,9 +219,11 @@ define postgresql::server::instance::config (
     }
   }
 
-  postgresql::server::config_entry { "data_directory_for_instance_${name}":
-    key   => 'data_directory',
-    value => $datadir,
+  unless $facts['service_provider'] == 'systemd' and $facts['os']['family'] in ['RedHat', 'Gentoo'] {
+    postgresql::server::config_entry { "data_directory_for_instance_${name}":
+      key   => 'data_directory',
+      value => $datadir,
+    }
   }
   if $timezone {
     postgresql::server::config_entry { "timezone_for_instance_${name}":
--- a/spec/defines/server_instance_spec.rb
+++ b/spec/defines/server_instance_spec.rb
@@ -123,7 +123,6 @@ describe 'postgresql::server_instance' do
     it { is_expected.to contain_postgresql_conn_validator('validate_service_is_running_instance_test1') }
     it { is_expected.to contain_postgresql_conf('port_for_instance_test1') }
     it { is_expected.to contain_postgresql_conf('log_statement_stats_test1') }
-    it { is_expected.to contain_postgresql_conf('data_directory_for_instance_test1') }
     it { is_expected.to contain_postgresql_conf('autovacuum_vacuum_scale_factor_test1') }
     it { is_expected.to contain_postgresql_conf('authentication_timeout_test1') }
     it { is_expected.to contain_postgresql__server__role('app_test1') }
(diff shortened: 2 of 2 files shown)


ekohl:
I'm not sure the systemd check is relevant. Debian also starts it via systemd, but somehow does read `data_directory` via `postgresql.conf`.

Debian uses `/usr/bin/pg_ctlcluster` to start, which is a perl script. That sets `-D $info{'pgdata'}`. My perl isn't great, but I think it reads `postgresql.conf` to get the value.

The official postgresql.org packages on Red Hat match what the distro packages do:
```ini
Environment=PGDATA=/var/lib/pgsql/13/data/
ExecStart=/usr/pgsql-13/bin/postmaster -D ${PGDATA}
```

evgeni:
> I'm not sure the systemd check is relevant. Debian also starts it via systemd, but somehow does read `data_directory` via `postgresql.conf`.

Well, this is where *we* set PGDATA explicitly. So it was a safe start.

> Debian uses `/usr/bin/pg_ctlcluster` to start, which is a perl script. That sets `-D $info{'pgdata'}`. My perl isn't great, but I think it reads `postgresql.conf` to get the value.

`-D` really means "find postgresql.conf in that directory and run with it", which can point at a `data_directory` somewhere *completely different* (see https://www.postgresql.org/message-id/3566642.1618422939%40sss.pgh.pa.us)

> The official postgresql.org packages on Red Hat match what the distro packages do:
> 
> ```ini
> Environment=PGDATA=/var/lib/pgsql/13/data/
> ExecStart=/usr/pgsql-13/bin/postmaster -D ${PGDATA}
> ```

"good"

manifests/server/instance/config.pp
```diff
@@ -219,9 +219,11 @@
     }
   }
 
-  postgresql::server::config_entry { "data_directory_for_instance_${name}":
-    key   => 'data_directory',
-    value => $datadir,
+  unless $facts['service_provider'] == 'systemd' and $facts['os']['family'] in ['RedHat', 'Gentoo'] {
+    postgresql::server::config_entry { "data_directory_for_instance_${name}":
```
ekohl, line 223: For upgrades, should we ensure it's absent on systemd + RH/Gentoo?

evgeni: Probably yes?

jst-cyr:
@evgeni : This PR seems to have been open for some time in the draft mode with some unresolved feedback and is out of date now with the main branch. Are you still needing this change? Or can we close this out?

evgeni:
This is still an issue in the PostgreSQL module, but I am not interested in contributing to Perforce modules anymore.

jst-cyr:
That's fair @evgeni , thanks for the response! I'll close this one out for now, then, and hopefully we can address the issue in a new PR in the future.

PR #1584, opened 2024-03-26
Add a global password_encryption parameter

ekohl:
## Summary

The function postgresql::password() looks at the default for encryption. If you want to use SCRAM on PostgreSQL < 14 then the default must be overridden, or specified for every use. This parameter allows it to be globally overridden.

## Related Issues (if any)

Found in https://github.com/theforeman/foreman-installer/pull/924 when testing out SCRAM support with PostgreSQL 13.

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -25,7 +25,7 @@ class postgresql::params inherits postgresql::globals {
   $manage_selinux               = pick($manage_selinux, false)
   $package_ensure               = 'present'
   $module_workdir               = pick($module_workdir,'/tmp')
-  $password_encryption          = versioncmp($version, '14') ? { -1 => 'md5', default => 'scram-sha-256' }
+  $password_encryption          = pick($password_encryption, versioncmp($version, '14') ? { -1 => 'md5', default => 'scram-sha-256' })
   $extra_systemd_config         = undef
   $manage_datadir               = true
   $manage_logdir                = true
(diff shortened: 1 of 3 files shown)


manifests/params.pp
```diff
@@ -25,7 +25,7 @@
   $manage_selinux               = pick($manage_selinux, false)
   $package_ensure               = 'present'
   $module_workdir               = pick($module_workdir,'/tmp')
-  $password_encryption          = versioncmp($version, '14') ? { -1 => 'md5', default => 'scram-sha-256' }
+  $password_encryption          = pick($password_encryption, versioncmp($version, '14') ? { -1 => 'md5', default => 'scram-sha-256' })
```
bastelfreak, line 28: Does that work? `pick()` refrences the variable itself? And line 21-24 follow the same pattern :thinking: 

should it be:
```suggestion
  $password_encryption          = pick($postgresql::globals::password_encryption, versioncmp($version, '14') ? { -1 => 'md5', default => 'scram-sha-256' })
```

ekohl: Yes, I was also surprised. I think it's (ab)using the parser a bit. I wanted to remain consistent.

bastelfreak: okay. so it's probably coming from globals. interesting. Not sure if that works on purpose or by accident :D

ekohl: "yes"

PR #1585, opened 2024-03-27
params.pp: Cleanup service_reload command; use rcctl on BSD

bastelfreak:
Every BSD supports (and uses by default) rcctl. The rest uses systemd.

## Summary

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -85,7 +85,6 @@ class postgresql::params inherits postgresql::globals {
         $postgresql_conf_mode   = pick($postgresql_conf_mode, '0600')
       }
 
-      $service_reload = "systemctl reload ${service_name}"
       $service_status = pick($service_status, "systemctl status ${service_name}")
 
       $psql_path           = pick($psql_path, "${bindir}/psql")
@@ -131,7 +130,6 @@ class postgresql::params inherits postgresql::globals {
       $psql_path              = pick($psql_path, "${bindir}/psql")
 
       $service_status         = pick($service_status, "systemctl status ${service_name}")
-      $service_reload         = "systemctl reload ${service_name}"
       $python_package_name    = pick($python_package_name, 'python-psycopg2')
       # Archlinux does not have a perl::DBD::Pg package
       $perl_package_name      = pick($perl_package_name, 'undef')
@@ -172,7 +170,6 @@ class postgresql::params inherits postgresql::globals {
       $bindir                 = pick($bindir, "/usr/lib/postgresql/${version}/bin")
       $datadir                = pick($datadir, "/var/lib/postgresql/${version}/main")
       $confdir                = pick($confdir, "/etc/postgresql/${version}/main")
-      $service_reload         = "systemctl reload ${service_name}"
       $service_status         = pick($service_status, "systemctl status ${service_name}")
       $psql_path              = pick($psql_path, '/usr/bin/psql')
       $postgresql_conf_mode   = pick($postgresql_conf_mode, '0644')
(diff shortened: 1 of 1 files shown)


manifests/params.pp
```diff
@@ -304,4 +297,10 @@
   $postgresql_conf_path = pick($postgresql_conf_path, "${confdir}/postgresql.conf")
   $recovery_conf_path   = pick($recovery_conf_path, "${datadir}/recovery.conf")
   $default_database     = pick($default_database, 'postgres')
+
+  $service_reload = if $facts['service_provider'] == 'systemd' {
+    "systemctl reload ${service_name}"
+  } else {
+    "rcctl ${service_name}"
```
ekohl, line 304: Does `rcctl` reload? The name implies it's similar to `systemctl` in that it controls `rc`.

manifests/params.pp
```diff
@@ -304,4 +297,10 @@
   $postgresql_conf_path = pick($postgresql_conf_path, "${confdir}/postgresql.conf")
   $recovery_conf_path   = pick($recovery_conf_path, "${datadir}/recovery.conf")
   $default_database     = pick($default_database, 'postgres')
+
+  $service_reload = if $facts['service_provider'] == 'systemd' {
+    "systemctl reload ${service_name}"
+  } else {
```
ekohl, line 303: Would it make sense to still guard this under `$facts['kernel']` or similar to catch BSDs?

ekohl (commented):
I wonder if there is  way to somehow implement reloading natively in Puppet. This particular case is a bit hard in that it sometimes needs to reload, sometimes restart. If there's a `reload => true` option you can use datacat-style options to change it to `false` in cases where a restart is needed.

Of course, before you can rely on that in Puppet it'll be a while. That shouldn't hold us back from investigating the solution though.

manifests/params.pp
bastelfreak: I was still on it, there was a `reload` missing

bastelfreak:
> I wonder if there is way to somehow implement reloading natively in Puppet.

I would like to see this since years.

A notify to a service always triggers the restart command. Now we can set the restart command to `systemctl reload`, but then we cannot do a restart anymore. I think an ideal solution would be to always use `systemctl reload-or-restart` and systemd/the service need to figure out what changed and if a reload/restart is required. But that's also tricky :(

manifests/params.pp
bastelfreak: I can, but what should we do if it's neither systemd nor BSD? We currently don't have a case for this.

bastelfreak: I wanted to check FacterDB for Factsets from netBSD/OpenBSD, because I'm not sure what they report as `kernel` fact, but we don't have any Factsets for those :(

smortex:
> Every BSD supports (and uses by default) rcctl. The rest uses systemd.

Huh… I have some bad news…

```
root@zappy ~ # rcctl
zsh: command not found: rcctl
root@zappy ~ # uname -a
FreeBSD zappy.blogreen.org 14.0-RELEASE-p3 FreeBSD 14.0-RELEASE-p3 #0: Mon Dec 11 04:56:01 UTC 2023     root@amd64-builder.daemonology.net:/usr/obj/usr/src/amd64.amd64/sys/GENERIC amd64
```

According to https://man.openbsd.org/rcctl:

> HISTORY
> rcctl first appeared in OpenBSD 5.7.

So I would expect that *only* OpenBSD has rcctl(8) :disappointed:.

bastelfreak:
@smortex ah derp. I know it's available on FreeBSD and I have to maintain some boxes, I assumed it's available by default :(

bastelfreak:
Ahhhhhhhh freebsd.org fooled me! I also checked https://man.freebsd.org/cgi/man.cgi?query=rcctl&sektion=8&apropos=0&manpath=OpenBSD+6.5 yesterday, but I didn't notice the drop down at the top where OpenBSD is selected. and it's not available on FreeBSD :(

SimonHoenscheid (approved):

PR #1591, opened 2024-04-10
Add EL9 support

bastelfreak:
## Summary

## Additional Context

## Related Issues (if any)

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/spec/acceptance/server_instance_spec.rb
+++ b/spec/acceptance/server_instance_spec.rb
@@ -3,7 +3,7 @@
 # run a test task
 require 'spec_helper_acceptance'
 
-describe 'postgresql instance test1', if: os[:family] == 'redhat' && os[:release].start_with?('8') do
+describe 'postgresql instance test1', if: os[:family] == 'redhat' && !os[:release].start_with?('7') do
   pp = <<-MANIFEST
   # set global defaults
   class { 'postgresql::globals':
@@ -12,7 +12,7 @@ describe 'postgresql instance test1', if: os[:family] == 'redhat' && os[:release
     manage_package_repo => false,
     manage_dnf_module   => true,
     needs_initdb        => true,
-    version             => '13',
+    version             => '16',
   }
   # stop default main instance
   class { 'postgresql::server':
--- a/metadata.json
+++ b/metadata.json
@@ -38,13 +38,15 @@
       "operatingsystem": "CentOS",
       "operatingsystemrelease": [
         "7",
-        "8"
+        "8",
+        "9"
       ]
     },
     {
       "operatingsystem": "OracleLinux",
       "operatingsystemrelease": [
-        "7"
+        "7",
+        "9"
       ]
     },
     {
@@ -79,13 +81,15 @@
     {
       "operatingsystem": "Rocky",
       "operatingsystemrelease": [
-        "8"
+        "8",
+        "9"
       ]
     },
     {
       "operatingsystem": "AlmaLinux",
       "operatingsystemrelease": [
-        "8"
+        "8",
+        "9"
       ]
     }
   ],


spec/acceptance/server_instance_spec.rb
```diff
@@ -3,7 +3,7 @@
 # run a test task
 require 'spec_helper_acceptance'
 
-describe 'postgresql instance test1', if: os[:family] == 'redhat' && os[:release].start_with?('8') do
+describe 'postgresql instance test1', if: os[:family] == 'redhat' && os[:release].split.first.to_i >= 8 do
```
ekohl, line 6: I'd just exclude EL7. IMHO easier to read.
```suggestion
describe 'postgresql instance test1', if: os[:family] == 'redhat' && !os[:release].start_with?('7') do
```

ekohl (approved):

SimonHoenscheid:
EL9 manages the default postgres in dnf modules, the postgres repo does not. 

Recommended way is to disable the postgres dnf modules and install from the postgres repo. 

the postgres repo requires a version which is defined via postgresql::globals::version, this is also the default version which is handed to the dnf module, ehich would need disabled here, to act correcty, the globals class requires a patch to make this possible.

My suggestion is to add an optional_parameter like dnf_module_version, unless set real_dnf_module_version defaults to $version/global_version.

https://github.com/puppetlabs/puppetlabs-postgresql/blob/main/manifests/globals.pp#L289

spec/acceptance/server_instance_spec.rb
```diff
@@ -3,7 +3,7 @@
 # run a test task
 require 'spec_helper_acceptance'
 
-describe 'postgresql instance test1', if: os[:family] == 'redhat' && os[:release].start_with?('8') do
+describe 'postgresql instance test1', if: os[:family] == 'redhat' && !os[:release].start_with?('7') do
   pp = <<-MANIFEST
   # set global defaults
   class { 'postgresql::globals':
```
ekohl, line 9: I'd change the version to 15 or 16. I don't have a RHEL 8 at hand to see if it ships 16 already, so I'd suggest 15.

ekohl (changes requested):
> EL9 manages the default postgres in dnf modules, the postgres repo does not.

For plain EL9 this isn't true:
```console
# dnf install postgresql
...
# dnf module list --enabled -q
```

You can install PostgreSQL 15 or 16 via modules, but the default (13) is not modular.

The reason the test fails is that it sets `manage_dnf_module` to `true` and `version` to `'13'`. That isn't a combination that exists.

The easiest fix is to pick a version that exists in both EL 8 & 9 as modular versions, which means picking 15 or 16.

spec/acceptance/server_instance_spec.rb
vchepkov:
```
Red Hat Enterprise Linux 8 for x86_64 - AppStream (RPMs)
Name                                                                     Stream                                                               Profiles                                                                        Summary                                                                                         
postgresql                                                               9.6                                                                  client, server [d]                                                              PostgreSQL server and client module                                                             
postgresql                                                               10 [d]                                                               client, server [d]                                                              PostgreSQL server and client module                                                             
postgresql                                                               12                                                                   client, server [d]                                                              PostgreSQL server and client module                                                             
postgresql                                                               13                                                                   client, server [d]                                                              PostgreSQL server and client module                                                             
postgresql                                                               15                                                                   client, server [d]                                                              PostgreSQL server and client module                                                             
postgresql                                                               16                                                                   client, server [d]                                                              PostgreSQL server and client module                                             
```

ekohl: @bastelfreak could you change line 15 from `'13'` to `'16'`? I'd then expect it to pass.

bastelfreak: done, let's see.

ekohl (approved):
Thanks

PR #1599, opened 2024-05-30
Avoid opening the file in postgresql_conf

ekohl:
## Summary

Using `File.open` without closing it can leak file descriptors. It's actually not needed at all because `File.foreach`, `File.readlines` and `File.write` all accept a filename. This simplifies the code in the process.

Then it simplifies the code by passing an offset to `with_index` to calculate line numbers. Rather than starting from 0 and always adding 1, this tells `with_index` to start at 1.

It also fixes the spelling of PostgreSQL in a debug statement.

## Checklist

At this point I haven't verified this yet, but submitting it so I can more easily test it on some system.

- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/lib/puppet/provider/postgresql_conf/ruby.rb
+++ b/lib/puppet/provider/postgresql_conf/ruby.rb
@@ -36,7 +36,7 @@ Puppet::Type.type(:postgresql_conf).provide(:ruby) do
         active_settings.push(attributes_hash)
       end
     end
-    Puppet.debug("DEBUG: parse_config Active Settings found in Postgreql config file: #{active_settings}")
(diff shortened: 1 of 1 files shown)


bastelfreak (approved):
I think this is fine to merge, we've acceptance tests that cover this

SimonHoenscheid (approved):

deric (approved):

bastelfreak:
did a bit more testing locally to verify this works

bastelfreak (approved):

Ramesh7 (approved):

Issue #1602, opened 2024-07-05
postgresql::server::config_entry should allow the empty string for `value` parameter


antaflos:
## Describe the Bug

PostgreSQL supports and allows config entries, such as those in `postgresql.conf`, to be set to the empty string. The `postgresql::server::config_entry` defined type, however, requires `String[1]` when supplying string values. This doesn't allow for the empty string.

## Expected Behavior

The `postgresql::server::config_entry` defined type should allow the empty string for its `value` parameter.

## Steps to Reproduce

A typical use case is setting database-specific parameters that can be empty:

```puppet
postgresql::server::config_entry { 'my_database.important_parameter':
  value => ''
}
```

This should render as follows in `postgresql.conf`, which is valid from PostgreSQL's PoV.

```
my_database.important_parameter = ''
```

However, the Puppet agent run fails with an error message:

```
Postgresql::Server::Config_entry[my_database.important_parameter]: parameter 'value' expects a value of type Undef, String[1], Numeric, or Array[String[1]], got String
```

This is because the empty string is not allowed here.

## Environment
 - Version 10.3.0
 - Platform Ubuntu 20.04, 22.04

PR #1603, opened 2024-07-05
Allow empty string value for config entries

antaflos:
## Summary

PostgreSQL supports and allows config entries, such as those in postgresql.conf, to be set to the empty string. The `postgresql::server::config_entry` defined type, however, requires String[1] when supplying string values. This doesn't allow for the empty string.

This change relaxes the allowed data types for the `value` parameter of `postgresql::server::config_entry` to `String` from `String[1]` and adds a spec test to support the change.

## Related Issues (if any)

#1602 

## Checklist
- [x] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.

### DIFF
--- a/manifests/server/config_entry.pp
+++ b/manifests/server/config_entry.pp
@@ -10,7 +10,7 @@
 define postgresql::server::config_entry (
   Enum['present', 'absent']                               $ensure  = 'present',
   String[1]                                               $key     = $name,
-  Optional[Variant[String[1], Numeric, Array[String[1]]]] $value   = undef,
+  Optional[Variant[String, Numeric, Array[String[1]]]]    $value   = undef,
   Stdlib::Absolutepath                                    $path    = $postgresql::server::postgresql_conf_path,
   Optional[String[1]]                                     $comment = undef,
   String[1]                                               $instance_name = 'main',
--- a/spec/defines/server/config_entry_spec.rb
+++ b/spec/defines/server/config_entry_spec.rb
@@ -79,4 +79,21 @@ describe 'postgresql::server::config_entry' do
         .that_notifies('Postgresql::Server::Instance::Service[main]')
     end
   end
+
+  # Example use case for empty values: using transaction-bound session
+  # variables, like `SET LOCAL awesome_app.tenant_id = 92834` requires the
+  # variable to be defined in postgresql.conf and initialised with a default
+  # value (at least for PostgreSQL 13 and older, possibly also in newer
+  # versions). The default value can (and sometimes, depending on the
+  # application should) be empty.
+  context 'set a config entry value to the empty string' do
+    let(:params) { { ensure: 'present', name: 'mydb.app_param', value: '' } }
+
+    it 'sets value to the empty string' do
+      expect(subject).to contain_postgresql_conf('mydb.app_param').with(
+        name: 'mydb.app_param',
+        value: '',
+      )
+    end
+  end
 end


antaflos:
I don't think I can do anything about the failing tests (labeller and mend). The rest looks good.

bastelfreak:
@antaflos can you please document in the spec test some examples of options where this makes sense? And please rebase after https://github.com/puppetlabs/puppetlabs-postgresql/pull/1599 got merged.

bastelfreak (dismissed):

deric (dismissed):

antaflos:
So this has been open for a long while now. Any chance to get it merged?

antaflos:
This is still open, even though approvals were given. What do I need to do to get this merged? 

Some acceptance tests are failing, but apparently not because of my changes.

spec/defines/server/config_entry_spec.rb
```diff
@@ -79,4 +79,19 @@
...
+  # variable to be defined in postgresql.conf and initialised with a default
+  # value (at least for PostgreSQL 13 and older, possibly also in newer
+  # versions). The default value can (and sometimes, depending on the
+  # application should) be empty.
+  context 'set a config entry value to the empty string' do
+    let(:params) { { ensure: 'present', name: 'mydatabase.app_specific_parameter', value: '' } }
+
+    it 'sets value to the empty string' do
+      expect(subject).to contain_postgresql_conf('mydatabase.app_specific_parameter').with(name: 'mydatabase.app_specific_parameter',
+                                                                                           value: '')
```
ekohl, line 94: Readability here isn't great. Have you considered
```suggestion
      expect(subject).to contain_postgresql_conf(mydatabase.app_specific_parameter').with(
        name: 'mydatabase.app_specific_parameter',
        value: '',
      )
```

Or perhaps using a shorter name.

manifests/server/config_entry.pp
```diff
@@ -10,7 +10,7 @@
 define postgresql::server::config_entry (
   Enum['present', 'absent']                               $ensure  = 'present',
   String[1]                                               $key     = $name,
-  Optional[Variant[String[1], Numeric, Array[String[1]]]] $value   = undef,
+  Optional[Variant[String, Numeric, Array[String[1]]]]    $value   = undef,
```
ekohl, line 13: Should the array also allow empty strings?

ekohl (commented):
I wondered about purging empty entries, but that's done with `Undef`:
https://github.com/puppetlabs/puppetlabs-postgresql/blob/d5911ecf6a87b7cddbe49f99d2c27d6ea512bce9/manifests/server.pp#L220-L225

spec/defines/server/config_entry_spec.rb
antaflos: Ok, I will make this more readable :+1:

manifests/server/config_entry.pp
antaflos: I am not aware of a use-case where an array would contain empty string values. 

But I am not even sure if array values make sense for Postgres config entries, because it seems to use just (quoted) strings, integers, floats and booleans (https://www.postgresql.org/docs/current/config-setting.html#CONFIG-SETTING-NAMES-VALUES). And from quickly looking at the `postgresql_conf` provider I don't immediately see where/how it even handles array values? But in any case, I don't think we need to touch this part of the functionality here. Just support for single, empty string values would be nice :)

antaflos: Indeed, it seems the accepted data types of `$value` were changed to include `Array[String[1]]`, and the "parsed" provider was replaced by a "ruby" provider, but there doesn't seem to have been any code added to actually handle such array values. The parsed provider had something like ` h[:value].join(', ')` when `h[:value]` was an array, but I see no equivalent in the ruby provider.

But I suppose this discussion is best moved elsewhere, probably to its own bug report?

ekohl: > Indeed, it seems the accepted data types of `$value` were changed to include `Array[String[1]]`, and the "parsed" provider was replaced by a "ruby" provider, but there doesn't seem to have been any code added to actually handle such array values. The parsed provider had something like ` h[:value].join(', ')` when `h[:value]` was an array, but I see no equivalent in the ruby provider.

If the implementation is ` h[:value].join(', ')` then an empty array should result in an empty string and allowing empty strings within the array doesn't make sense. It'd result in `, ` which is probably never valid. In other words, consider my comments addressed.

> But I suppose this discussion is best moved elsewhere, probably to its own bug report?

Yes, agreed.

ekohl (dismissed):

antaflos:
@ekohl Thank you for taking the time to look at and review this!

antaflos:
This is like pulling teeth. The spec tests finally work, so I updated mine to fix Rubocop complaints.

But now the acceptance tests on Debian and Ubuntu are broken, because they can not install Puppet agent because of the expired GPG keys at apt.puppet.com.

jst-cyr:
@antaflos : Sorry about the issues with the test framework. With so many of the gems and downloads behind auth we need to run acceptance tests on branches now. I'm rebasing this branch and spinning up a separate PR to get some acceptance test feedback on this PR soon.

jst-cyr:
Running CI tests for this PR here: https://github.com/puppetlabs/puppetlabs-postgresql/pull/1705

jst-cyr:
**Acceptance Test results:** https://github.com/puppetlabs/puppetlabs-postgresql/actions/runs/34372996860?pr=1705

All green except for provisioning issues with Debian 11 (known issue, matches nightly results against main). Tests look good!

Still needs an engineering review when the team has time.

PR #1609, opened 2024-08-23
Add a fact for installed postgres version

maggu:
## Summary
There is sometimes a need to handle versions differently.  (The last exampel for us being deciding whether to monitor the stats collector process.)  This fact provides a basis for writing such code.

## Checklist
- [X] 🟢 `pdk validate`
- [X] 🟢 `pdk test unit`

### DIFF
--- a/lib/facter/postgres_version.rb
+++ b/lib/facter/postgres_version.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+Facter.add('postgres_version') do
+  confine { Facter::Core::Execution.which('postgres') }
+  setcode do
+    version = Facter::Core::Execution.execute('postgres -V 2>/dev/null')
+    version.match(%r{\d+\.\d+$})[0] if version
+  end
+end
--- a/spec/unit/facter/postgres_version_spec.rb
+++ b/spec/unit/facter/postgres_version_spec.rb
@@ -1,0 +1,20 @@
+require 'spec_helper'
+
+describe Facter::Util::Fact.to_s do
+  before(:each) do
+    Facter.clear
+  end
+
+  describe 'postgres_version' do
+    context 'with value' do
+      before :each do
+        allow(Facter::Core::Execution).to receive(:which).and_return('/usr/bin/postgres')
+        allow(Facter::Core::Execution).to receive(:execute).with('postgres -V 2>/dev/null').and_return('postgres (PostgreSQL) 10.0')
+      end
+
+      it 'postgres_version' do
+        expect(Facter.fact(:postgres_version).value).to eq('10.0')
+      end
+    end
+  end
+end


lib/facter/postgres_version.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Facter.add('postgres_version') do
+  confine { Facter::Core::Execution.which('postgres') }
```
kenyon, line 4: At least on Debian family, `postgres` is not in `PATH`, so this wouldn't work. Maybe it'd be better to get the version from `psql`, which is in the `PATH`?

lib/facter/postgres_version.rb
```diff
@@ -0,0 +1,9 @@
...
+  setcode do
+    version = Facter::Core::Execution.execute('postgres -V 2>/dev/null')
+    version.match(%r{\d+\.\d+$})[0] if version
```
kenyon, line 7: This regex won't match the output of Debian family packages:

```console
kenyon@lithium ~ % /usr/lib/postgresql/14/bin/postgres -V
postgres (PostgreSQL) 14.13 (Debian 14.13-1.pgdg110+1)
```

lib/facter/postgres_version.rb
maggu: The problem with that is that `psql`might be a different version, and I would like to ensure that it's actually the database server version that gets reported. I guess this needs a bit more thought.

lib/facter/postgres_version.rb
maggu: Okay. Guess the regex will have to be adjusted. This is RHEL 9 by the way, and I believe also vanilla PostgreSQL:

```
% /bin/postgres -V
postgres (PostgreSQL) 13.14
%
```

maggu:
> There might be multiple PostgreSQL clusters installed with different versions. Debian has `pg_lsclusters` script.

RHEL doesn't unfortunately, do we can't depend on it for this. A quick Google search indicates that pg_lsclusters` is a Debian specific thing.

deric:
@maggu Yes, it's Debian specific. It's fairly simple Perl script. AFAIK postgresql doesn't ship any tool to manage/list clusters. You could try to parsing `$confdir` but that can be in very different locations. You could export clusters managed by Pupppet to a text/JSON file and then load it as a fact.

bastelfreak:
I think in this case it's fine to only have a 80% solution. If it works for the common setups and the edge cases are documented, that is fine in my opinion.

maggu:
> I think in this case it's fine to only have a 80% solution. If it works for the common setups and the edge cases are documented, that is fine in my opinion.

I agree. If there's a need for the rest, a second hierarchical fact confined by and using `pg_lsclusters` could always be added in the future.

ekohl:
Fedora 40 has already moved to parallel installable packages for PostgreSQL (by creating `postgresqlXX` packages) and I expect future RHEL versions to follow. It sounds to me that tooling like `pg_lsclusters` should be included.

> There is sometimes a need to handle versions differently. (The last exampel for us being deciding whether to monitor the stats collector process.) This fact provides a basis for writing such code.

The version is already available in the globals as `$postgresql::globals::version`. By default detected rather statically but if you need a non-standard version then you modify it. The further code sets up the right repositories or in case of RHEL 8 & 9 OS packages the right DNF module. I don't see a need for a fact here.

maggu:
Debian appears to have `pg_config --version` as part of the standard installation. I believe that could perhaps solve the problem of making this work on the major Linux platforms. I don't currently plan to put more work into it though unless the consensus is that a version fact would indeed be useful.

lib/facter/postgres_version.rb
maggu: As stated below, perhaps `pg_config --version` could be used.

ekohl:
> It has worked reasonably well sometimes. Other times though, when the `version` value is `undef` and the fallback used, the result has turned out to not be what we wanted. I would be interested to hear how this approach has worked out for others though. Switching to using a fact to determine the _actual_ version so far at least seems to be more consistent over time.

I think it should be the other way around: you specify the version you want and the module makes it so.

Fundamentally the problem is that until you install the package, you don't have the version. So it can never be idempotent.

* Run 1
  * Install package with some configuration
* Run 2
  * Fact is available
  * Modifications may be made

In addition to that, we have multi instance mode where I think even one instance may run a different version than another and I question how this fact is going to be useful in the generic case.

> I don't see us going back to using the globals parameter, even though I suppose to could be made to work with enough effort. We have a lot of people working with Puppet code in multiple departments and can't depend on it being explicitly set everywhere. Perhaps we're special enough though, that it's better that we just use a local fact instead of trying to merge this upstream.

Perhaps other maintainers have a different opinion, but as a maintainer of this module I'd say that you're not using the module correctly. It is designed that you pass in the version you want.

maggu:
> Fundamentally the problem is that until you install the package, you don't have the version. So it can never be idempotent.

It would indeed require two runs to reach a final state. Isn't that the case for _any_ fact except for the core facts? While I agree that in general fewer runs are better than more, there are also other aspects to consider. In this case, first and foremost the amount of work required to keep explicit versions updated.

> In addition to that, we have multi instance mode where I think even one instance may run a different version than another and I question how this fact is going to be useful in the generic case.

That would be a limitiation to this PR, yes (as discussed above). Unless the scope is changed, and that is of course a possibility. 

> Perhaps other maintainers have a different opinion, but as a maintainer of this module I'd say that you're not using the module correctly. It is designed that you pass in the version you want.

I personally don't think the documentation currently reflects that indention very well, and perhaps ought to be updated. (Also, for reasons stated above I don't believe this approach would work very well for us.)

ekohl:
> It would indeed require two runs to reach a final state. Isn't that the case for _any_ fact except for the core facts? While I agree that in general fewer runs are better than more, there are also other aspects to consider. In this case, first and foremost the amount of work required to keep explicit versions updated.

That is why I think detecting the version via a fact is considered unreliable. You can query the repositories for the available version and use that, but what if the manifests set up the repositories?

We have this in https://github.com/voxpupuli/puppet-nginx and there it's only causing problems.

> I personally don't think the documentation currently reflects that intention very well, and perhaps ought to be updated.

That may be true. I see https://github.com/puppetlabs/puppetlabs-postgresql?tab=readme-ov-file#override-defaults is there, but it could be promoted to its own chapter so I opened https://github.com/puppetlabs/puppetlabs-postgresql/pull/1610

(Also, for reasons stated above I don't believe this approach would work very well for us.)

Could you elaborate on that? Because I'm inclined to consider it unsupported. This module takes great care of minimizing the amount of changes to a database server because restarting a database server is considered undesirable. Anything that can risk that should be treated with great caution.

jst-cyr:
Hello @maggu ... It seems this PR has been sitting for some time and I've found it while reviewing old open PRs. Based on the conversation here, it seems that the general feeling is that this PR is not quite ready and may not be a good fit. I'm going to close this out for now and it can always be re-opened if needed!

Issue #1575, opened 2024-02-18
postgres_password throws errors when Deferred


bobnegri:
## Describe the Bug
When `postgres_password` is Deferred, it throws one of two possible errors:

Error 1:
> $hash_postgresql_db_pw = Deferred('postgresql::postgresql_password', ['puppetdb', $plain_postgresql_db_pw])
> 
> Error: Failed to apply catalog: Function postgresql::postgresql_password(): Unknown function: 'postgresql::default'

Error 2:
> # Get the hash value: md5 or scram-sha-256
> $pwd_encrypt = $postgresql::params::password_encryption
> 
> $hash_postgresql_db_pw = Deferred('postgresql::postgresql_password', ['puppetdb', $plain_postgresql_db_pw, false, $pwd_encrypt])
> 
> Failed to apply catalog: 'postgresql::postgresql_password' parameter 'hash' references an unresolved type 'Postgresql::Pg_password_encryption'

## Expected Behavior

Expectation 1:
When function `postgresql::postgresql_password` is Deferred and only passed `username` and `password`, the function should use the built in defaults without throwing an error.

Expectation 2:
When function `postgresql::postgresql_password` is Deferred and values are provided for `username`, `password`, `sensitive`, and `hash`, the function should not throw a type error.

## Environment
 - postgresql-12 10.0.3
 - puppet-agent   7.27.0
 - puppetdb          7.15.0
 - puppetserver    7.14.0
 - Ubuntu 20.04.6 LTS

## Additional Context

Under postgresql 9.2.0, we did not provide `sensitive` and `hash` values and the function worked when Deferred.

ekohl:
Is this a duplicate of https://github.com/puppetlabs/puppetlabs-postgresql/issues/1560?

bobnegri:
I'm going to go with, no.

I'm Deferring function 'postgresql::postgresql_password':

`$hash_postgresql_db_pw = Deferred('postgresql::postgresql_password', ['puppetdb', $plain_postgresql_db_pw])`

Issue [#1560](https://github.com/puppetlabs/puppetlabs-postgresql/issues/1560) is passing in a Deferred to the class.
```
class { '::postgresql::server':
    postgres_password => Deferred('unwrap', [$admin_password_vault]),
}
```

bastelfreak:
> Expectation 2:
When function postgresql::postgresql_password is Deferred and values are provided for username, password, sensitive, and hash, the function should not throw a type error.

[...]

PR #1611, opened 2024-08-30
define Enum on supported encryption types for postgresql_password

figless:
## Summary
This PR fixes [1575](https://github.com/puppetlabs/puppetlabs-postgresql/issues/1575)

## Checklist
I was hitting [1575](https://github.com/puppetlabs/puppetlabs-postgresql/issues/1575) when using a Deferred password. With this fix, puppet creates the database as before (v9.2.0)

```
postgresql::server::db { $db_name:
      user     => $db_username,
      password => Deferred('unwrap', [$db_password]),
}
```

### DIFF
--- a/spec/acceptance/db_deferred_spec.rb
+++ b/spec/acceptance/db_deferred_spec.rb
@@ -2,45 +2,32 @@
 
 require 'spec_helper_acceptance'
 
-describe 'postgresql::server::db' do
-  before(:all) do
-    LitmusHelper.instance.run_shell("cd /tmp; su 'postgres' -c 'pg_ctl stop -D /var/lib/pgsql/data/ -m fast'", acceptable_exit_codes: [0, 1]) unless os[:family].match?(%r{debian|ubuntu})
-  end
+describe 'postgresql::server::db:' do
+  let(:user) { 'user_test' }
+  let(:password) { 'deferred_password_test' }
+  let(:database) { 'test_database' }
 
-  it 'creates a database with a deferred password' do
-    tmpdir = run_shell('mktemp').stdout
-    pp = <<-MANIFEST
-      class { 'postgresql::server':
-        postgres_password => 'space password',
-      }
-      postgresql::server::tablespace { 'postgresql-test-db':
-        location => '#{tmpdir}',
-      } ->
-      postgresql::server::db { 'postgresql-test-db':
-        comment    => 'testcomment',
-        user       => 'test-user',
-        password   => Deferred('unwrap', ['test1'])
-        tablespace => 'postgresql-test-db',
+  let(:pp_one) do
+    <<-MANIFEST.unindent
+      $user = #{user}
+      $password = #{password}
+      $database = #{database}
+
+      include postgresql::server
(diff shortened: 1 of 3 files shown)


ekohl (changes requested):
We already have defined this as an enum:
https://github.com/puppetlabs/puppetlabs-postgresql/blob/05a78097377eebd34297e9a01054f4bd50507b79/types/pg_password_encryption.pp#L2

And it's used in multiple places now, so this would risk them going out of sync. I'd at least expect a comment in both places to remind people to keep them in sync.

figless:
> We already have defined this as an enum:
> 
> https://github.com/puppetlabs/puppetlabs-postgresql/blob/05a78097377eebd34297e9a01054f4bd50507b79/types/pg_password_encryption.pp#L2
> 
> And it's used in multiple places now, so this would risk them going out of sync. I'd at least expect a comment in both places to remind people to keep them in sync.

I've added a comment to both files.

bastelfreak:
Is there a chance you can add a test with `Deferred()`, to ensure this doesn't break in the future?

figless:
> Is there a chance you can add a test with `Deferred()`, to ensure this doesn't break in the future?

Sure - i've copied `spec/acceptance/db_spec.rb` to `spec/acceptance/db_deferred_spec.rb` and configured the test to use a `Deferred` password.

figless:
The previous code in `db_deferred_spec.rb` failed the CI, even though it was essentially a copy from `db_spec.rb`....
I have adapted the code, which now passes locally.

smortex:
I can't trigger the CI, maybe because the branch is outdated. Let's try to rebase it…

ekohl:
I needed to approve ci, which I just did

figless:
Hello,
I've just rebased this branch.
Can the CI please be retriggered?

bastelfreak (approved):

spec/acceptance/db_deferred_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+require 'spec_helper_acceptance'
+
+describe 'postgresql::server::db:' do
+  let(:user) { 'user_test' }
+  let(:password) { 'deferred_password_test' }
+  let(:database) { 'test_database' }
+
+  let(:pp_one) do
+    <<-MANIFEST.unindent
```
ekohl:
```suggestion
    <<~MANIFEST
```

spec/acceptance/db_deferred_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+         user     => $user,
+         password => Deferred('unwrap', [$password]),
+      }
+    MANIFEST
+  end
+
+  it 'creates a database with with the password in the deferred function' do
+    if run_shell('puppet --version').stdout[0].to_i < 7
+      skip # Deferred function fixes only in puppet 7, see https://tickets.puppetlabs.com/browse/PUP-11518
+    end
```
ekohl: I wouldn't care about Puppet 6 and older
```suggestion
```

PR #1624, opened 2024-12-18
Role Valid Until Date

mooreandrew:
## Summary
I have a requirement to set valid until dates for roles that I need to manage via this puppet postgresql module. This change adds support for setting a `valid_until` attribute for a role.

## Related Issues (if any)
None

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.

### DIFF
--- a/manifests/server/role.pp
+++ b/manifests/server/role.pp
@@ -11,6 +11,7 @@
 # @param inherit Specifies whether to grant inherit capability for the new role.
 # @param superuser Specifies whether to grant super user capability for the new role.
 # @param replication Provides provides replication capabilities for this role if set to true.
+# @param valid_until Specifies whether to set a valid until date for the role.
 # @param connection_limit Specifies how many concurrent connections the role can make. Default value: '-1', meaning no limit.
 # @param username Defines the username of the role to create.
 # @param connect_settings Specifies a hash of environment variables used when connecting to a remote server.
--- a/manifests/server/role.pp
+++ b/manifests/server/role.pp
@@ -35,6 +35,7 @@ define postgresql::server::role (
   Boolean                                     $inherit          = true,
   Boolean                                     $superuser        = false,
   Boolean                                     $replication      = false,
+  Optional[String[1]]                         $valid_until      = undef,
   String[1]                                   $connection_limit = '-1',
   String[1]                                   $username         = $title,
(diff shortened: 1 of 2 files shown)


smortex:
See CI failures about missing documentation. Also, adding tests to `spec/defines/server/role_spec.rb` to ensure the feature don't break in the future would be nice.

mooreandrew:
@smortex I've updated the code, adding documentation for the variable and tests to `spec/defines/server_instance_spec.rb` in the same way as the other alter commands.

Let me know if that is good or if you need me to make any further changes.

smortex (approved):

mooreandrew:
@smortex The pipeline all appears to look good. How can I got this change merged?

alexjfisher:
Looks okay to me too.  Only query is should this have a mechanism for removing the expiry date?  eg. perhaps if the parameter is set to a special value such as the string `'absent'` or maybe the boolean `false` (in which case you'd probably need `unless $valid_until == undef { `instead of `if $valid_until {` for your conditional), then `rolvaliduntil ` should be set to `NULL` (unless it's already `NULL`)

smortex:
> Looks okay to me too. Only query is should this have a mechanism for removing the expiry date? eg. perhaps if the parameter is set to a special value such as the string `'absent'` or maybe the boolean `false` (in which case you'd probably need `unless $valid_until == undef { `instead of `if $valid_until {` for your conditional), then `rolvaliduntil ` should be set to `NULL` (unless it's already `NULL`)

That would make sense, @mooreandrew feel free to tackle this if you want to. In the meantime, this PR is fine so I'll merge it as it is now.

Thanks!

alexjfisher:
> > Looks okay to me too. Only query is should this have a mechanism for removing the expiry date? eg. perhaps if the parameter is set to a special value such as the string `'absent'` or maybe the boolean `false` (in which case you'd probably need `unless $valid_until == undef { `instead of `if $valid_until {` for your conditional), then `rolvaliduntil ` should be set to `NULL` (unless it's already `NULL`)
> 
> That would make sense, @mooreandrew feel free to tackle this if you want to. In the meantime, this PR is fine so I'll merge it as it is now.
> 
> Thanks!

Yeah, I'm fine with that.  This MR is safe as-is, and the additional functionality can be added later in a non disruptive way.

PR #1633, opened 2025-04-01
Rename spec files as _spec.rb so they run

traylenator:
A batch of spec files were never being executed due to having the wrong name.

Tests have all been marked pending as I expect they have never passed.

Fixing them , deleting them or updating how the module works is best handled elsewhere as not very obvious.


traylenator:
Easiest solution is declare the types private api and remove the tests but I'm not sure that was the original intention.

ekohl:
I recall there was a rubocop-rspec plugin that looked for these wrong filenames.

> Easiest solution is declare the types private api and remove the tests but I'm not sure that was the original intention.

@SimonHoenscheid mind taking a look since you wrote this?

jst-cyr:
This requires #1691 to be able to run tests.

jst-cyr:
Running CI for this PR here: https://github.com/puppetlabs/puppetlabs-postgresql/pull/1698

jst-cyr:
**Acceptance test results:** https://github.com/puppetlabs/puppetlabs-postgresql/actions/runs/32537000723

Mostly green acceptance tests. All acceptance-test failures in this PR's CI run match existing failures already present on main's nightly run (CentOS-7/OracleLinux-7/Scientific-7 systemd D-Bus service-start issue, Debian-13 default_privileges idempotency failure, SLES-12 provisioning timeout). None are caused by this change.

Next step requires an engineering review.

PR #1638, opened 2025-05-29
Add support for Debian 13 (trixie)

trefzer:
add support for the upcoming debian trixie release.

### DIFF
--- a/data/os/Debian/11.yaml
+++ b/data/os/Debian/11.yaml
@@ -1,0 +1,4 @@
+--- 
+# Defaults for Debian Bullseye (11)
+
+postgresql::globals::python_package_name: 'python3-psycopg2'
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -210,6 +210,7 @@ class postgresql::globals (
         '10'    => '11',
         '11'    => '13',
         '12'    => '15',
+        '13'    => '17',
         default => undef,
       },
       'Ubuntu' => $facts['os']['release']['major'] ? {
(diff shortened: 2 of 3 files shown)


kenyon (approved):

data/os/Debian/11.yaml
```diff
@@ -0,0 +1,4 @@
+--- 
+# Defaults for Debian Bullseye (11)
+
+postgresql::globals::python_package_name: 'python3-psycopg2'
```
bastelfreak, line 4: This change broke the tests. And I don't know why a Debian 11 change is in the Debian 13 PR. The default for Debian 11 is set at https://github.com/puppetlabs/puppetlabs-postgresql/blob/main/manifests/params.pp#L97 . But that default seems to be wrong, because the actual package name is python3-psycopg2 ? See https://packages.debian.org/search?suite=bullseye&searchon=names&keywords=python3-psycopg2

* So the code in the module for debian 11 was wrong for years?
* This change fixed the code, but in the wrong place
* it broke the tests, they still use the old name
* this change doesn't relate to adding debian 13 support?

ekohl: > So the code in the module for debian 11 was wrong for years?

It appears that way. In https://github.com/puppetlabs/puppetlabs-postgresql/pull/1557#discussion_r1403626309 I wonder why I stated that there as a `python-psycopg2` for Debian 11.

> This change fixed the code, but in the wrong place

Correct. Including this here is weird: it negates the code we have for it:
https://github.com/puppetlabs/puppetlabs-postgresql/blob/b85be451c87dfee0ec245615b0293a04b9df6220/manifests/params.pp#L166-L170

And then all your other observations are also correct.

PR #1647, opened 2025-09-30
Debian 11: Use correct package name for python bindings 

bastelfreak:
Since Debian 11 was added to this module, the package name was wrong. It
was set to python-psycopg2, but the correct name is python3-psycopg2.

There's a longer analysis about this bug in https://github.com/puppetlabs/puppetlabs-postgresql/pull/1638/files#r2390588710

tl;dr:

* Debian 10 used python-psycopg2 and that was set by the module
* Debian 11 was added without proper acceptance testing
* Later on unit tests were added for Debian 11, but still with the wrong
  package name
* https://github.com/puppetlabs/puppetlabs-postgresql/commit/651c2b4ec0a3d9a76770369625b0a91bcba444ff fixed the package name, but
  in the wrong place and didn't adjust the tests
* That resulted in a broken pipeline, that's how I found all of this

This patch:
* Cleans up the hiera data
* sets the package name at the correct location
* adjusts the debian 11 unit test for the new package name
* Adds tests for Debian 12 and 13 (won't be executed right now because Perforce ships a too old version of FacterDB)

Edit: Had to remove Debian 13 because the pipeline fails if it cannot find a factset.

### DIFF
--- a/data/os/Debian/11.yaml
+++ b/data/os/Debian/11.yaml
@@ -1,4 +1,0 @@
---- 
-# Defaults for Debian Bullseye (11)
-
-postgresql::globals::python_package_name: 'python3-psycopg2'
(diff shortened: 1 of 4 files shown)


deric (approved):

bastelfreak:
The acceptance tests are a mess, due to the broken pipeline. The adjusted unit tests pass now again, so I'm going to merge it.

PR #1652, opened 2025-12-13
Replace systemctl `status` with systemctl `is-active`

bastelfreak:
Replace systemctl `status` with systemctl `is-active`

`systemctl status $service` will output a bunch of text. Some of this text, depending on the locale, contains unicode characters:

```
● postgresql.service - PostgreSQL RDBMS
     Loaded: loaded (/lib/systemd/system/postgresql.service; enabled; preset: enabled)
     Active: active (exited) since Wed 2025-09-24 21:40:00 UTC; 2 months 18 days ago
   Main PID: 1028 (code=exited, status=0/SUCCESS)
        CPU: 3ms

Notice: journal has been rotated since unit was started, output may be incomplete.
```

`●` is unicode. Puppet will fail with:

```
Error: /Stage[main]/Postgresql::Server::Reload/Postgresql::Server::Instance::Reload[main]/Exec[postgresql_reload_main]: Failed to call refresh: invalid byte sequence in US-ASCII
Error: /Stage[main]/Postgresql::Server::Reload/Postgresql::Server::Instance::Reload[main]/Exec[postgresql_reload_main]: invalid byte sequence in US-ASCII
```

`systemctl` will print `*` if it's executed with `LC_ALL=C`, but we cannot assume that this locale is always used.

The correct command to check if a service is up and running is `is-active`, not `status`. `status` is for humans only. the systemd provider for the service type also uses `is-active`. See https://github.com/puppetlabs/puppet/blob/main/lib/puppet/provider/service/systemd.rb#L201-L203

Since we only care about the exit code, we can even add `--quiet`. See https://www.freedesktop.org/software/systemd/man/latest/systemctl.html#is-active%20PATTERN%E2%80%A6

## Summary

Ensure the service detection works with all locales.

## Additional Context
- [x] Root cause and the steps to reproduce. (If applicable)
- [x] Thought process behind the implementation.

## Related Issues (if any)

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -86,7 +86,7 @@ class postgresql::params inherits postgresql::globals {
       }
 
       $service_reload = "systemctl reload ${service_name}"
-      $service_status = pick($service_status, "systemctl status ${service_name}")
+      $service_status = pick($service_status, "systemctl is-active --quiet ${service_name}")
 
       $psql_path           = pick($psql_path, "${bindir}/psql")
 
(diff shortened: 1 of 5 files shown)


deric (approved):

bastelfreak:
Yeah, I hope someone fixes the CI and removes the puppet 7 checks soon :(

Issue #1656, opened 2026-01-12
config_entry removal incorrectly depends on existing value in postgresql.yml


davidassigbi:
## Describe the Bug
When setting `postgresql::server::config_entry` resources to `ensure => absent`, the provider incorrectly requires the provided value to either match the one currently in `postgresql.yml` or be empty (`undef`, or anything that renders as an empty string).

The confusing behavior is that if the value provided does not match what's in `postgresql.yml`, Puppet will output that it's removing the configuration, but the entry remains in the file. As a result, every subsequent run repeats the same “removal” message, creating an endless loop of false reporting.

## Expected Behavior
When ensure is set to `absent`, the provider should just completely ignore the `value` field.

## Steps to Reproduce
1. Have a this module properly setup and already running and managing a postgres instance
This is a sample `site.pp`
```puppet
  class { 'postgresql::globals':
    manage_package_repo => true,
    version             => '16',
  }

  class { 'postgresql::server': }
```
2. Define a `postgresql::server::config_entry` resource in your` site.pp`. For example, initially set `log_rotation_size` to `45MB`.
```puppet
  postgresql::server::config_entry { 'log_rotation_size':
    value  => '45MB',
  }
```
3. Run `puppet agent -t` and check that `postgresql.yml` contains the correct value
```puppet
  postgresql::server::config_entry { 'log_rotation_size':
    ensure  => absent,
    value    => '50MB',
  }
``` 
4. Add `ensure => absent,` and change the value to `50MB`
5. Run `puppet agent -t`. There should be a notice that `log_rotation_size` was removed ✅ 
6. Check `postgresql.yml` for `log_rotation_size` and be amazed it's still there 🥹 
7. Repeat the last 2 steps and go insane 

[pp_postgresql_conif_entries_bug.log](https://github.com/user-attachments/files/24564927/pp_postgresql_conif_entries_bug.log)

## Environment 
- Version v10.6.1  
- Platform [e.g. Ubuntu 24.04]

PR #1657, opened 2026-01-12
fix: ignore postgresql_conf resource value when set to absent

davidassigbi:
## Summary
This PR makes the `postgresql_conf` provider ignore the `value` field when `ensure `is set to `absent`.

## Additional Context
Much more info in the related issue.

## Related Issues (if any)
config_entry removal incorrectly depends on existing value in postgresql.yml #1656

## Checklist
- [x] 🟢 Spec tests.
- [x] 🟢 Acceptance tests.

### DIFF
--- a/lib/puppet/provider/postgresql_conf/ruby.rb
+++ b/lib/puppet/provider/postgresql_conf/ruby.rb
@@ -80,7 +80,7 @@ Puppet::Type.type(:postgresql_conf).provide(:ruby) do
 
   # remove resource if exists and is set to absent
   def destroy
-    entry_regex = %r{#{resource[:key]}.*=.*}
+    entry_regex = %r{^\s*#{resource[:key]}\s*=}
     lines = File.readlines(resource[:target])
 
     lines.delete_if do |entry|


lib/puppet/provider/postgresql_conf/ruby.rb
```diff
@@ -80,7 +80,7 @@ def exists?
 
   # remove resource if exists and is set to absent
   def destroy
-    entry_regex = %r{#{resource[:key]}.*=.*#{resource[:value]}}
+    entry_regex = %r{#{resource[:key]}.*=.*}
```
smortex: Wow, this regex seems inadequate 😄 

I have not checked in details, but I guess that if `resource[:key]` is "foo" we want to remove only a single line:

```
foo = bar # We want to remove this one
foobar = baz # We do not want to remove this one
barfoo = baz # We do not want to remove this one
```

If so, I suggest:

```suggestion
    entry_regex = %r{\A#{resource[:key]}[[:blank:]]*=}
```

or in case a line can start with some space:


```suggestion
    entry_regex = %r{\A[[:blank:]]*#{resource[:key]}[[:blank:]]*=}
```

davidassigbi: Ah, u are completely right ! The previous regex despite being 'wrong' at least would use the `value` as a sort of safeguard in cases where the `value` was still provided with `ensure => absent`. But with this change it gets easier to mess with lines we shouldn't.

Thx for pointing this out ! 🙏 

I’ll apply this suggestion, but I’ll use `\s` instead of `[[:blank:]]` to stay consistent with how other regexes are written in the provider.

smortex (approved):

PR #1662, opened 2026-02-23
Update table_grant.pp to allow passing array as table name

mikhaildudzin:
## Summary
Array required to allow grants on tables in specific schema

## Additional Context
it should be consistent with
https://github.com/puppetlabs/puppetlabs-postgresql/blob/49368d1eb7585b151960642b95bfe3f8b09a15c5/manifests/server/grant.pp#L41

## Checklist
- [ ] 🟢 Spec tests.
- [ ] 🟢 Acceptance tests.


jst-cyr:
Sorry for the delay in responding @mikhaildudzin , going through PRs to try to catch up on them now. Are you able to make sure that tests are added to check that this functionality works as expected and also does not impact existing users?

jst-cyr:
@mikhaildudzin : As a note, there is feedback from the spec tests:

`manifests/server/table_grant.pp - WARNING: Optional defined type parameter doesn't have a default on line 19 (check: optional_default)`

PR #1673, opened 2026-05-27
add loop through grant_role type to allow predefined roles to be used in existing databases

ikonia:
## Summary
previous MR's have added parameters, roles to align to the role type, grants to allow to the grant type and databases to align to the database type, which allow the creation of resources from values within hiera. 

This additional loop following the same pattern as the other approved MR's loops through grant_roles to the grant_role type, which allow the use of pre-defined roles within Postgres such as https://www.postgresql.org/docs/current/predefined-roles.html which cannot be set with the grant type alone.

## Additional Context

Following existing working patterns to extend the functionality of the module in a least complex way

## Related Issues (if any)

## Checklist
- [ x] 🟢 Spec tests.
- [x ] 🟢 Acceptance tests.
- [x ] Manually verified. (For example `puppet apply`)

no changes needed to the spec, acceptance tests not required I believe based on the other MR's of this type. Tested against pre-defined roles.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -103,6 +103,7 @@
 # @param grants Specifies a hash from which to generate postgresql::server::grant resources.
 # @param config_entries Specifies a hash from which to generate postgresql::server::config_entry resources.
 # @param pg_hba_rules Specifies a hash from which to generate postgresql::server::pg_hba_rule resources.
+# @param grant_roles Specifies a hash from which to generate postgresql::server::grant_role resources. 
 #
 # @param backup_enable Whether a backup job should be enabled.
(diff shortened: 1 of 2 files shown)


ikonia:
would appreciate a review on approach on this, especially the database creation fix for if the database doesn't exist before the extension is applied

ikonia:
I see my change to fix the dependency cycle removing Class['postgresql::server'] -> Postgresql_psql["grant_role:${name}"] that looked obsolete because it appeared to never be called from outside the class has broke a test suit. I'll look into if this test is actually still valid (I suspect it is and I'll need to change this PR) or if it's orphaned from an earlier version of the module

PR #1686, opened 2026-07-07
(MODULES-11858) Restore Puppet 7 support broken by 10.6.2

### LINKED ISSUE #1685: Invalid ruby syntax in 10.6.2
## Describe the Bug

When using postgresql::postgresql_password in 10.6.2, it fails like so:
```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Internal Server Error: org.jruby.exceptions.SyntaxError: (SyntaxError) /etc/puppetlabs/code/environments/<redacted>/modules/postgresql/lib/puppet/functions/postgresql/postgresql_password.rb:69: syntax error, unexpected ','                                                                                                                                                                                      
      salt:,                                                                                                                                                                                                                                                
           ^                                                                                                                                                                                                                                                
```

Reverting to 10.6.1 fixes the issue. It looks like a typo in the ruby code after the latest version. I don't think the trailing `:` is valid ruby.

## Expected Behavior
It doesn't fail.

## Steps to Reproduce
1. Use puppet-postgresql version 10.6.2
2. Run `postgresql::postgresql_password('test', "test123")`
3. Running puppet should make it fail

## Environment
 - Version 10.6.2
 - Platform Rocky 9.8

imaqsood:
## Summary

Fixes #1685 (MODULES-11858). `10.6.2` was tagged a **patch** but PR #1646 (`CAT-2385 Puppetcore update`) introduced a hard syntax error that breaks existing installations on Puppet 7.

`.rubocop.yml` bumped `TargetRubyVersion` to `3.1`, so `rubocop -A` rewrote hash literals to Ruby 3.1 **shorthand** (`salt:` instead of `salt: salt`) in three custom functions/providers. That shorthand is a *syntax error* on Ruby 2.7 — the Ruby that Puppet 7 ships — so any Puppet 7 catalog using `postgresql::postgresql_password`, `postgresql_conf`, or `postgresql_replication_slot` fails to compile with `syntax error, unexpected ','` (#1685).

## What this PR does

Reverts **only** the Ruby-side breakage:

| File(s) | Change |
|---|---|
| `postgresql_password.rb`, `postgresql_conf/ruby.rb`, `postgresql_replication_slot/ruby.rb` | Revert Ruby 3.1 shorthand hash → explicit form (keeps the unrelated `postgresql_conf` `entry_regex` bugfix from #1657) |
| `.rubocop.yml` | `TargetRubyVersion` back to `'2.6'` so autocorrect can't reintroduce the shorthand |
| 4 spec files | Revert the same shorthand autocorrection so they parse under the `2.6` target |

The puppet requirement in `metadata.json` is left at `>= 8.0.0` per review — this PR is scoped to the Ruby syntax fix only. The puppetcore8 `Gemfile`/CI changes from #1646 also stay in place (dev/test only, not consumer-facing).

## Testing

- `pdk validate` → clean (exit 0)
- Full unit suite → **685 examples, 0 failures**
- **Litmus end-to-end** on Rocky 9, both agents:
  - **Puppet 7.34 / Ruby 2.7.8** — reproduced #1685 with the old code, then confirmed fixed: `postgresql::server` install → idempotent second run, `testdb` created, and a **SCRAM-SHA-256** password hash written to the live database (the exact `digest_key` path that was crashing).
  - **Puppet 8.10 / Ruby 3.2.5** — same install + idempotency + SCRAM checks pass (no forward regression).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,7 +5,7 @@ require:
 AllCops:
   NewCops: enable
   DisplayCopNames: true
-  TargetRubyVersion: 3.1
+  TargetRubyVersion: '2.6'
   Include:
   - "**/*.rb"
   Exclude:
--- a/lib/puppet/functions/postgresql/postgresql_password.rb
+++ b/lib/puppet/functions/postgresql/postgresql_password.rb
@@ -66,7 +66,7 @@ Puppet::Functions.create_function(:'postgresql::postgresql_password') do
   def digest_key(password, salt)
     OpenSSL::KDF.pbkdf2_hmac(
       password,
-      salt:,
+      salt: salt,
       iterations: 4096,
       length: 32,
       hash: OpenSSL::Digest.new('SHA256'),
--- a/lib/puppet/provider/postgresql_conf/ruby.rb
+++ b/lib/puppet/provider/postgresql_conf/ruby.rb
@@ -29,7 +29,7 @@ Puppet::Type.type(:postgresql_conf).provide(:ruby) do
                 else
                   matches[:value].delete("'")
                 end
-        attributes_hash = { line_number:, key: matches[:key], ensure: 'present', value:, comment: matches[:comment] }
+        attributes_hash = { line_number: line_number, key: matches[:key], ensure: 'present', value: value, comment: matches[:comment] }
         active_settings.push(attributes_hash)
       end
     end
--- a/lib/puppet/provider/postgresql_replication_slot/ruby.rb
+++ b/lib/puppet/provider/postgresql_replication_slot/ruby.rb
@@ -7,7 +7,7 @@ Puppet::Type.type(:postgresql_replication_slot).provide(:ruby) do
   def self.instances
     run_sql_command('SELECT * FROM pg_replication_slots;')[0].split("\n").select { |l| l.include?('|') }.map do |l|
       name, *_others = l.strip.split(%r{\s+\|\s+})
-      new(name:,
+      new(name: name,
           ensure: :present)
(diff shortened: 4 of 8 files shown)


metadata.json
```diff
@@ -103,7 +103,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 8.0.0 < 9.0.0"
+      "version_requirement": ">= 7.0.0 < 9.0.0"
```
ekohl: You don't have to do this part. I'd keep this just focused on the Ruby side.

imaqsood: Good call — dropped the metadata.json change and force-pushed. The PR is now focused purely on the Ruby side: reverting the Ruby 3.1 shorthand hash syntax in the three lib files (plus four spec files that `rubocop -A` had also rewritten) and putting `.rubocop.yml` `TargetRubyVersion` back to `2.6` so it can't be reintroduced. The puppet requirement stays at `>= 8.0.0`.

ekohl (approved):
Untested, but from a first glance this looks correct.

metadata.json
jst-cyr: While it might be good to keep the PR focused on just Ruby so that the change is encapsulated nicely, we cannot introduce a breaking change like removing Puppet 7 support in a patch release. That would require a major version release.

We can do it as two PRs, if folks prefer, but ultimately we also need the puppet requirements to stay the same on the 10.6.x versions.

PR #1661, opened 2026-02-15
support database creation from hiera parameters

ikonia:
## Summary
added $databases parameter to server.pp to allow iteration from server::database.pp to create databases 
reusing the existing included pattern of $roles iterating through server::role.pp to manage roles

## Additional Context
Add any additional context about the problem here. 
Inability to create database using hiera parameters

## Related Issues (if any)
new functionality but could also be considered a fix for #1620

## Checklist
- [ x] 🟢 Spec tests.
- [ x] 🟢 Acceptance tests.
- [ x] Manually verified. (For example `puppet apply`)

PR #1693, opened 2026-08-12
enable replication support slot management via hiera

ikonia:
## Summary
the branch is rebased from master as of 12/08/2026 and includes a few outstanding PR's to get the functionality needed. 
enables replication support for Postgres r 

Adds primary-side named replication slot management to postgresql::server closing the last gap in this branch's physical streaming replication support: a standby configured via postgresql::server::recovery

postgresql::server::recovery primary_slot_name had nothing on the primary side to actually create the slot it connects to. 
Previously this required hand-declaring the native postgresql_replication_slot resource directly in a manifest; it's now Hiera-drivable the same way databases/roles/grant_roles/extensions already are.

Validated end-to-end against a real PostgreSQL 18 master/replica pair

postgresql::server::replication_slots (new hash parameter)
- Mirrors the existing grant_roles pattern: a single-level hash, splatted straight onto the native postgresql_replication_slot resource.
- Example: postgresql::server::replication_slots: { standby1_slot: { ensure: present } }
- Known, deliberate limitation: the underlying native type has no port/psql_path/connect_settings parameters, unlike every sibling resource in this module. This only matters for multi-instance-per-host deployments, this can be expanded on if desired in later MR's but is out of scope for this initial enhancement 

Testing

- Full base-backup + streaming replication cycle between a real PG18 master and replica, using a named slot created via this new parameter.
- Validated under real load, not just synthetic test data: a ~170-table Foreman database, restored via pg_dump/pg_restore into the master, replicated correctly to the replica — confirmed via LSN comparison and slot health (wal_status, active) on both sides.
- Spec coverage added/fixed at both the wrapper level (spec/classes/server_spec.rb) and the type level (spec/unit/puppet/type/postgresql_replication_slot_spec.rb). 

## Additional Context

## Related Issues (if any)
this branch and merge contains open PR's already merged in
#1661
#1674

## Checklist
- [ x] 🟢 Spec tests.
- [x ] 🟢 Acceptance tests.
- [ x] Manually verified. (For example `puppet apply`)


manifests/server.pp
```diff
@@ -238,6 +276,25 @@
...
+        dbname => $database,
+      }
+    }
+  }
+
+  $extensions.each |$database, $extensions_hash| {
+    $extensions_hash.each |$extension_name, $extension_params| {
+      postgresql::server::extension { "${database}:${extension_name}":
+        database => $database,
+          *      => $extension_params,
```
corporate-gadfly: By fixing `strict_indent`, it will allow tests to progress further:
```suggestion
        *        => $extension_params,
```

corporate-gadfly (changes requested):

manifests/server.pp
ikonia: nice, I resolved in alignment of the => I didn't notice the * before it being wrong.

ikonia:
messed up the repo trying to fix a commit error will re-submit from a correctly formatted repo,

PR #1707, opened 2026-09-14
(TESTS-1694) enable replication support slot management via hiera

jst-cyr:
This PR is created to execute tests for the original PR (https://github.com/puppetlabs/puppetlabs-postgresql/pull/1694)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ikonia:
glancing through this, this doesn't look as bad as the 'red' suggests

jst-cyr:
**Acceptance Test results:** https://github.com/puppetlabs/puppetlabs-postgresql/actions/runs/34894654788?pr=1707

There were some infra provisioning issues, so I needed to do some rerunning today on targeted systems. Results:

- The Debian 11, AlmaLinux-8, and AlmaLinux-9 failures are an exact match to the failures in the latest nightly run against main. This PR does not seem to introduce any new issues.
- Debian 13 failure is new. Have rerun multiple times and it seems to be a repeatable issue that is not seen in main.

Debian-13 nightly passes, but this PR fails 6 examples in `spec/acceptance/server/default_privileges_spec.rb` with idempotency/apply-exit-code mismatches around `postgresql::server::default_privileges` (grant/revoke default privileges, target-role variants, schema unset). Worth checking whether the replication slot-management change in this branch is affecting default-privileges idempotency on Debian-13, or if it's an environment quirk specific to that image.

This branch will be closed off and feedback shared on the original PR.

